PR #1, opened 2020-10-26
feat: implement general-client (CORE-4030)

z4o4z:


DecathectZero (approved):

DecathectZero:
since this is not on prod, can just merge it in first and test everything

PR #4, opened 2020-11-10
feat: persist chips on reprompt (CORE-4355)

JoshTheDeveloper:


effervescentia (approved):

DecathectZero (changes requested):
@JoshTheDeveloper we now have a dedicated choice trace, and the speak trace should not have choices. So basically separate this out into two distinct traces instead

PR #10, opened 2020-12-17
chore: prepare for general-service nlu

DecathectZero:


server.ts
```diff
@@ -29,19 +37,32 @@ class Server {
...
 
     const app = express();
-    const server = http.createServer(app);
 
     const { middlewares, controllers } = this.serviceManager;
 
     this.app = app;
-    this.server = server;
+
+    const protocol = process.env.NODE_ENV === 'local' && fs.existsSync(KEY_PATH) && fs.existsSync(CERT_PATH) ? Protocol.HTTPS : Protocol.HTTP;
```
DecathectZero, line 45: sometimes people using our SDK might want to run it locally but without https

PR #14, opened 2021-01-06
wip: test cached version

DecathectZero:
we call get version at least 3 or 4 times per interaction. Ideally we can cache it per interaction somehow

### DIFF
--- a/types.ts
+++ b/types.ts
@@ -1,4 +1,5 @@
-import { GeneralRequest } from '@voiceflow/general-types';
+import { Version } from '@voiceflow/api-sdk';
+import { GeneralRequest, GeneralTrace } from '@voiceflow/general-types';
 import * as Runtime from '@voiceflow/runtime';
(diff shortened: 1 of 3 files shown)


types.ts
```diff
@@ -66,6 +67,10 @@ export type MiddlewareGroup = Record<string, Middleware>;
 export type Class<T, A extends any[]> = { new (...args: A): T };
 export type AnyClass = Class<any, any[]>;
 
-export type Context = Runtime.Context<GeneralRequest>;
-export type ContextHandler = Runtime.ContextHandler<GeneralRequest>;
-export type InitContextHandler = Runtime.InitContextHandler<GeneralRequest>;
+export type ContextData = {
+  locale?: string;
+  getVersion: () => Promise<Version<any>>;
```
z4o4z, line 72: I guess we can use GeneralVersion type, isn't it?

z4o4z (approved):

PR #16, opened 2021-01-08
feat: chips

DecathectZero:
so what I wanted to do here:

To show a chip for an intent, we show the utterance with the minimal number of potential slots: 
"I want pizza" over "I want a {small} pizza"

For dialog management, show up to 3 choices to fill in the slots. So if it asks "what size do you want your pizza?"
the chips you would get are "small" "medium" and "large"

this behavior used to be handled in the frontend here, but has now been enhanced:
https://github.com/voiceflow/creator-app/blob/859c14317c5a92bf258457dfe9724dd29acd1c64/src/pages/Prototype/PrototypeTool/utils/intent.ts

### DIFF
--- a/lib/services/chips/utils.ts
+++ b/lib/services/chips/utils.ts
@@ -1,39 +1,75 @@
 import { PrototypeModel } from '@voiceflow/api-sdk';
-import { replaceVariables } from '@voiceflow/runtime';
 import _ from 'lodash';
 
-type Chip = { name: string };
-// eslint-disable-next-line import/prefer-default-export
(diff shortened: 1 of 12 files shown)


lib/services/chips/utils.ts
```diff
@@ -0,0 +1,75 @@
+import { PrototypeModel } from '@voiceflow/api-sdk';
```
leartgjoni-voiceflow, line 1: these 3 funcs would really benefit from some unit tests

leartgjoni-voiceflow (approved):

PR #17, opened 2021-01-11
feat: nlc dialog management

z4o4z:


lib/services/dialog/index.ts
```diff
@@ -93,23 +84,30 @@ class DialogManagement extends AbstractManager<{ utils: typeof utils }> implemen
...
     if (dmStateStore?.intentRequest) {
       logger.debug('@DM - In dialog management context');
 
-      const dmPrefixedResult = await this.makeDMPrefixedInference(
-        incomingRequest.payload.query,
-        dmStateStore.intentRequest.payload.intent.name,
-        version.projectID
-      );
-      const isFallback = this.handleDMContext(dmStateStore, dmPrefixedResult, incomingRequest, languageModel);
+      const dmPrefixedResult = await this.services.nlu.predict({
```
DecathectZero, line 102: yeah we can't really do this because the NLC will completely crap out with the {DM} prefix

DecathectZero (approved):

PR #19, opened 2021-01-11
feat: cached data

DecathectZero:
attach a special dataAPI that caches versions/projects to the context for use through all the handlers

### DIFF
--- a/lib/controllers/interact.ts
+++ b/lib/controllers/interact.ts
@@ -26,7 +26,7 @@ class InteractController extends AbstractController {
     const turn = new TurnBuilder<Context>(stateManager);
     turn.addHandlers(asr, nlu, dialog, runtime).addHandlers(tts, chips);
 
-    return turn.handle({ state, request, versionID, data: { locale } as any });
+    return turn.handle({ state, request, versionID, data: { locale } });
   }
 }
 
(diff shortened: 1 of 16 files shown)


lib/controllers/interact.ts
```diff
@@ -25,7 +26,7 @@ class InteractController extends AbstractController {
     const turn = new TurnBuilder<Context>(stateManager);
     turn.addHandlers(asr, nlu, dialog, runtime).addHandlers(tts, chips);
 
-    return turn.handle({ state, request, versionID, data: { locale } });
+    return turn.handle({ state, request, versionID, data: { locale } as any });
```
DecathectZero, line 29: this is kinda a hack, need to update `@voiceflow/runtime` to make the init handler accept partial data

lib/services/state/index.ts
```diff
@@ -52,13 +52,17 @@ class StateManager extends AbstractManager<{ utils: typeof utils }> implements I
     if (!context.versionID) {
       throw new Error('context versionID not defined');
     }
+
+    // cache per interaction (save version call during request/response cycle)
+    const api = new CacheDataAPI(this.services.dataAPI);
```
DecathectZero, line 57: this is the most important thing that ties everything together

lib/controllers/interact.ts
DecathectZero: done this now

leartgjoni-voiceflow (approved):
solid

PR #22, opened 2021-01-14
fix: allow dm prompts

DecathectZero:
Related to CORE-3090.

### DIFF
--- a/lib/services/dialog/utils.ts
+++ b/lib/services/dialog/utils.ts
@@ -1,8 +1,12 @@
-import { PrototypeModel } from '@voiceflow/api-sdk';
+import { IntentInput, PrototypeModel } from '@voiceflow/api-sdk';
 import { IntentRequest, SLOT_REGEXP } from '@voiceflow/general-types';
 import * as crypto from 'crypto';
 
 export const VF_DM_PREFIX = 'dm_';
+
+export const inputToString = ({ text, voice }: IntentInput) => {
+  return voice?.trim() ? `<voice name="${voice}">${text}</voice>` : text;
+};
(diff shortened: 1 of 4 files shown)


lib/services/dialog/utils.ts
```diff
@@ -1,11 +1,15 @@
...
+import { IntentInput, PrototypeModel } from '@voiceflow/api-sdk';
 import { IntentRequest, RequestType, SLOT_REGEXP } from '@voiceflow/general-types';
 import * as crypto from 'crypto';
 
 import { Context } from '@/types';
 
 export const VF_DM_PREFIX = 'dm_';
 
+export const inputToString = ({ text, voice }: IntentInput) => {
+  return voice?.trim() ? `<voice name="${voice}">${text}</voice>` : text;
```
zhihil, line 10: So I found out that `<voice name="Alexa">` is not valid on Alexa. You simply omit the `<voice>` tag if you want Alexa's voice and only add the tag if you want a custom voice. Do we have an edge case like this to deal with on the general project

Also SSML needs to be wrapped in `<speak>`, do we need that for our general project or no?

DecathectZero: @zhihil yes if you check the TTS code on general-service https://github.com/voiceflow/general-service/blob/2f0cb68c0df3c1de2618d5b43ecfe310e5115565/lib/services/tts/index.ts we automatically put a `<speak>` tag around it.

We also have a thing in there to check for for `DEFAULT_NAMES`, so I think all of your concerns should be addressed

zhihil (approved):

PR #23, opened 2021-01-15
feat: nlc dm

DecathectZero:
The reason why this works is because we already have dialog management built into the NLC:
https://github.com/voiceflow/natural-language-commander/blob/master/src/NaturalLanguageCommander.ts#L222

This was already being used by the FE:
https://github.com/voiceflow/creator-app/blob/master/src/pages/Prototype/PrototypeTool/Dialog.ts#L147

To test this, just go to the NLU `predict` function and make it throw an error from from the axios call every time (comment out the code, throw an error, and ignore dependencies)

turns out we had a bug in the NLC the whole time
https://github.com/voiceflow/natural-language-commander/commit/14dd8094200a519879800dc529315c46041b1f3d

### DIFF
--- a/lib/services/dialog/index.ts
+++ b/lib/services/dialog/index.ts
@@ -6,7 +6,8 @@ import logger from '@/logger';
 import { Context, ContextHandler } from '@/types';
 
 import { generateVariations } from '../chips/utils';
-import { getNoneIntentRequest } from '../nlu/utils';
+import { handleNLCDialog } from '../nlu/nlc';
+import { getNoneIntentRequest, NONE_INTENT } from '../nlu/utils';
 import { isIntentRequest } from '../runtime/types';
 import { AbstractManager, injectServices } from '../utils';
 import { dmPrefix, fillStringEntities, getDMPrefixIntentName, getIntentEntityList, getUnfulfilledEntity, VF_DM_PREFIX } from './utils';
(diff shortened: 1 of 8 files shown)


lib/services/dialog/index.ts
```diff
@@ -102,14 +96,17 @@ class DialogManagement extends AbstractManager<{ utils: typeof utils }> implemen
...
         model: version.prototype.model,
-        nlcQuery: incomingRequest.payload.query,
         projectID: version.projectID,
+        dmIntent: dmStateStore.intentRequest,
       });
 
       const isFallback = this.handleDMContext(dmStateStore, dmPrefixedResult, incomingRequest, version.prototype.model);
 
       if (isFallback) {
-        return fallbackIntent(context);
```
DecathectZero, line 112: this was just some code cleanup, fallbackIntent is not needed now that we have `DialogManagement.setDMStore`

lib/services/dialog/index.ts
```diff
@@ -99,17 +93,29 @@ class DialogManagement extends AbstractManager<{ utils: typeof utils }> implemen
...
-      const dmPrefixedResult = await this.services.nlu.predict({
-        query: `${dmPrefix(dmStateStore.intentRequest.payload.intent.name)} ${incomingRequest.payload.query}`,
-        model: version.prototype.model,
-        nlcQuery: incomingRequest.payload.query,
-        projectID: version.projectID,
-      });
-
-      const isFallback = this.handleDMContext(dmStateStore, dmPrefixedResult, incomingRequest, version.prototype.model);
+      const { query } = incomingRequest.payload;
+      try {
```
DecathectZero, line 97: this whole try catch section is a little ugly, if it can be improved, please do make a commit

z4o4z (approved):

PR #24, opened 2021-01-16
fix: built in nlc intents

z4o4z:


DecathectZero:
merging this instead for now - coming up with a different solution for built in intents
https://github.com/voiceflow/general-runtime/pull/25

PR #53, opened 2021-02-02
feat: only add tts handler if specified in config (CORE-4944)

jgoping:
**Fixes or implements CORE-4944**

A config is now passed to the interact endpoint to specify whether the tts handler should be included or not.

The config type was specified in general-types and is added to the request body.

### Related PRs

| justin/advanced-config-options/CORE-4944 | [link](https://github.com/voiceflow/runtime-client-js/pull/6) |

### Checklist

- [ Check ] title of PR reflects the branch name
- [ Check ] all commits adhere to conventional commits


lib/controllers/interact.ts
```diff
@@ -17,19 +17,25 @@ class InteractController extends AbstractController {
...
+      body: { state, request = null, config },
       params: { versionID },
       query: { locale },
     } = req;
 
     const turn = new TurnBuilder<Context>(stateManager);
-    turn.addHandlers(asr, nlu, dialog, runtime).addHandlers(tts, chips);
+
+    turn.addHandlers(asr, nlu, dialog, runtime);
+    if (config?.tts) {
```
zhihil, line 34: There's a bug here. If `config` doesn't exist (which is the case for all of our current requests to the `/interact` calls in the codebase), then the expression `config?.tts` evaluates to `undefined` which is falsy. 

This will cause all existing requests to strip the TTS and probably break the prototyping tool. 

You have make sure that for the two edge-cases where `config` is `undefined` or `config.tts` is `undefined`, then these are treated as truthy.

lib/controllers/interact.ts
```diff
@@ -17,19 +17,25 @@ class InteractController extends AbstractController {
...
+      body: { state, request = null, config },
       params: { versionID },
       query: { locale },
     } = req;
 
     const turn = new TurnBuilder<Context>(stateManager);
-    turn.addHandlers(asr, nlu, dialog, runtime).addHandlers(tts, chips);
+
+    turn.addHandlers(asr, nlu, dialog, runtime);
+    if (config?.tts !== false) {
```
zhihil, line 34: This is fine but the logic is a little obscure. 

You're relying on the specific mechanics of the TypeScript `?` operator, where it evaluates to `undefined` if `config` is `null` or `undefined`. 

Then you're relying on the fact that JavaScript strict equality doesn't cast the types of the left and right operands before making a comparison. Thus, because `undefined` is a different type than `boolean` then they are not equal, which captures the edge cases. 

I think it helps to write the condition more explicitly:

`!config || _.isUndefined(config.tts) || config.tts` 

to better state what you want to do here, otherwise, someone who needs to work on this code might not understand what's going on.

zhihil (approved):

DecathectZero (approved):

PR #58, opened 2021-02-03
feat: add intent confidence to debug message (CORE-5004)

z4o4z:
**Fixes or implements CORE-5004**

### Related PRs

https://github.com/voiceflow/general-service/pull/55
https://github.com/voiceflow/creator-app/pull/2923

### DIFF
--- a/lib/services/nlu/index.ts
+++ b/lib/services/nlu/index.ts
@@ -34,6 +34,7 @@ class NLU extends AbstractManager<{ utils: typeof utils }> implements ContextHan
         query,
       })
       .catch(() => ({ data: null }));
+
     if (data) {
       return data;
     }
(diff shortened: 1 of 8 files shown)


effervescentia (approved):

effervescentia (changes requested):
should add tests for the new debug messages

z4o4z:
> should add tests for the new debug messages

we do not have tests for handlers at all, I guess we should create a ticket for next cooldown sprint to add tests for handlers

DecathectZero (changes requested):
I think we should actually put the debug traces one level higher - in either the nlu/dialog handlers seen here https://github.com/voiceflow/general-runtime/blob/master/lib/controllers/interact.ts this way we don’t have to implement it for every single block, but rather when requests come in.

So I think the dialog handler or the very beginning of the runtime handler is a good place - as long as we receive an intent request we can show the trace.

DecathectZero (approved):

PR #59, opened 2021-02-04
fix: nested replace (CORE-000)

DecathectZero:

`test = "{test_nested}"`
and if we put that into a speak block with `{test}` and have it display the value of `{test_nested}` we want it to replace twice?

so if we want to pass in a variable that contains a string with "variables" you would technically need to replace it twice. This is an edge case. I think in the future we make build this into the replaceVariables function.

Although I'm not sure if this is the base behavior, might need some further discussion. While we allow it to go into a depth of 2 here - is there an actual limit we want to enforce.


leartgjoni-voiceflow (approved):
can u add a test case that shows the situation where this would be needed. good to have as "built-in" docs

PR #64, opened 2021-02-18
feat: one shot intent handler (CORE-000)

DecathectZero:

So there is this edge case if the user sends a text request when the stack hasn't loaded yet or their session is not initialized. That means that they will be on the start block.

What we do is see if there are any intent blocks or command blocks that can handle their request. If not, just proceed down the start block as normal, and consider the request handled.

We already use this pattern for alexa

### Related PRs

### Checklist


leartgjoni-voiceflow (approved):
tests?

PR #65, opened 2021-02-19
feat: API Key Verification (CORE-5221)

jgoping:
**Fixes or implements CORE-5221**

Allows for API key support when hitting the interact endpoint.

For req with origin voiceflow.com, dataAPI uses server-data-api. Otherwise, it uses creator-api and attaches the API key that the user request should have. The API key could come from an environment variable, or overrode by one in the request header.

This keeps server-data-api as our internal interface to the db and creator-api as the external interface protected by auth_vf or api-key.

DataAPI is now a class instead of a function, which creates all three APIs if it has the information to. When an API call needs to be made, get() is called, which takes in the authorization (API key or auth_vf) and the origin of the request and determines which API it should make the call to.

⚠️⚠️⚠️ CREATOR_APP_ORIGIN is a new environment variable which needs to be updated in production. This is needed so that dataAPI can use server-data-api if a request's origin is CREATOR_APP_ORIGIN. ⚠️⚠️⚠️

### Related PRs
| api-sdk     | [link](https://github.com/voiceflow/api-sdk/pull/50) |

### Checklist

- [ ✅  ] title of PR reflects the branch name
- [ ✅  ] all commits adhere to conventional commits
- [ ✅  ] appropriate tests have been written
- [ ✅  ] all the dependencies are upgraded

### DIFF
--- a/config.ts
+++ b/config.ts
@@ -41,7 +41,7 @@ const CONFIG: Config = {
   CREATOR_API_AUTHORIZATION: getOptionalProcessEnv('CREATOR_API_AUTHORIZATION'),
 
   // creator-app config
-  CREATOR_APP_ENDPOINT: getOptionalProcessEnv('CREATOR_APP_ENDPOINT'),
+  CREATOR_APP_ORIGIN: getOptionalProcessEnv('CREATOR_APP_ORIGIN'),
 
   AWS_ACCESS_KEY_ID: getOptionalProcessEnv('AWS_ACCESS_KEY_ID'),
   AWS_SECRET_ACCESS_KEY: getOptionalProcessEnv('AWS_SECRET_ACCESS_KEY'),
--- a/config.ts
+++ b/config.ts
@@ -36,9 +36,12 @@ const CONFIG: Config = {
   VF_DATA_ENDPOINT: getOptionalProcessEnv('VF_DATA_ENDPOINT'), // server-data-api endpoint
   ADMIN_SERVER_DATA_API_TOKEN: getOptionalProcessEnv('ADMIN_SERVER_DATA_API_TOKEN'), // Server-data-api auth token
 
-  // creator-api conifg
+  // creator-api config
   CREATOR_API_ENDPOINT: getOptionalProcessEnv('CREATOR_API_ENDPOINT'),
   CREATOR_API_AUTHORIZATION: getOptionalProcessEnv('CREATOR_API_AUTHORIZATION'),
+
+  // creator-app config
+  CREATOR_APP_ENDPOINT: getOptionalProcessEnv('CREATOR_APP_ENDPOINT'),
 
   AWS_ACCESS_KEY_ID: getOptionalProcessEnv('AWS_ACCESS_KEY_ID'),
   AWS_SECRET_ACCESS_KEY: getOptionalProcessEnv('AWS_SECRET_ACCESS_KEY'),
--- a/lib/clients/dataAPI.ts
+++ b/lib/clients/dataAPI.ts
@@ -13,7 +13,7 @@ class DataAPI {
 
   remoteDataApi: RemoteDataAPI | undefined;
 
-  creatorAppEndpoint: string;
+  creatorAppOrigin: string | null;
 
   creatorAPIEndpoint: string | null;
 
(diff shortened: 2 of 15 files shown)


lib/clients/dataAPI.ts
```diff
@@ -8,32 +8,75 @@ import Static from './static';
...
+      }
+
+      return this.remoteDataApi;
+    }
+    if (!this.creatorDataApi) {
+      throw new Error('no creator data API env configuration set');
+    }
+
+    this.creatorDataApi!.updateAuthorization(authorization);
+    return this.creatorDataApi;
```
leartgjoni-voiceflow, line 78: I think this might be a problem. Here we are re-using and updating the same obj possibly between multiple requests. In case of a race condition, you could have req1 having their apiClient authorization updated by req2 before req1's call to creator-api is sent. I think a better way is on dataAPI.get() to generate creatorDataApi on the fly. This will also remove the need for modifying `creatorDataAPI` in `runtime`.

lib/controllers/interact.ts
```diff
@@ -39,7 +40,7 @@ class InteractController extends AbstractController {
 
     turn.addHandlers(chips);
 
-    return turn.resolve({ state, request, versionID, data: { locale } });
+    return turn.resolve({ state, request, versionID, data: { locale, authorization, origin } });
```
leartgjoni-voiceflow, line 43: I'd be better to wrap `authorization` and `origin` in a parent object, just for organization

leartgjoni-voiceflow (changes requested):

config.ts
```diff
@@ -36,10 +36,13 @@ const CONFIG: Config = {
   VF_DATA_ENDPOINT: getOptionalProcessEnv('VF_DATA_ENDPOINT'), // server-data-api endpoint
   ADMIN_SERVER_DATA_API_TOKEN: getOptionalProcessEnv('ADMIN_SERVER_DATA_API_TOKEN'), // Server-data-api auth token
 
-  // creator-api conifg
+  // creator-api config
   CREATOR_API_ENDPOINT: getOptionalProcessEnv('CREATOR_API_ENDPOINT'),
   CREATOR_API_AUTHORIZATION: getOptionalProcessEnv('CREATOR_API_AUTHORIZATION'),
 
+  // creator-app config
+  CREATOR_APP_ENDPOINT: getOptionalProcessEnv('CREATOR_APP_ENDPOINT'),
```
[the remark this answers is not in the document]
jgoping: Fixed

lib/clients/dataAPI.ts
```diff
@@ -8,32 +8,75 @@ import Static from './static';
...
+    if (ADMIN_SERVER_DATA_API_TOKEN && VF_DATA_ENDPOINT) {
+      this.remoteDataApi = new API.RemoteDataAPI(
+        { platform: 'general', adminToken: ADMIN_SERVER_DATA_API_TOKEN, dataEndpoint: VF_DATA_ENDPOINT },
+        { axios: Static.axios }
+      );
+    }
+
+    // fetch from creator-api
+    if (CREATOR_API_ENDPOINT) {
+      this.creatorDataApi = new API.CreatorDataApi({ endpoint: `${CREATOR_API_ENDPOINT}/v2`, authorization: CREATOR_API_AUTHORIZATION ?? undefined });
```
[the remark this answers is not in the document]
jgoping: Fixed

lib/clients/dataAPI.ts
jgoping: Fixed

lib/controllers/interact.ts
jgoping: Fixed

lib/services/runtime/index.ts
```diff
@@ -43,7 +45,11 @@ class RuntimeManager extends AbstractManager<{ utils: typeof utils }> implements
   public async handle({ versionID, state, request, ...context }: Context) {
     if (!isRuntimeRequest(request)) throw new Error(`invalid runtime request type: ${JSON.stringify(request)}`);
 
-    const runtime = this.client.createRuntime(versionID, state, request, { api: context.data.api });
+    if (!this.client) {
+      this.createClient(context.data.api);
+    }
```
leartgjoni-voiceflow, line 50: I think this means we are only setting it once with whatever `context.data.api` we have for that first request. Instead we should just construct the new client on the fly inside of handle everytime. `handle` gets called once per request

leartgjoni-voiceflow (changes requested):

lib/services/runtime/index.ts
jgoping: Yeah, it was getting set with the updated context.data.api in createRuntime() but this code isn't very clear. Updated it now to construct the client each time.

leartgjoni-voiceflow (approved):

DecathectZero (approved):

PR #68, opened 2021-02-24
fix: remove origin requirement (CORE-000)

DecathectZero:

If we check for origin in the rate limiter middleware, there's no point to do the checkin dataAPI, since it is assume any call that comes in is already validated. Also did a cleaner pattern for `creatorDataApi` with less instance variables.

For deciding whether to use `remoteAPI` or `creatorDataApi`: if the user tries to make an authenticated call, just use `creatorDataApi`

### Related PRs

### Checklist

### DIFF
--- a/tests/lib/clients/dataAPI.unit.ts
+++ b/tests/lib/clients/dataAPI.unit.ts
@@ -21,9 +21,7 @@ describe('dataAPI client unit tests', () => {
       PROJECT_SOURCE: 'cool.vf',
       ADMIN_SERVER_DATA_API_TOKEN: 'token',
       VF_DATA_ENDPOINT: 'endpoint',
-      CREATOR_API_AUTHORIZATION: 'creator auth',
       CREATOR_API_ENDPOINT: 'creator endpoint',
-      CREATOR_APP_ORIGIN: 'voiceflow.com',
     };
 
     expect(await new DataAPI(config as any, API as any).get()).to.eql({ type: 'local' });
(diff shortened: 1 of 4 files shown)


tests/lib/clients/dataAPI.unit.ts
```diff
@@ -100,13 +91,11 @@ describe('dataAPI client unit tests', () => {
...
       CREATOR_API_AUTHORIZATION: 'creator auth',
       CREATOR_API_ENDPOINT: 'creator endpoint',
-      CREATOR_APP_ORIGIN: 'voiceflow.com',
     };
 
-    const origin = 'voiceflow.com';
     const dataAPI = new DataAPI(config as any, API as any);
 
-    expect(dataAPI.get('', origin)).to.be.rejectedWith('no remote data API env configuration set');
+    expect(dataAPI.get()).to.be.rejectedWith('no remote data API env configuration set');
```
jgoping, line 98: This error message no longer matches with what you have

tests/lib/clients/dataAPI.unit.ts
```diff
@@ -122,9 +111,8 @@ describe('dataAPI client unit tests', () => {
       CREATOR_APP_ORIGIN: 'voiceflow.com',
     };
 
-    const origin = 'other-site.com';
     const dataAPI = new DataAPI(config as any, API as any);
 
-    expect(dataAPI.get('', origin)).to.be.rejectedWith('no creator data API env configuration set');
+    expect(dataAPI.get()).to.be.rejectedWith('no creator data API env configuration set');
```
jgoping, line 116: This error message no longer matches with what you have

jgoping (changes requested):
Update unit test expected error messages, otherwise LGTM

PR #72, opened 2021-03-03
feat: trace handler (CORE-5344)

leartgjoni-voiceflow:
to-test:
- import [test trace block.zip](https://github.com/voiceflow/general-runtime/files/6111400/test.trace.block.zip)

manual tests performed:
- default path (doesn't stop)
- inline events (event1, event2)
- global jump events (event3, event4)
- global push event (help)
- hit intent (travel_intent) from trace block
- hit command (second_flow) from trace block

### DIFF
--- a/lib/services/nlu/index.ts
+++ b/lib/services/nlu/index.ts
@@ -69,7 +69,7 @@ class NLU extends AbstractManager<{ utils: typeof utils }> implements ContextHan
     }
 
     const request = await this.predict({
-      query: context.request.payload,
+      query: (context.request.payload as unknown) as string,
       model: version.prototype?.model,
       locale: version.prototype?.data.locales[0] as Locale,
       projectID: version.projectID,
--- a/lib/services/nlu/index.ts
+++ b/lib/services/nlu/index.ts
@@ -4,8 +4,9 @@
  */
 
 import { PrototypeModel } from '@voiceflow/api-sdk';
-import { IntentRequest, Locale, RequestType } from '@voiceflow/general-types';
+import { IntentRequest, Locale } from '@voiceflow/general-types';
 
+import { isTextRequest } from '@/lib/services/runtime/types';
 import { Context, ContextHandler } from '@/types';
 
 import { AbstractManager, injectServices } from '../utils';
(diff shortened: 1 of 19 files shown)


DecathectZero (approved):

lib/services/runtime/handlers/event/index.ts
```diff
@@ -1,17 +1,17 @@
...
+import { GeneralRuntime, isGeneralRequest, isIntentRequest } from '@/lib/services/runtime/types';
 
 import { mapEntities } from '../../utils';
 
 export const intentEventMatcher = {
-  match: (context: { runtime: GeneralRuntime; event: GeneralEvent | null }): context is { runtime: Runtime<IntentRequest>; event: IntentEvent } => {
+  match: (context: { runtime: GeneralRuntime; event: Event | null }): context is { runtime: Runtime<IntentRequest>; event: IntentEvent } => {
     const request = context.runtime.getRequest();
     if (!isIntentRequest(request)) return false;
     if (context.event?.type !== EventType.INTENT) return false;
```
DecathectZero, line 13: I guess we can do an intent event typeguard here too, instead of casting `as`

DecathectZero: nevermind, I guess the whole point of this function is to be a typeguard

lib/services/runtime/handlers/trace.ts
```diff
@@ -0,0 +1,54 @@
...
+import CommandHandler from './command';
+import { findEventMatcher } from './event';
+
+const utilsObj = {
+  commandHandler: CommandHandler(),
+  findEventMatcher,
+};
+
+export const TraceHandler: HandlerFactory<Node, typeof utilsObj> = (utils) => ({
+  canHandle: (node) => !!node._v,
```
DecathectZero, line 15: I would do `_v === 1` so its even more explicit

lib/services/nlu/index.ts
```diff
@@ -69,7 +69,7 @@ class NLU extends AbstractManager<{ utils: typeof utils }> implements ContextHan
     }
 
     const request = await this.predict({
-      query: context.request.payload,
+      query: (context.request.payload as unknown) as string,
```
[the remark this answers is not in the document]
leartgjoni-voiceflow: Yea I've tried to create a typeguard but the types are not compatible.

Screen Shot 2021-03-10 at 9 35 40 AM

cause payload is `Record<string, unknown>`, I guess we could change it to just `unknown`? @DecathectZero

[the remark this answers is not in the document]
DecathectZero: yeah, since TextRequest is actually just `payload: string`

PR #81, opened 2021-03-16
feat: custom stop on v1 (CORE-5457)

leartgjoni-voiceflow:
to-do:
- update general-types

[general-types](https://github.com/voiceflow/general-types/pull/59)
[runtime-client-js](https://github.com/voiceflow/runtime-client-js/pull/52)

### DIFF
--- a/lib/services/runtime/handlers/_v1.ts
+++ b/lib/services/runtime/handlers/_v1.ts
@@ -3,6 +3,7 @@
 import { Node, TraceFrame } from '@voiceflow/general-types/build/nodes/_v1';
 import { Action, HandlerFactory } from '@voiceflow/runtime';
 
+import { StorageType } from '../types';
 import CommandHandler from './command';
 import { findEventMatcher } from './event';
 
@@ -45,9 +46,10 @@ export const _V1Handler: HandlerFactory<Node, typeof utilsObj> = (utils) => ({
       payload: { data: node.payload, paths: node.paths },
     });
 
+    const stop = runtime.storage.get(StorageType.STOP_TYPES) ?? node.stop;
     // if !stop continue to defaultPath otherwise
     // quit cycleStack without ending session by stopping on itself
-    return !node.stop ? defaultPath : node.id;
+    return !stop ? defaultPath : node.id;
   },
 });
 
--- a/lib/services/runtime/handlers/_v1.ts
+++ b/lib/services/runtime/handlers/_v1.ts
@@ -1,7 +1,7 @@
 /* eslint-disable no-restricted-syntax */
-
 import { Node, TraceFrame } from '@voiceflow/general-types/build/nodes/_v1';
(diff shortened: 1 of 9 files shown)


lib/services/runtime/handlers/_v1.ts
```diff
@@ -42,12 +43,15 @@ export const _V1Handler: HandlerFactory<Node, typeof utilsObj> = (utils) => ({
 
     runtime.trace.addTrace<TraceFrame>({
       type: node.type,
-      payload: { data: node.payload, paths: node.paths },
+      payload: { data: node.payload, paths: node.paths, stop: node.stop, defaultPath: node.defaultPath },
     });
 
+    const stopTypes = runtime.storage.get<string[]>(StorageType.STOP_TYPES) || [];
+
+    const stop = stopTypes.includes(_.get(node.payload, 'name')) || node.stop;
```
DecathectZero, line 51: I'm curious why we have to use _.get instead of `node.payload?.name`?

DecathectZero (approved):

lib/services/runtime/types.ts
```diff
@@ -18,11 +19,12 @@ export const isIntentRequest = (request: RuntimeRequest): request is IntentReque
...
 
 export const isRuntimeRequest = (request: any): request is RuntimeRequest => {
-  return request === null || !!(typeof request.type === 'string' && !!request.type && request.payload);
+  return request === null || !!(typeof request.type === 'string' && !!request.type);
 };
 
-export const isGeneralRequest = (request: RuntimeRequest): request is Request<string, { name: string }> => {
-  return !!(request as Request<string, { name: string }>)?.payload.name;
+type GeneralRequest = Request<string, undefined | {}>;
+export const isGeneralRequest = (request: RuntimeRequest): request is GeneralRequest => {
```
DecathectZero, line 26: I don't even know if it's the best to do a `isGeneralRequest` check anymore - I think it's okay for the request to contain data. We simply just assume every request is a general request if it doesn't match anything else prior (like intent request?)

lib/services/runtime/handlers/_v1.ts
leartgjoni-voiceflow: `payload` is of type `unknown`

lib/services/runtime/handlers/event/index.ts
```diff
@@ -28,9 +28,9 @@ export const generalEventMatcher = {
     event: Event | null;
   }): context is { runtime: Runtime<Request>; event: Event<string, GeneralEvent> } => {
     const request = context.runtime.getRequest();
-    if (!isGeneralRequest(request)) return false;
+    if (!request || isIntentRequest(request)) return false;
```
leartgjoni-voiceflow, line 31: if intent request - don't match. Example, if stopped on an interaction block and an intent comes in, without this check the `generalEventMatcher` will return `true` thus matching the first path in the interaction node no matter what.

PR #83, opened 2021-03-17
feat: preliminary handler (CORE-4993)

leartgjoni-voiceflow:
https://creator-env-hopeful-morse.development.voiceflow.com/

### DIFF
--- a/lib/services/runtime/handlers/state/preliminary.ts
+++ b/lib/services/runtime/handlers/state/preliminary.ts
@@ -1,0 +1,29 @@
+import { Node } from '@voiceflow/api-sdk';
+import { HandlerFactory } from '@voiceflow/runtime';
+
+import { eventHandlers } from '..';
+import { isGeneralRequest, isIntentRequest } from '../../types';
+import CommandHandler from '../command';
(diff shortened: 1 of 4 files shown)


lib/services/runtime/handlers/state/preliminary.ts
```diff
@@ -0,0 +1,43 @@
+import { Node } from '@voiceflow/api-sdk';
+import { Action, HandlerFactory } from '@voiceflow/runtime';
+
+import _V1Handler from '../_v1';
+import { isGeneralRequest, isIntentRequest } from '../../types';
+import CaptureHandler from '../capture';
+import CommandHandler from '../command';
+import InteractionHandler from '../interaction';
+
+export const eventHandlers = [CaptureHandler(), InteractionHandler(), _V1Handler()];
```
leartgjoni-voiceflow, line 10: the bug was caused by this living in `handlers/index.ts`, that created a circular dependency and `eventHandlers` ended up being undefined. Declared locally to avoid it

DecathectZero (approved):

PR #92, opened 2021-03-31
fix: state management api revisions (CORE-5571)

leartgjoni-voiceflow:


lib/controllers/stateManagement/requests.ts
```diff
@@ -0,0 +1,27 @@
...
+import { ArrayField, ObjectField, RecordField, StringField } from '@/lib/controllers/schemaTypes';
+
+const StackFrame = {
+  type: 'object',
+  additionalProperties: false,
+  required: ['programID', 'storage', 'variables'],
+  properties: {
+    nodeID: StringField('nodeID'),
+    programID: StringField('programID'),
+    storage: ObjectField('storage'),
```
DecathectZero, line 11: make storage, commands, and variables optional - the runtime can handle it

lib/controllers/stateManagement/requests.ts
```diff
@@ -0,0 +1,27 @@
...
+    variables: ObjectField('variables'),
+  },
+};
+
+export const UpdateSchema = {
+  type: 'object',
+  additionalProperties: false,
+  required: ['stack', 'storage', 'variables'],
+  properties: {
+    turn: ObjectField('turn'),
```
DecathectZero, line 22: we don't need turn actually, or make it optional I guess

lib/services/interact.ts
```diff
@@ -29,6 +29,12 @@ class Interact extends AbstractManager {
       headers: { authorization, origin },
     } = req;
 
+    const isLaunchRequest = request?.type === RequestType.LAUNCH;
+    if (isLaunchRequest && state) {
```
DecathectZero, line 33: I know it is kind of bad practice but if its possible to override typescript maybe `request = null` is cleaner than doing the ternary below

lib/services/session/mongo.ts
```diff
@@ -14,33 +14,46 @@ class SessionManager extends AbstractManager {
     return config.SESSIONS_SOURCE === Source.MONGO;
   }
 
-  async saveToDb(userId: string, state: State) {
+  private getSessionID(versionID: string, userID: string) {
```
DecathectZero, line 17: should we move `getSessionID` to a common session/utils file? since this is the same as `lib/services/session/local.ts` but also I kind of get what you're going for just in case different DBs might need different indexing

DecathectZero (changes requested):
Left some comments, most of them just questions. 

This is for way later on down the road - I think one thing we haven't fully considered is when versioning gets good and people use the API for production apps - how are we going to move sessions between different versions. I was thinking maybe some kind of automated migration system where it reads all the sessions of a certain version and just changes their ids to the new version. Because the other way we would do it is by projectID - but sometimes a version might change so drastically your stack becomes useless.

Or maybe we just don't even need session migrations at all. If its a new version, they reference a new API endpoint and everyone just gets new sessions lol.

lib/controllers/stateManagement/requests.ts
leartgjoni-voiceflow: it's optional already

lib/controllers/stateManagement/requests.ts
leartgjoni-voiceflow: State:
Screen Shot 2021-04-01 at 8 59 21 AM
Frame state:
Screen Shot 2021-04-01 at 8 59 35 AM

followed this types. what's optional there I left as optional. Also users should be sending us a state that they first pulled from our api, they shouldn't be making up their own, that's why I'd prefer to be more strict

lib/services/interact.ts
leartgjoni-voiceflow: `request` was a const that's why I went this way, but yea turning it into `let`

lib/services/session/mongo.ts
leartgjoni-voiceflow: yea I'd prefer for them to be separate since they are specific to class they live in. (also applying a change to the mongo one to make it more useful)

leartgjoni-voiceflow:
> Left some comments, most of them just questions.
> 
> This is for way later on down the road - I think one thing we haven't fully considered is when versioning gets good and people use the API for production apps - how are we going to move sessions between different versions. I was thinking maybe some kind of automated migration system where it reads all the sessions of a certain version and just changes their ids to the new version. Because the other way we would do it is by projectID - but sometimes a version might change so drastically your stack becomes useless.
> 
> Or maybe we just don't even need session migrations at all. If its a new version, they reference a new API endpoint and everyone just gets new sessions lol.

@DecathectZero I think we should go with the prefixing by projectID, the problem about "the version changing so drastically your stack becomes useless" it's something we are have on production currently, and we haven't seen much problems from it. I think it's a way better solution than the "automatic migration" which seems way more complex for a problem that isn't that impactful. Like on alexa you can wipe your session from the app, our API also provides the `reset` endpoint if things get that bad.

DecathectZero (approved):

DecathectZero:
merge this into `leart/state-management/CORE-5560` and I will do my testing and look at the overall diff there before merging to master @leartgjoni-voiceflow

PR #94, opened 2021-04-07
fix: stop all (CORE-000)

DecathectZero:

For the prototype tool, if we want to stop on all trace/action blocks - there's no way to do it because we actually don't know all the types beforehand. So we actually need a separate command.

The other way we could do this is to have a special keyword like `any` in the stopTypes.

### Related PRs

### Checklist

- [ ] API documention is up to date


leartgjoni-voiceflow (approved):
good idea, fix/add tests

PR #102, opened 2021-04-13
feat: if and set v2 (VF-182)

leartgjoni-voiceflow:
To test:
- download [Archive.zip](https://github.com/voiceflow/general-runtime/files/6320430/Archive.zip)
- import mongodb export
- import postman collection


lib/services/runtime/handlers/index.ts
```diff
@@ -33,10 +35,12 @@ export default ({ API_HANDLER_ENDPOINT, INTEGRATIONS_HANDLER_ENDPOINT, CODE_HAND
   EndHandler(),
   FlowHandler(),
   IfHandler(),
+  IfV2Handler({ endpoint: CODE_HANDLER_ENDPOINT }),
```
DecathectZero, line 38: are we shifting everything over to executing locally?

I guess as long as `CODE_HANDLER_ENDPOINT` is not declared it is fine.

DecathectZero (approved):

lib/services/runtime/handlers/_v1.ts
```diff
@@ -13,7 +13,8 @@ const utilsObj = {
 };
 
 export const _V1Handler: HandlerFactory<Node, typeof utilsObj> = (utils) => ({
-  canHandle: (node) => node._v === 1,
+  canHandle: (node, runtime) =>
+    node._v === 1 || (node.type === NodeType.IF_V2 && (runtime.turn.get<string[]>(TurnType.STOP_TYPES) || []).includes(NodeType.IF_V2)),
```
DecathectZero, line 17: I think ideally - inside the `IfV2 handler` we call this `_V1Handler` - rather than have the logic in the `canHandle` of both, probably a little more confusing in the future.

leartgjoni-voiceflow: not sure about this, it'd need for `_v1` to live in platform, see here: https://github.com/voiceflow/runtime/pull/107#discussion_r618396578
Otherwise I could inject `_v1` during `ifV2` init, here: 
https://github.com/voiceflow/general-runtime/pull/102/files#diff-de87f9769e26c15dc082d4959f9a1bfd5fe811a2a8d1e691b66e67d70b90ad66R38

lib/services/runtime/handlers/index.ts
leartgjoni-voiceflow: yea, we should exe locally for the current code to work. will notify Frank to remove the env var before merging
Otherwise we can go back to consider something like [this](https://github.com/voiceflow/runtime/pull/108/files) for remote exe

DecathectZero (approved):

PR #108, opened 2021-04-28
fix: no slot filling if intent out of scope (VF-775)

leartgjoni-voiceflow:
[Voiceflow_Chatbot-6089a76a5f8f1766cc4569ab.vf.zip](https://github.com/voiceflow/general-runtime/files/6401603/Voiceflow_Chatbot-6089a76a5f8f1766cc4569ab.vf.zip)


DecathectZero (approved):

lib/services/dialog/utils.ts
```diff
@@ -60,3 +69,37 @@ export const getIntentEntityList = (intentName: string, model: PrototypeModel) =
...
+  const client = new Client({
+    api,
+  });
+
+  const runtime = client.createRuntime(versionID, state, request);
+
+  const currentFrame = runtime.stack.top();
+  const program = await runtime.getProgram(currentFrame.getProgramID());
+  const node = program.getNode(currentFrame.getNodeID());
+  const variables = Store.merge(runtime.variables, currentFrame.variables);
```
DecathectZero, line 83: I think it's really likely for us to have a null/undefined anywhere along here. We might want to do some checks.

I.e. empty stack, end of flow, etc.

lib/services/dialog/utils.ts
```diff
@@ -60,3 +69,37 @@ export const getIntentEntityList = (intentName: string, model: PrototypeModel) =
...
+  if (runtime.getAction() === Action.RESPONSE || !node) return false;
+  // if no event handler can handle, intent req is out of scope => no dialog management required
+  if (!eventHandlers.find((h) => h.canHandle(node as any, runtime, variables, program))) return false;
+
+  // if interaction node - check if req intent matches one of the node intents
+  for (const interaction of (node.interactions || []) as any) {
+    const { event } = interaction;
+
+    const matcher = findEventMatcher({ event, runtime, variables });
+    if (matcher) {
```
DecathectZero, line 94: `if(findEventMatcher({ event, runtime, variables }))`

lib/services/dialog/utils.ts
```diff
@@ -60,3 +69,37 @@ export const getIntentEntityList = (intentName: string, model: PrototypeModel) =
...
+  const program = await runtime.getProgram(currentFrame.getProgramID());
+  const node = program.getNode(currentFrame.getNodeID());
+  const variables = Store.merge(runtime.variables, currentFrame.variables);
+
+  if (runtime.getAction() === Action.RESPONSE || !node) return false;
+  // if no event handler can handle, intent req is out of scope => no dialog management required
+  if (!eventHandlers.find((h) => h.canHandle(node as any, runtime, variables, program))) return false;
+
+  // if interaction node - check if req intent matches one of the node intents
+  for (const interaction of (node.interactions || []) as any) {
```
DecathectZero, line 90: I don't think this solution accounts for intent blocks/command blocks

DecathectZero:
made a few changes in my PR: a lot more type checking and null/undefined safety. It is very easy for `stack.top()` or `getProgram` to be undefined.

Moved the command check before the node interaction check. This is because commands are baked into the stack and don't require the `getProgram` to determine validity

leartgjoni-voiceflow:
> made a few changes in my PR: a lot more type checking and null/undefined safety. It is very easy for `stack.top()` or `getProgram` to be undefined.
> 
> Moved the command check before the node interaction check. This is because commands are baked into the stack and don't require the `getProgram` to determine validity

awesome thanks

PR #121, opened 2021-05-27
feat: patch variables (VF-000)

DecathectZero:

So several users using the State API have asked about how they can use an external service to update the user's runtime variables. Right now the only way is to GET the user's state, and then modify the variable, and then PUT it back in.

I'm just condensing this into a single endpoint. I totally understand that there is a way to atomically merge objects with the mongo syntax, and that this has a slight race condition, but it is pretty clean and should work for the time being.

I've tested it a few times - interesting edge case being if the user has no state at all, and we update the variables, the `interact` still works fine.

Looking for feedback on the endpoint and presentation of this.

### Related PRs

### Checklist

### DIFF
--- a/lib/controllers/stateManagement/index.ts
+++ b/lib/controllers/stateManagement/index.ts
@@ -11,6 +11,7 @@ const { body, header, query } = Validator;
 const VALIDATIONS = {
   BODY: {
     UPDATE_SESSION: body().custom(customAJV(UpdateSchema)),
+    OBJECT: body().exists(),
   },
   HEADERS: {
     PROJECT_ID: header('project_id')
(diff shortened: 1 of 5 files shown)


lib/controllers/stateManagement/index.ts
```diff
@@ -11,6 +11,7 @@ const { body, header, query } = Validator;
 const VALIDATIONS = {
   BODY: {
     UPDATE_SESSION: body().custom(customAJV(UpdateSchema)),
+    OBJECT: body().isObject(),
```
jonahsnider: Is this validation safe? Seems like users could pass garbage in through request body which gets saved straight to the DB. Also possible is passing something like `{ __proto__: 'garbage' }` and messing with the values at runtime, but the body parser middleware may already take care of that for us.

leartgjoni-voiceflow (approved):
lgtm, the race condition is pretty isolated as well, as it's specific to one user's session

jonahsnider (approved):

lib/controllers/stateManagement/index.ts
DecathectZero: @jonahsnider I think this might need to get covered in a different PR, since in the other state CRUD endpoints we are allowing users to inject whatever they want into the variables as well.

But good point through. Now that you've brought it up someone could easily keep injecting massive objects full of hundreds of thousands of entries and they would all get persisted. So far no malicious use yet.

jonahsnider: Probably very annoying to implement, leaving it as-is will probably be ok. Also, I saw that the validation was changed from `.isObject()` to `.exists()`, which might allow a top-level array in the JSON to be passed in. Maybe change it back?

PR #130, opened 2021-06-09
feat: add buttons (VF-1058)

z4o4z:
**Fixes or implements VF-1058**

Refactor types, add buttons support with fallback to chips

### Depends on PRs

https://github.com/voiceflow/libs/pull/15

### Related PRs

https://github.com/voiceflow/creator-app/pull/3507
https://github.com/voiceflow/general-service/pull/105


leartgjoni-voiceflow (approved):
looks good, I think we might need to communicate to our users that the `TraceType.CHOICE` doesn't have a `chips` property anymore, and it has been replaced with `buttons` (the format has changed as well). Idk if we have anybody using chips tho. @DecathectZero might have more context

lib/services/dialog/utils.ts
```diff
@@ -70,6 +71,9 @@ export const getIntentEntityList = (intentName: string, model: PrototypeModel) =
   return intentEntityIDs?.map((id) => model.slots.find((entity) => entity.key === id));
 };
 
+export const isInteractionsInNode = (node: BaseNode & { interactions?: NodeInteraction[] }): node is BaseNode & { interactions: NodeInteraction[] } =>
+  'interactions' in node && Array.isArray(node.interactions);
```
DecathectZero, line 75: I guess `interactions' in node` is not needed since `Array.isArray(undefined)` is going to be false anyways

lib/services/buttons/utils.ts
```diff
@@ -0,0 +1,88 @@
...
+      if (name) {
+        buttons.push({ name, request: { type: RequestType.TEXT, payload: name } });
+      }
+    }
+  }
+
+  return _.uniqBy(buttons, (button) => button.name);
+};
+
+export const getChoiceButtons = (rawButtons: AnyRequestButton[], model: PrototypeModel): AnyRequestButton[] => {
```
DecathectZero, line 56: do you think we can move all this logic into the compile. I think the less things we do in runtime the better
and we can just attach the utterance to intent.query as well

lib/services/runtime/handlers/interaction.ts
```diff
@@ -26,15 +26,30 @@ export const InteractionHandler: HandlerFactory<Node, typeof utilsObj> = (utils)
     if (runtime.getAction() === Action.RESPONSE) {
       utils.addRepromptIfExists(node, runtime, variables);
 
-      if (!utils.addChipsIfExists(node, runtime, variables)) {
+      if (!utils.addButtonsIfExists(node, runtime, variables)) {
```
DecathectZero, line 29: what if chips always exist, and they are set to default values during the compile?

lib/services/buttons/utils.ts
z4o4z: yeah, good idea, going to try to implement it 👍🏻

DecathectZero (approved):

PR #135, opened 2021-06-11
feat: poc analytics (VF-000)

xavidop:

### Related PRs

| https://github.com/voiceflow/creator-app/pull/3599 | [link](https://github.com/voiceflow/creator-app/pull/3599) |

### Checklist


frankgu968:
Looks good to me overall. I'd move the debug calls to the trace level to avoid polluting the logs when we are actually debugging. The JSON body is quite large and verbose.

lib/clients/analytics.ts
```diff
@@ -0,0 +1,131 @@
...
+import { GeneralTrace } from '@voiceflow/general-types';
+import { AxiosResponse } from 'axios';
+
+import log from '@/logger';
+import { Config } from '@/types';
+
+import IngestApi, { EventsType, InteractBody } from './ingest-client';
+
+export class AnalyticsSystem {
+  private analyticsClient: any;
```
jonahsnider, line 11: We should be doing our best to avoid using `any` types in the source as it makes it very difficult for other developers to maintain in the future.

lib/clients/analytics.ts
```diff
@@ -0,0 +1,131 @@
...
+      properties: {
+        metadata,
+      },
+    };
+    this.analyticsClient.track(interactAnalyticsBody);
+  }
+
+  private createInteractBody(id: string, eventId: string, metadata: any): InteractBody {
+    let sessionId: string;
+    if (metadata.data && metadata.data.reqHeaders && metadata.data.reqHeaders.sessionid) {
```
jonahsnider, line 53: We can use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) to clean up logic like this:
```suggestion
    if (metadata.data?.reqHeaders?.sessionid) {
```

lib/clients/ingest-client.ts
```diff
@@ -0,0 +1,55 @@
...
+  private authorization?: string;
+
+  public constructor(endpoint: string, auhtorization?: string) {
+    super(endpoint);
+    this.authorization = auhtorization;
+
+    this._initializeRequestInterceptor();
+  }
+
+  public static getInstance(endpoint: string, auhtorization: string) {
```
jonahsnider, line 32: This pattern of having a static `getInstance()` method is not really the idiomatic way to work in JS/TS. It's totally fine to just export a single instance of the class if you really need a singleton to work with. [The TypeScript docs explain this in a bit more detail here](https://www.typescriptlang.org/docs/handbook/2/classes.html#why-no-static-classes).

jonahsnider (changes requested):
Looks good overall, just needs some unit tests. There are also a few very minor issues I pointed out. I'd be happy to make those changes if you'd like.

lib/clients/ingest-client.ts
xavidop: fixed!

lib/clients/analytics.ts
xavidop: fixed!

lib/clients/analytics.ts
xavidop: fixed

types.ts
```diff
@@ -86,6 +93,7 @@ export type ContextData = {
   reqHeaders?: {
     authorization?: string;
     origin?: string;
+    sessionid?: string;
```
DecathectZero, line 96: just for minor styling - https://stackoverflow.com/questions/5258977/are-http-headers-case-sensitive#:~:text=Header%20names%20are%20not%20case,%22)%20and%20the%20field%20value.

```suggestion
    sessionId?: string;
```

DecathectZero:
front end component to this PR:
https://github.com/voiceflow/creator-app/pull/3599

jonahsnider (approved):

jonahsnider:
waiting for final approval from @xavidop then we can merge this

PR #139, opened 2021-06-28
fix: vm2 execute (VF-000)

DecathectZero:

I know this looks absolutely disgusting but I couldn't find any other good way. So I believe this is because whenever you pass a non-primitive into VM2, the prototype is completely stripped from it, and I was ending up with a ton of `[Object: null prototype] {}` errors.

example: I pass in a variables `months` that looks like this:
```
months: [
    'January',   'February',
    'March',     'April',
    'May',       'June',
    'July',      'August',
    'September', 'October',
    'November',  'December'
  ],
```
When we try to do `vm._context.months` it comes out like this: 
```
[Array(0): null prototype] []
```
If you stringify and then parse the string, all the data is still there, and prototype is applied again.

~~The reason why this doesn't happen on prod is because we run it on a lambda, which stringifys the JSON response for HTTP, and it gets parsed on the other end by axios. So all we are doing is simulating that process~~

### Related PRs

### Checklist

### DIFF
--- a/runtime/lib/Handlers/code/utils.ts
+++ b/runtime/lib/Handlers/code/utils.ts
@@ -31,10 +31,10 @@ export const vmExecute = (
 
   vm.run(`${safe ? clearContext : ''} ${data.code}`);
 
-  const context = JSON.parse(JSON.stringify(_.get(vm, '_context')));
-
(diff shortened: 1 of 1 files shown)


DecathectZero:
turns out of you run with `NODE_ENV=production` this totally doesn't happen any more....

PR #140, opened 2021-07-05
fix: code handler (VF-1170)

z4o4z:
**Fixes or implements VF-1170**

Get the variables from `vm2` via official API and add prototypes to the incoming variables. Better formating for the changed vars.


leartgjoni-voiceflow (approved):
looks good. tests failing

PR #149, opened 2021-08-02
feat: add `last_utterance` global variable (VF-1162)

jonahsnider:
**Fixes or implements VF-1162**

Add a global `last_utterance` variable.

### Related PRs

- voiceflow/google-runtime#121
- voiceflow/creator-app#3804

### Checklist

### DIFF
--- a/lib/services/runtime/index.ts
+++ b/lib/services/runtime/index.ts
@@ -55,6 +55,7 @@ class RuntimeManager extends AbstractManager<{ utils: typeof utils }> implements
       runtime.trace.debug(`matched intent **${request.payload.intent.name}** - confidence interval _${confidence}%_`);
 
       runtime.variables.set('intent_confidence', Number(confidence));
+      runtime.variables.set(Variables.LAST_UTTERANCE, request.payload.query);
     }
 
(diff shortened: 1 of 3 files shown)


lib/services/runtime/index.ts
```diff
@@ -54,7 +54,8 @@ class RuntimeManager extends AbstractManager<{ utils: typeof utils }> implements
 
       runtime.trace.debug(`matched intent **${request.payload.intent.name}** - confidence interval _${confidence}%_`);
 
-      runtime.variables.set('intent_confidence', Number(confidence));
+      runtime.variables.set(Variables.INTENT_CONFIDENCE, Number(confidence));
+      runtime.variables.set(Variables.LAST_UTTERANCE, request.payload.query);
```
DecathectZero, line 58: only write it if `request.payload.query` exists, I think it is possible to pass in empty queries?

DecathectZero (approved):

lib/services/runtime/index.ts
jonahsnider: typings say it's always defined but just to be safe i'll add a check that query is truthy

PR #151, opened 2021-08-03
fix: safely handle ingest errors (VF-000)

DecathectZero:

we can get rid of these if they never say anything and the ingest is pretty stable. 

### Related PRs

### Checklist

### DIFF
--- a/lib/services/runtime/index.ts
+++ b/lib/services/runtime/index.ts
@@ -5,6 +5,7 @@
 
 import { Event } from '@/lib/clients/ingest-client';
 import { Variables } from '@/lib/services/runtime/types';
+import logger from '@/logger';
 import Client from '@/runtime';
 import { Config, Context, ContextHandler } from '@/types';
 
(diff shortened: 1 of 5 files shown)


xavidop (approved):

lib/services/runtime/index.ts
```diff
@@ -79,7 +80,7 @@ class RuntimeManager extends AbstractManager<{ utils: typeof utils }> implements
       trace: runtime.trace.get(),
     };
 
-    await this.services.analyticsClient!.track({ versionID, event: Event.TURN, metadata, timestamp });
+    await this.services.analyticsClient!.track({ versionID, event: Event.TURN, metadata, timestamp }).catch((error) => logger.error(error));
```
jonahsnider, line 83: this & the other snippet both have a non null assertion on `this.services.analyticsClient!`, is this good? should it be replaced with `?.track()` instead?

jonahsnider (approved):

tests/lib/clients/analytics.unit.ts
```diff
@@ -72,7 +72,7 @@ describe('Analytics client unit tests', () => {
 
       (client as any).rudderstackClient = rudderstack;
 
-      const ingestClient = { doIngest: sinon.stub() };
+      const ingestClient = { doIngest: sinon.stub().resolves({ data: { turn_id: 1 } }) };
```
DecathectZero, line 75: I don't know how tests were passing before, but 
https://github.com/voiceflow/general-runtime/blob/master/lib/clients/analytics.ts#L153 was setting `response` to undefined and we were trying to fetch `response.data.turn_id`

jonahsnider (approved):

PR #166, opened 2021-08-24
feat: do not handle action requests in runtime (VF-1586)

z4o4z:
**Fixes or implements VF-1586**

Do not handle action requests in runtime, but track it in analytics


leartgjoni-voiceflow (changes requested):
looks good, but there's some type issues

leartgjoni-voiceflow (approved):
rebase needed

PR #167, opened 2021-08-25
fix: dm analytics (VF-1593)

DecathectZero:
**Fixes or implements VF-1593**

So originally the analytics was part of the runtime handler, and they would never be hit if the DM handler decided to exit early and skip the next handler in the sequence. 

Now it is in its own dedicated handler. I didn't add it at the very end because we don't want the trace to get polluted by the TTS MP3 and have too much data stored.

### Related PRs

### Checklist

### DIFF
--- a/lib/services/interact.ts
+++ b/lib/services/interact.ts
@@ -19,7 +19,7 @@ class Interact extends AbstractManager {
     query: { locale?: string };
     headers: { authorization?: string; origin?: string; sessionid?: string };
   }) {
-    const { runtime, metrics, nlu, tts, dialog, asr, speak, slots, state: stateManager, filter } = this.services;
+    const { analytics, runtime, metrics, nlu, tts, dialog, asr, speak, slots, state: stateManager, filter } = this.services;
 
     const {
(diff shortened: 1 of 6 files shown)


lib/services/interact.ts
```diff
@@ -46,6 +46,7 @@ class Interact extends AbstractManager {
     const turn = new TurnBuilder<Context>(stateManager);
 
     turn.addHandlers(asr, nlu, slots, dialog, runtime);
+    turn.addHandlers(analytics);
```
jonahsnider, line 49: can we combine these into 1 call?

jonahsnider (approved):

leartgjoni-voiceflow (approved):
nice

lib/services/interact.ts
DecathectZero: no, everything on the same level of `addHandlers` can be broken anywhere in the chain. So dialog can decide to exit early and not hit the runtime handler, but it can't affect and other level of handlers.

This is an array of arrays.

PR #168, opened 2021-08-30
feat: text prompts (VF-000)

z4o4z:

The most important thing to know is this type:
`export type Output = Text.SlateTextValue | string;` The output we use around the place could either be a string or a slate rich text format. This includes prompts, nomatches, etc. A `outputTrace` function is provided to give the appropriate trace based on the output.

Major or breaking changes:
* Remove the `StorageType.OUTPUT` flag entirely. This seemed like relic of Alexa/Google stuff, since general-runtime doesnt depend on the storage.output to compose the response, but relies purely on the Trace. I looked around the code and couldn't find anywhere that `StorageType.OUTPUT` was being read or used. It was only being written to. (On that note, the repeat handler was broken this whole time seems)

* switch Frame Storage `FrameType.SPEAK` to `FrameType.OUTPUT`, this gets used when people are coming out of a command flow, and it repeats that last thing before they went into the command flow. So temporarily might skip a sentence for a few people but I think the risk is really low. This got changed to better reflect that Output can be a variety of types

### Related PRs

### Checklist


DecathectZero (approved):

lib/services/runtime/utils.ts
```diff
@@ -1,15 +1,24 @@
...
-import { Node, Request, Trace } from '@voiceflow/base-types';
-import { replaceVariables, transformStringVariableToNumber } from '@voiceflow/common';
+import { Node, Request, Text, Trace } from '@voiceflow/base-types';
+import { Node as ChatNode } from '@voiceflow/chat-types';
+import { replaceVariables, sanitizeVariables, transformStringVariableToNumber } from '@voiceflow/common';
 import { Node as VoiceNode } from '@voiceflow/voice-types';
+import cuid from 'cuid';
 import _ from 'lodash';
+import _cloneDeepWith from 'lodash/cloneDeepWith';
+import _isString from 'lodash/isString';
```
jonahsnider, line 9: unintentional import of both individual methods and the entire lib?

lib/services/runtime/utils.ts
```diff
@@ -100,3 +123,34 @@ export const addButtonsIfExists = <N extends Request.NodeButton>(node: N, runtim
...
+  variables?: Record<string, unknown>;
+}
+
+export function outputTrace(params: OutputParams<Text.SlateTextValue>): Trace.TextTrace;
+export function outputTrace(params: OutputParams<string>): Trace.SpeakTrace;
+export function outputTrace(params: OutputParams<Output>): Trace.TextTrace | Trace.SpeakTrace;
+export function outputTrace({ output, variables = {} }: OutputParams<Output>) {
+  const sanitizedVars = sanitizeVariables(variables);
+
+  if (output && Array.isArray(output)) {
```
jonahsnider, line 141:
```suggestion
  if (Array.isArray(output)) {
```

jonahsnider (approved):

lib/services/runtime/utils.ts
DecathectZero: yeah, we should be importing specific functions, this can be fixed

leartgjoni-voiceflow (approved):

PR #178, opened 2021-09-16
fix: entities mapping (VF-1765)

DecathectZero:
**Fixes or implements VF-1765**

This is a really legacy issue coming to bite us in the ass. So a long time ago every single interaction step contained the specific mappings from slot to variables. (So you could choose to not map certain ones)
But that's long gone and for backwards compatibility we kept the system. (context.mappings) and they get automatically declared in the compiler in general-service.

The new buttons step compiles into an interaction step but we didn't include these implicit mappings.
I think the fix going forward if there are no mappings, is to just turn every entity/slot in the intent into a variable.

### Related PRs

### Checklist

### DIFF
--- a/lib/services/runtime/handlers/event/index.ts
+++ b/lib/services/runtime/handlers/event/index.ts
@@ -5,6 +5,8 @@ import { Runtime, Store } from '@/runtime';
 
 import { mapEntities } from '../../utils';
 
+const entitiesToMappings = (entities: Request.Entity[]) => entities.map(({ name }) => ({ slot: name, variable: name }));
+
 export const intentEventMatcher = {
   match: (context: {
     runtime: GeneralRuntime;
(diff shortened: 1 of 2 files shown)


lib/services/runtime/handlers/event/index.ts
```diff
@@ -19,7 +21,8 @@ export const intentEventMatcher = {
   sideEffect: (context: { runtime: Runtime<Request.IntentRequest>; event: Node.Utils.IntentEvent; variables: Store }) => {
     // use event slot mappings map request entities to variables
     const request = context.runtime.getRequest() as Request.IntentRequest;
-    context.variables.merge(mapEntities(context.event.mappings || [], request.payload.entities || []));
+    const entities = request.payload.entities || [];
+    context.variables.merge(mapEntities(context.event.mappings || entitiesToMappings(entities), entities));
```
DecathectZero, line 25: so if `context.event.mappings` doesn't exist, use the `request.payload.entities` themselves as the source for the mapping

jonahsnider (approved):

DecathectZero:
TODO: in the future we can just remove mappings from the general-runtime interaction as well.

PR #184, opened 2021-09-30
feat: add no reply response trace (VF-1901)

leartgjoni-voiceflow:
**Fixes or implements VF-1901**

write no reply response to trace

### Checklist

- [x] all the dependencies are upgraded


leartgjoni-voiceflow:
requirements change. this logic is not valid anymore

PR #192, opened 2021-10-05
fix: move Docker build logic out of package.json (VF-000)

jonahsnider:

Docker builds only worked because

1. TypeScript was bundling `package.json` into `build/` because it was being imported in the source code
2. The `build` script in `package.json` was doing things only required for Docker builds (ex. copying `yarn.lock`)

Instead, this moves all Docker-related build scripts into the `Dockerfile`.

### Related PRs

An alternative to #191

### Checklist


xavidop (approved):
we should move this pattern to all the services

PR #215, opened 2021-11-03
feat: Set user_id to UserId for state api calls (VF-2153)

ConnorByers:
**Fixes or implements VF-2153**

Made user_id variable reflect the UserID passed in the network request for state api requests 
(see: https://www.voiceflow.com/api/dialog-manager#operation/stateInteract ).

For first time UserIDs, the `user_id` variable is set in the initial generation of the context. 
See lib/services/state/index.ts -> StateManager -> generate.

For existing userIds to be able to access the `user_id`, the addition of the variable is done in the case of a `LAUNCH` request. I thought it would be overkill to set it for every state api request, so a reset for any user will add the ability to use the `user_id` variable.

### Related PRs

### Checklist

### DIFF
--- a/lib/services/interact.ts
+++ b/lib/services/interact.ts
@@ -1,6 +1,6 @@
 import { Request } from '@voiceflow/base-types';
 
-import { RuntimeRequest } from '@/lib/services/runtime/types';
+import { RuntimeRequest, Variables } from '@/lib/services/runtime/types';
 import { State, TurnBuilder } from '@/runtime';
 import { Context } from '@/types';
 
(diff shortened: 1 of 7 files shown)


ConnorByers:
Only mild testing done due to being blocked to general-service prototype jobs not working for me.

leartgjoni-voiceflow (changes requested):
the API supports the ability to start a conversation with an Intent request (matching one of the open intents as the starting block) hence it can't be set only for launch requests. I think the easiest way is to follow the same pattern we use for the [timestamp variable](https://github.com/voiceflow/general-runtime/blob/75816eeda662edc5cfe459063451442bb82c7d3a/lib/services/runtime/index.ts#L69)

leartgjoni-voiceflow (approved):

PR #225, opened 2021-11-16
fix: fix base flow logic (VF-000)

DecathectZero:

goofed. check the comments

### Related PRs

### Checklist

### DIFF
--- a/lib/services/runtime/init.ts
+++ b/lib/services/runtime/init.ts
@@ -10,9 +10,12 @@ const init = (client: Client) => {
   client.setEvent(EventType.stackDidChange, ({ runtime }) => {
     const top = runtime.stack.top();
 
+    if (!top || top.getProgramID() === runtime.getVersionID()) {
+      return;
+    }
     runtime.trace.addTrace<Node.Flow.TraceFrame>({
       type: Node.Utils.TraceType.FLOW,
-      payload: { diagramID: top?.getProgramID(), name: top?.getName() },
+      payload: { diagramID: top.getProgramID(), name: top.getName() },
     });
   });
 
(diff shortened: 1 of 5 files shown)


lib/services/runtime/init.ts
```diff
@@ -10,9 +10,12 @@ const init = (client: Client) => {
   client.setEvent(EventType.stackDidChange, ({ runtime }) => {
     const top = runtime.stack.top();
 
+    if (!top || top.getProgramID() === runtime.getVersionID()) {
+      return;
+    }
```
DecathectZero, line 15: do not show flow trace if it's the base flow or nothing is in the top

runtime/lib/Runtime/cycleStack.ts
```diff
@@ -13,7 +13,8 @@ const cycleStack = async (runtime: Runtime, depth = 0): Promise<void> => {
     return;
   }
 
-  if (runtime.stack.get(0)?.getProgramID() === runtime.getVersionID()) {
+  if (runtime.stack.top()?.getProgramID() === runtime.getVersionID()) {
+    runtime.stack.flush();
```
DecathectZero, line 17: it's important to flush this because the FE has logic dependant on the stack being empty

runtime/lib/Runtime/cycleStack.ts
```diff
@@ -13,7 +13,8 @@ const cycleStack = async (runtime: Runtime, depth = 0): Promise<void> => {
     return;
   }
 
-  if (runtime.stack.get(0)?.getProgramID() === runtime.getVersionID()) {
```
DecathectZero, line 16: yeah this was horribly wrong `get(0)` will result in this ALWAYS being true, since we can have 100 frames but if the first frame is base, stop everything

PR #227, opened 2021-11-23
feat: no-reply/no-match improvements (VF-1997)

z4o4z:
**Fixes or implements VF-1997**

### Related PRs
https://github.com/voiceflow/libs/pull/115


lib/services/runtime/handlers/noReply.ts
```diff
@@ -0,0 +1,61 @@
...
+    runtime.storage.set(StorageType.NO_REPLIES_COUNTER, noReplyCounter + 1);
+
+    runtime.trace.addTrace<Trace.PathTrace>({
+      type: BaseNode.Utils.TraceType.PATH,
+      payload: { path: 'reprompt' },
+    });
+
+    const output = node.noReply?.randomize
+      ? _.sample<string | Text.SlateTextValue>(nonEmptyNoReplies)
+      : nonEmptyNoReplies?.[runtime.storage.get<NoMatchCounterStorage>(StorageType.NO_REPLIES_COUNTER)! - 1];
```
DecathectZero, line 53:
```suggestion
      : nonEmptyNoReplies?.[noReplyCounter];
```

move line 44 (increment) after this to make it easier to read

DecathectZero (changes requested):
`request: null` might not be the best implementation here, because we've essentially tied `null` to the launch function. Like you mentioned earlier:

https://github.com/voiceflow/general-runtime/blob/a8da9c94b2fcb72687d6ad8fe318c56c43c762e8/runtime/lib/Runtime/index.ts#L148

When you start the prototype tool, it sends a `request: null`, and will ALWAYS trigger the no reply logic. The only alternative I've found to this is changing the line above to just `this.setAction(Action.REQUEST)` and in the start block doing: 
Screen Shot 2021-11-23 at 9 37 54 PM

Although I do worry about the side effects.

DecathectZero:
I'm hoping this fixes a lot of the actions/request/response issues:
https://github.com/voiceflow/general-runtime/pull/228

lib/services/runtime/utils.ts
```diff
@@ -139,11 +125,14 @@ export const addButtonsIfExists = <N extends Request.NodeButton>(node: N, runtim
   }
 };
 
-export const getReadableConfidence = (confidence?: number) => ((confidence ?? 1) * 100).toFixed(2);
+export const getReadableConfidence = (confidence?: number): string => ((confidence ?? 1) * 100).toFixed(2);
 
 export const slateToPlaintext = (content: Text.SlateTextValue = []): string =>
   content.reduce<string>((acc, node) => acc + (SlateText.isText(node) ? node.text : slateToPlaintext(node.children)), '');
 
+export const removeEmptyPrompts = (prompts: Array<Text.SlateTextValue | string>): Array<Text.SlateTextValue | string> =>
```
DecathectZero, line 133: shouldn't this mostly be done on the service?

lib/services/runtime/handlers/noMatch.ts
```diff
@@ -1,57 +1,64 @@
...
-  ) as T | undefined;
+const convertDeprecatedNoMatch = ({ noMatch, elseId, noMatches, randomize, ...node }: NoMatchNode) =>
+  ({
+    noMatch: {
+      prompts: noMatch?.prompts ?? noMatches,
+      randomize: noMatch?.randomize ?? randomize,
+      nodeID: noMatch?.nodeID ?? elseId,
+    },
+    ...node,
+  } as NoMatchNode);
```
DecathectZero, line 26: this is the one hack I have to do....the union type on `NoMatchNode` gets mad :(

lib/services/runtime/utils.ts
z4o4z: we have it in the service too, but on runtime, we are injecting variables, and some prompts can be empty due to empty strings in vars

DecathectZero (approved):

lib/services/runtime/handlers/noMatch.ts
```diff
@@ -1,61 +1,71 @@
...
-    });
+export const NoMatchHandler = (utils: typeof utilsObj) => ({
+  handle: (_node: NoMatchNode, runtime: Runtime, variables: Store) => {
+    const node = convertDeprecatedNoMatch(_node);
+
+    const nonEmptyNoMatches = removeEmptyNoMatches(node);
+
+    const noMatchCounter = runtime.storage.get<NoMatchCounterStorage>(StorageType.NO_MATCHES_COUNTER) ?? 0;
+
+    if (noMatchCounter >= nonEmptyNoMatches.length) {
```
DecathectZero, line 44:
```suggestion
    if (noMatchCounter > nonEmptyNoMatches.length) {
```

lib/services/runtime/handlers/noMatch.ts
```diff
@@ -1,61 +1,71 @@
...
 
-      draft[StorageType.NO_MATCHES_COUNTER] = counter ? counter + 1 : 1;
-    });
+export const NoMatchHandler = (utils: typeof utilsObj) => ({
+  handle: (_node: NoMatchNode, runtime: Runtime, variables: Store) => {
+    const node = convertDeprecatedNoMatch(_node);
+
+    const nonEmptyNoMatches = removeEmptyNoMatches(node);
+
+    const noMatchCounter = runtime.storage.get<NoMatchCounterStorage>(StorageType.NO_MATCHES_COUNTER) ?? 0;
```
DecathectZero, line 42:
```suggestion
    const noMatchCounter = runtime.storage.get<NoMatchCounterStorage>(StorageType.NO_MATCHES_COUNTER) ?? 1;
```

PR #228, opened 2021-11-23
feat: new action system (VF-000)

DecathectZero:

Previously, there were two states: `Action.RESPONSE` and `Action.REQUEST`

When we hit a choice/interaction/prompt/capture node, if it was `Action.RESPONSE` we would stop the execution and wait for the user to give a request. If we were on the same node but it was `Action.REQUEST` we would take the user request and actually compare it against the ports/commands, update the action to `Action.RESPONSE` and continue the execution.

Some problems with that model:
1. handlers had to declare `setAction(Action.RESPONSE)` explicitly
2. https://github.com/voiceflow/general-runtime/blob/a8da9c94b2fcb72687d6ad8fe318c56c43c762e8/runtime/lib/Runtime/index.ts#L148 we've conflated `null` with a launch request. Sending `null` means it is always already handled.

The improvements I'm making:
1. it is no longer the responsibility of handlers to update the action, instead, the cycleHandler does it automatically after the first node is processed.
2. all runtime calls now start in an `Action.REQUEST` state just in case the first handler has use for it, but never after.
3. `Action.RESPONSE` is very ambiguous, `Action.RUNNING` means that it is in standard execution mode.

Explanation of Action States:
`IDLE` - runtime instance created, but hasn't been run yet
`REQUEST` - user request incoming and needs to be handled
`RUNNING` - standard execution
`END` - finished

Each instance of runtime is only allowed to call `update` once. It will only work if its state is in `IDLE`

### Checklist

- [x] this is a breaking change and should publish a new major version

### DIFF
--- a/lib/services/runtime/handlers/state/oneShot.ts
+++ b/lib/services/runtime/handlers/state/oneShot.ts
@@ -11,7 +11,7 @@ const utilsObj = {
 
 export const OneShotIntentHandler: HandlerFactory<Node.Start.Node, typeof utilsObj> = (utils) => ({
   canHandle: (node, runtime) => {
-    return isIntentRequest(runtime.getRequest()) && runtime.stack.getSize() === 1 && node.type === Node.NodeType.START;
+    return isIntentRequest(runtime.getRequest()) && runtime.stack.getSize() <= 2 && node.type === Node.NodeType.START;
   },
   handle: (node, runtime, variables) => {
     if (utils.commandHandler.canHandle(runtime)) {
--- a/lib/services/runtime/handlers/state/oneShot.ts
+++ b/lib/services/runtime/handlers/state/oneShot.ts
@@ -1,6 +1,6 @@
 import { Node } from '@voiceflow/base-types';
 
-import { Action, HandlerFactory } from '@/runtime';
+import { HandlerFactory } from '@/runtime';
 
 import { isIntentRequest } from '../../types';
 import CommandHandler from '../command';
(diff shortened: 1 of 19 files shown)


runtime/lib/Runtime/cycleHandler.ts
```diff
@@ -42,6 +43,11 @@ const cycleHandler = async (runtime: Runtime, program: ProgramModel, variableSta
         await runtime.callEvent(EventType.handlerDidCatch, { error });
       }
 
+      // after the first handler, the request is processed
+      if (runtime.getAction() === Action.REQUEST) {
+        runtime.setAction(Action.RUNNING);
+      }
```
DecathectZero, line 49: after the first handler call, it is guaranteed the Action will not be REQUEST

lib/services/runtime/handlers/state/oneShot.ts
```diff
@@ -11,12 +11,9 @@ const utilsObj = {
 
 export const OneShotIntentHandler: HandlerFactory<Node.Start.Node, typeof utilsObj> = (utils) => ({
   canHandle: (node, runtime) => {
-    return isIntentRequest(runtime.getRequest()) && runtime.stack.getSize() === 1 && node.type === Node.NodeType.START;
+    return isIntentRequest(runtime.getRequest()) && runtime.stack.getSize() <= 2 && node.type === Node.NodeType.START;
```
DecathectZero, line 14: @leartgjoni-voiceflow for the new `baseProgram`, stack is now default size 2?

leartgjoni-voiceflow (approved):
very clean. lgtm

PR #230, opened 2021-12-03
feat: add new local user code runner (VF-2375)

ConnorByers:
**Fixes or implements VF-2375**

Add the use of isolated-vm to be used to run code (provided from code block or IFV2 expressions) if requireFromUrl is not used inside the code.

### Deployment Notes

Isolated-vm requires extra installation steps, mentioned here:
https://github.com/laverdet/isolated-vm#:~:text=Alternatives-,REQUIREMENTS,-This%20project%20requires
So will need to coordinate with infra to add.

### Related PRs

### Checklist

### DIFF
--- a/runtime/lib/Handlers/code/index.ts
+++ b/runtime/lib/Handlers/code/index.ts
@@ -5,7 +5,7 @@ import safeJSONStringify from 'safe-json-stringify';
 
 import { HandlerFactory } from '@/runtime/lib/Handler';
 
-import { vmExecute } from './utils';
+import { ivmExecute, vmExecute } from './utils';
 
 export type CodeOptions = {
   endpoint?: string | null;
@@ -13,7 +13,7 @@ export type CodeOptions = {
   safe?: boolean;
 };
 
-const CodeHandler: HandlerFactory<Node.Code.Node, CodeOptions | void> = ({ endpoint, callbacks, safe } = {}) => ({
+const CodeHandler: HandlerFactory<Node.Code.Node, CodeOptions | void> = ({ endpoint, callbacks } = {}) => ({
   canHandle: (node) => !!node.code,
   handle: async (node, runtime, variables) => {
     try {
@@ -24,7 +24,14 @@ const CodeHandler: HandlerFactory<Node.Code.Node, CodeOptions | void> = ({ endpo
         variables: variablesState,
       };
 
-      const data = endpoint ? (await axios.post(endpoint, reqData)).data : vmExecute(reqData, safe, callbacks);
+      // if requireFromUrl is used, use either axios or vm2
+      // if requireFromUrl is not used, use isolated-vm
+      // eslint-disable-next-line no-nested-ternary
+      const data = node.code.includes('requireFromUrl')
+        ? endpoint
+          ? (await axios.post(endpoint, reqData)).data
+          : vmExecute(reqData, true, callbacks)
+        : ivmExecute(reqData, callbacks);
 
       // debugging changes find variable value differences
       const changes = _.union(Object.keys(variablesState), Object.keys(data)).reduce<string>((acc, variable) => {
--- a/runtime/lib/Handlers/code/index.ts
+++ b/runtime/lib/Handlers/code/index.ts
@@ -10,11 +10,11 @@ import { ivmExecute, vmExecute } from './utils';
 export type CodeOptions = {
   endpoint?: string | null;
   callbacks?: Record<string, (...args: any) => any>;
-  useIVM?: boolean;
+  useStrictVM?: boolean;
   testingEnv?: boolean;
 };
 
-const CodeHandler: HandlerFactory<Node.Code.Node, CodeOptions | void> = ({ endpoint, callbacks, useIVM = false, testingEnv = false } = {}) => ({
+const CodeHandler: HandlerFactory<Node.Code.Node, CodeOptions | void> = ({ endpoint, callbacks, useStrictVM = false, testingEnv = false } = {}) => ({
   canHandle: (node) => !!node.code,
   handle: async (node, runtime, variables) => {
     try {
@@ -26,8 +26,8 @@ const CodeHandler: HandlerFactory<Node.Code.Node, CodeOptions | void> = ({ endpo
       };
 
       let data: Record<string, any>;
-      // useIVM used for IfV2 and SetV2
-      if (useIVM) {
+      // useStrictVM used for IfV2 and SetV2 to use isolated-vm
+      if (useStrictVM) {
         data = await ivmExecute(reqData, callbacks);
       } else {
         data = endpoint ? (await axios.post(endpoint, reqData)).data : vmExecute(reqData, testingEnv, callbacks);
--- a/runtime/lib/Handlers/code/index.ts
+++ b/runtime/lib/Handlers/code/index.ts
@@ -10,10 +10,11 @@ import { ivmExecute, vmExecute } from './utils';
 export type CodeOptions = {
   endpoint?: string | null;
   callbacks?: Record<string, (...args: any) => any>;
-  safe?: boolean;
+  useIVM?: boolean;
+  testingEnv?: boolean;
 };
 
-const CodeHandler: HandlerFactory<Node.Code.Node, CodeOptions | void> = ({ endpoint, callbacks } = {}) => ({
+const CodeHandler: HandlerFactory<Node.Code.Node, CodeOptions | void> = ({ endpoint, callbacks, useIVM = false, testingEnv = false } = {}) => ({
   canHandle: (node) => !!node.code,
   handle: async (node, runtime, variables) => {
     try {
@@ -24,14 +25,13 @@ const CodeHandler: HandlerFactory<Node.Code.Node, CodeOptions | void> = ({ endpo
         variables: variablesState,
       };
 
-      // if requireFromUrl is used, use either axios or vm2
-      // if requireFromUrl is not used, use isolated-vm
-      // eslint-disable-next-line no-nested-ternary
-      const data = node.code.includes('requireFromUrl')
-        ? endpoint
-          ? (await axios.post(endpoint, reqData)).data
-          : vmExecute(reqData, true, callbacks)
-        : ivmExecute(reqData, callbacks);
+      let data: Record<string, any>;
+      // useIVM used for IfV2 and SetV2
+      if (useIVM) {
+        data = await ivmExecute(reqData, callbacks);
+      } else {
+        data = endpoint ? (await axios.post(endpoint, reqData)).data : vmExecute(reqData, testingEnv, callbacks);
+      }
 
       // debugging changes find variable value differences
       const changes = _.union(Object.keys(variablesState), Object.keys(data)).reduce<string>((acc, variable) => {
--- a/runtime/lib/Handlers/code/utils.ts
+++ b/runtime/lib/Handlers/code/utils.ts
@@ -5,56 +5,73 @@ import { VM } from 'vm2';
 
 import Store from '@/runtime/lib/Runtime/Store';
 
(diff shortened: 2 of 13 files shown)


runtime/lib/Handlers/code/utils.ts
```diff
@@ -1,9 +1,56 @@
+import ivm from 'isolated-vm';
 import _ from 'lodash';
 import requireFromUrl from 'require-from-url/sync';
 import { VM } from 'vm2';
 
 import Store from '@/runtime/lib/Runtime/Store';
 
+export const ivmExecute = (data: { code: string; variables: Record<string, any> }, callbacks?: Record<string, (...args: any) => any>) => {
+  // Create isolate with 8mb max memory
+  const isolate = new ivm.Isolate({ memoryLimit: 8 });
```
jonahsnider, line 10: this number seems pretty low, could we bump this to like 64 or 128 MB?

also, please move all limits (memory, execution time, etc) into a top level config object

ex.
```js
const ISOLATED_VM_LIMITS = {
	 maxMemoryMB: 64,
	 maxExecutionTimeMs: 30 * 1000,
}
```

runtime/lib/Handlers/code/utils.ts
```diff
@@ -1,9 +1,56 @@
...
+  // Create a context inside this isolate to run the code in and inject variables/funcs
+  const context = isolate.createContextSync();
+
+  // Get a Reference{} to the global object within the context.
+  const jail = context.global;
+
+  // set Promise to not do anthing in code
+  jail.setSync('Promise', null);
```
jonahsnider, line 18: what does this mean? will users not being to use `Promise.resolve`, `.reject`, etc in their scripts?

runtime/lib/Handlers/code/utils.ts
```diff
@@ -1,9 +1,56 @@
...
+    });
+  }
+
+  // add variables to context
+  Object.keys(data.variables).forEach((key) => {
+    jail.setSync(key, data.variables[key], { copy: true });
+  });
+
+  // run code with 2s timeout
+  context.evalSync(data.code, { timeout: 2000 });
```
jonahsnider, line 36: this number should be extracted into a config object, see above comment

runtime/lib/Handlers/code/utils.ts
```diff
@@ -1,9 +1,56 @@
...
+    if (val instanceof ivm.Reference) {
+      // For non-primitives like arrays or objects
+      val = val.copySync();
+    }
+    acc[key] = Store.formatPayloadValue(val);
+    return acc;
+  }, {} as Record<string, any>);
+
+  // delete isolate to regain memory
+  isolate.dispose();
```
jonahsnider, line 50: if an error occurs during execution what happens? will this `isolate.dispose()` always be reached? or should we wrap all the code in a `try { /* ... */ } finally { isolate.dispose(); }`

runtime/lib/Handlers/code/index.ts
```diff
@@ -24,7 +24,14 @@ const CodeHandler: HandlerFactory<Node.Code.Node, CodeOptions | void> = ({ endpo
         variables: variablesState,
       };
 
-      const data = endpoint ? (await axios.post(endpoint, reqData)).data : vmExecute(reqData, safe, callbacks);
+      // if requireFromUrl is used, use either axios or vm2
+      // if requireFromUrl is not used, use isolated-vm
+      // eslint-disable-next-line no-nested-ternary
+      const data = node.code.includes('requireFromUrl')
```
jonahsnider, line 30: heads up: this is not a very thorough check. users might have renamed the function (false negatives) or just include the string `requireFromUrl` in their source (false positives)
that's fine. the only proper check would be very complicated and would involve checking values at runtime.

runtime/lib/Handlers/code/index.ts
```diff
@@ -24,7 +24,14 @@ const CodeHandler: HandlerFactory<Node.Code.Node, CodeOptions | void> = ({ endpo
...
+        ? endpoint
+          ? (await axios.post(endpoint, reqData)).data
+          : vmExecute(reqData, true, callbacks)
+        : ivmExecute(reqData, callbacks);
```
jonahsnider, line 34: eslint is right, this double ternary is ugly

you should refactor the code to be something like this:

```js
let data: MyData;

if (node.code.includes('requireFromUrl')) {
	if (endpoint) {
		 data = ...;
	} else {
		 data = ...;
	}
} else {
	data = ivmExecute(reqData, callbacks);
}
```

runtime/lib/Handlers/code/utils.ts
DecathectZero: 30 seconds is way too long, we can keep at like 3 seconds max probably

runtime/lib/Handlers/code/index.ts
trs: I don't think we should be actively trying to disable eslint with these comments. We should work with the rules, not around them.

runtime/lib/Handlers/code/utils.ts
trs: Even further, I would imagine this means a user cannot use promises _at all_ in their scripts.

runtime/lib/Handlers/code/utils.ts
```diff
@@ -1,9 +1,56 @@
...
+
+  // Get a Reference{} to the global object within the context.
+  const jail = context.global;
+
+  // set Promise to not do anthing in code
+  jail.setSync('Promise', null);
+
+  // add callbacks if exists
+  if (callbacks) {
+    Object.keys(callbacks).forEach((callbackName) => {
```
trs, line 22: We could look into using `ts-extras` for improved typings: https://github.com/sindresorhus/ts-extras/blob/main/source/object-keys.ts

runtime/lib/Handlers/code/utils.ts
```diff
@@ -1,9 +1,56 @@
...
 import requireFromUrl from 'require-from-url/sync';
 import { VM } from 'vm2';
 
 import Store from '@/runtime/lib/Runtime/Store';
 
+export const ivmExecute = (data: { code: string; variables: Record<string, any> }, callbacks?: Record<string, (...args: any) => any>) => {
+  // Create isolate with 8mb max memory
+  const isolate = new ivm.Isolate({ memoryLimit: 8 });
+  // Create a context inside this isolate to run the code in and inject variables/funcs
+  const context = isolate.createContextSync();
```
trs, line 12: Do we want to execute this all synchronously?

runtime/lib/Handlers/code/utils.ts
trs: Do we need to `release` the context too? Or will disposing the isolate also handle that?

runtime/lib/Handlers/code/utils.ts
ConnorByers: This is just transported functionality from the previous vm2 implementation so this is on prod. From what I've read on random github threads, adding Promises/async makes things complicated.

runtime/lib/Handlers/code/utils.ts
ConnorByers: Context exists within the isolate so I assume that releasing it will destroy the context too. Isolates get picked up by the garbage collector eventually. Calling `.dispose()` just forces it to free the memory immediately.

runtime/lib/Handlers/code/utils.ts
jonahsnider: ah ok. quick googled and found a thread of weird issues with isolated-vm and async stuff: https://github.com/laverdet/isolated-vm/issues/125

must be a v8 limitation or something

ConnorByers:
Difference between isolated-vm and vm2 is minimal but isolated-vm comes out on top.
```
yalo = 'hello';
const x = [];
for (let i = 0; i < 100000; i+=1) {
  x.push(i);
}
x.forEach((idk)=>{});
```
isolated-vm: 13ms
vm2: 21ms

runtime/lib/Handlers/code/utils.ts
ConnorByers: Async is better, updated

runtime/lib/Handlers/code/utils.ts
```diff
@@ -1,13 +1,77 @@
...
+    // add variables to context
+    Promise.all(
+      Object.keys(data.variables).map(async (key) => {
+        await jail.set(key, data.variables[key], { copy: true });
+      })
+    ).catch((err) => {
+      throw new Error(err);
+    });
+
+    // run code with 2s timeout
```
trs, line 49: Comment lies, I think it's safe to remove it.

runtime/lib/Handlers/code/utils.ts
```diff
@@ -1,13 +1,77 @@
...
+  const isolate = new ivm.Isolate({ memoryLimit: ISOLATED_VM_LIMITS.maxMemoryMB });
+  let ret = {};
+  try {
+    // Create a context inside this isolate to run the code in and inject variables/funcs
+    const context = await isolate.createContext();
+
+    // Get a Reference{} to the global object within the context.
+    const jail = context.global;
+
+    // set Promise to not do anthing in code
```
trs, line 24: Adding a "why" might help future code archeologists.

runtime/lib/Handlers/code/utils.ts
```diff
@@ -1,13 +1,77 @@
...
+        await jail.set(key, data.variables[key], { copy: true });
+      })
+    ).catch((err) => {
+      throw new Error(err);
+    });
+
+    // run code with 2s timeout
+    await context.eval(data.code, { timeout: ISOLATED_VM_LIMITS.maxExecutionTimeMs });
+
+    ret = await Object.keys(data.variables).reduce(async (accPromise, key) => {
```
trs, line 52: No need to set this variable outside the `try`, you can return here inside the `try`. `finally` will always run.

runtime/lib/Handlers/code/utils.ts
```diff
@@ -1,13 +1,77 @@
...
+    // add callbacks if exists
+    if (callbacks) {
+      Promise.all(
+        Object.keys(callbacks).map(async (callbackName) => {
+          await jail.set(callbackName, (...args: any[]) => {
+            callbacks[callbackName](...args);
+          });
+        })
+      ).catch((err) => {
+        throw new Error(err);
```
trs, line 36: We aren't `await`-ing these `Promise.all` calls, so when we throw here we could be outside the execution context. Not sure, but that may produce an unhandled rejection error.

It seems to me that we'd want to be `await`-ing these, avoiding the catch.

If we want them to be in the "background", we shouldn't be throwing.

runtime/lib/Handlers/code/utils.ts
```diff
@@ -1,13 +1,77 @@
...
 import { VM } from 'vm2';
 
 import Store from '@/runtime/lib/Runtime/Store';
 
+const ISOLATED_VM_LIMITS = {
+  maxMemoryMB: 64,
+  maxExecutionTimeMs: 3 * 1000,
+};
+
+export const ivmExecute = async (data: { code: string; variables: Record<string, any> }, callbacks?: Record<string, (...args: any) => any>) => {
```
trs, line 13: Just curious, any reason `callbacks` is separate from the other of the parameters? Why not:
```js
ivmExecute({
  code: ``,
  variables: {},
  callbacks: {}
})
```

ConnorByers: I took this structure from the previous implementation but here's my interpretation: In the POST call with axios, callbacks aren't used and we pass in the the `reqData` object which is also passed to the `ivmExecute` and `vmExecute`. That way the same object is passed to all three and if callbacks are needed, they're passed in seperately.

trs (approved):

jonahsnider (approved):

PR #234, opened 2021-12-13
feat: api block localization (VF-2376)

ConnorByers:
**Fixes or implements VF-2376**

Change api block implementation to make calls within general-service vs custom-api. 

We use redis to keep track out of outgoing calls and throttle if exceeds limits.
We limit the response size of api calls to prevent massive calls through the request config in axios.

### Related PRs

### Checklist


ConnorByers:
Some things for debate:
-Limit api call amounts for private/public
-time to reset tracking in redis
-throttling time amount

runtime/lib/Runtime/OutgoingApiLimiter/index.ts
```diff
@@ -0,0 +1,40 @@
...
+    return !this.runtime.authorization;
+  }
+
+  private makeRedisHostnameHash(hostname: string): string {
+    return `${this.REDIS_PREFIX}_${this.isPublic() ? 'public' : this.runtime.authorization!}_${hostname}`;
+  }
+
+  private async getHostnameUses(hostname: string): Promise<string | null> {
+    return this.runtime.services.redis.get(this.makeRedisHostnameHash(hostname));
+  }
```
jonahsnider, line 24: instead of writing `Number(getHostnameUses())` you should just alter this function to return `number | null` directly

```suggestion
  private async getHostnameUses(hostname: string): Promise<number | null> {
    const uses = await this.runtime.services.redis.get(this.makeRedisHostnameHash(hostname));
    
    return uses ? Number(uses) : null;
  }
```

runtime/lib/Runtime/index.ts
```diff
@@ -46,18 +47,22 @@ class Runtime<R extends any = any, DA extends DataAPI = DataAPI> extends Abstrac
...
+  // eslint-disable-next-line max-params
   constructor(
     public versionID: string,
     state: State,
     private request: R | null = null,
     { services = {}, handlers = [], api }: Options<DA>,
-    events: Lifecycle
+    events: Lifecycle,
+    public authorization: string | undefined
   ) {
```
jonahsnider, line 66: i think eslint is right here, is it possible to combine anything here into an object?

runtime/lib/Handlers/api/utils.ts
```diff
@@ -43,90 +59,160 @@ export const getVariable = (path: string, data: any) => {
...
+      data.mapping.forEach((m) => {
+        if (m.var) {
+          newVariables[m.var] = getVariable(m.path, response.data);
+        }
+      });
+    }
+    return { variables: newVariables, response };
+  } catch (e) {
+    log.info(e.stack);
+    throw new Error();
```
jonahsnider, line 216: this is still kinda strange. is there a reason we can't just do

```suggestion
    throw e;
```

runtime/lib/Handlers/api/utils.ts
```diff
@@ -1,12 +1,28 @@
...
+import ipRangeCheck from 'ip-range-check';
 import _ from 'lodash';
 import querystring from 'querystring';
+import validator from 'validator';
+
+import log from '@/logger';
+import Runtime from '@/runtime/lib/Runtime';
 
 export type APINodeData = Node.Api.NodeData['action_data'];
+const PROHIBITED_IP_RANGES = ['10.0.0.0/8', '172.16.0.0/12', '192.168.0.0/16', '127.0.0.0/8', '0.0.0.0/8', 'fd00::/8', '169.254.169.254/32'];
```
trs, line 15: Are these typical "local" IPs?

runtime/lib/Handlers/api/utils.ts
```diff
@@ -43,90 +59,160 @@ export const getVariable = (path: string, data: any) => {
...
+        throw new VError(`cannot resolve hostname: ${hostname}`, VError.HTTP_STATUS.BAD_REQUEST);
+      });
+
+  PROHIBITED_IP_RANGES.forEach((prohibitedRange) => {
+    ips.forEach((ip) => {
+      if (ipRangeCheck(ip, prohibitedRange)) {
+        throw new VError(`url resolves to IP: ${ip} in prohibited range`, VError.HTTP_STATUS.BAD_REQUEST);
+      }
+    });
+  });
```
trs, line 89: I think using some more specific array methods can help clarify this piece of code:

```suggestion
  const badIP = ips.find((ip) => PROHIBITED_IP_RANGES.some((range) => ipRangeCheck(ip, range)));
  if (badIP) {
    throw new VError(`url resolves to IP: ${badIP} in prohibited range`, VError.HTTP_STATUS.BAD_REQUEST);
  }
```

runtime/lib/Handlers/api/utils.ts
```diff
@@ -43,90 +59,160 @@ export const getVariable = (path: string, data: any) => {
...
+    content = JSON.parse(data.content);
+  } catch (e) {
+    ({ content } = data);
+  }
+
+  BLACKLISTED_URLS.forEach((regex) => {
+    if (regex.test(url)) {
+      throw new VError('url endpoint is blacklisted', VError.HTTP_STATUS.BAD_REQUEST);
+    }
+  });
```
trs, line 112: `Array.prototype.some` could be used here too.

```suggestion
  if (BLACKLISTED_URLS.some((regex) => regex.test(url))) {
    throw new VError('url endpoint is blacklisted', VError.HTTP_STATUS.BAD_REQUEST);
  }
```

runtime/lib/Handlers/api/utils.ts
```diff
@@ -43,90 +59,160 @@ export const getVariable = (path: string, data: any) => {
...
 
-  if (params?.length > 0) {
-    const formattedParams = ReduceKeyValue(params);
+  if (params && params.length > 0) {
+    const formattedParams: Record<string, any> = {};
+    params.forEach((p) => {
+      if (p.key) {
+        formattedParams[p.key] = p.val;
+      }
+    });
```
trs, line 132: Can build an object using `Object.fromEntries`.

```suggestion
    const formattedParams = Object.fromEntries(params.map((param) => [param.key, param.val]));
```

runtime/lib/Handlers/api/utils.ts
```diff
@@ -43,90 +59,160 @@ export const getVariable = (path: string, data: any) => {
...
   }
 
   if (headers && headers.length > 0) {
-    const formattedHeaders = ReduceKeyValue(headers);
+    const formattedHeaders: Record<string, any> = {};
+    headers.forEach((h) => {
+      if (h.key) {
+        formattedHeaders[h.key] = h.val;
+      }
+    });
```
trs, line 142:
```suggestion
    const formattedHeaders = Object.fromEntries(headers.map((header) => [header.key, header.value]));
```

runtime/lib/Handlers/api/utils.ts
```diff
@@ -43,90 +59,160 @@ export const getVariable = (path: string, data: any) => {
...
-      options.data = JSON.parse(content);
-    } catch (e) {
-      options.data = data;
-    }
-  } else if (bodyInputType === Node.Api.APIBodyType.FORM_DATA) {
-    const formData = new FormData();
-    body.forEach((b) => {
-      if (b.key) {
-        formData.append(b.key, b.val);
+  if (method !== 'GET') {
```
trs, line 147: Looking at the previous code, these changes seem to make this less readable and more confusing. I'd recommend taking another go at this.

runtime/lib/Handlers/api/utils.ts
```diff
@@ -43,90 +59,160 @@ export const getVariable = (path: string, data: any) => {
...
+  try {
+    const response = await getResponse(data, runtime);
+    const newVariables: Record<string, any> = {};
+    if (data.mapping) {
+      data.mapping.forEach((m) => {
+        if (m.var) {
+          newVariables[m.var] = getVariable(m.path, response.data);
+        }
+      });
+    }
```
trs, line 212: Trying to utilize the array methods better.

```suggestion
    const newVariables = Object.fromEntries((data.mapping ?? [])
      .filter((map) => map.var)
      .map((map) => [map.var, getVariable(m.path, response.data)])
    );
```

trs (changes requested):

runtime/lib/Handlers/api/utils.ts
jonahsnider: i agree. is it possible to split this up into a bunch of smaller functions?

ConnorByers: I copied all of this code directly from `custom-api`. The version existing in `general-runtime` previously was just a simplified version by Leart I believe. I'll break it up though

runtime/lib/Handlers/api/utils.ts
ConnorByers: I took these directly from `custom-api`. Some are localhost and some are private ip ranges.

runtime/lib/Handlers/api/utils.ts
```diff
@@ -51,16 +66,59 @@ export const ReduceKeyValue = (values: { key: string; val: string }[]) =>
...
+  }
+};
+
+// eslint-disable-next-line sonarjs/cognitive-complexity
+export const formatRequestConfig = async (data: APINodeData, runtime: Runtime) => {
   const { method, bodyInputType, headers, body, params, url, content } = data;
 
+  if (BLACKLISTED_URLS.some((regex) => regex.test(url))) {
+    throw new VError('url endpoint is blacklisted', VError.HTTP_STATUS.BAD_REQUEST);
+  }
```
trs, line 107: This should probably go in the `validateUrl` function, yeah?

runtime/lib/Handlers/api/utils.ts
```diff
@@ -107,26 +162,29 @@ export const formatRequestConfig = (data: APINodeData) => {
...
+    const options = await formatRequestConfig(nodeData, runtime);
 
-  const { data, headers, status } = await axios(requestConfig);
+    const { data, headers, status } = await axios(options);
 
-  if (_.isObject(data) as any) {
-    data.VF_STATUS_CODE = status;
-    data.VF_HEADERS = headers;
-  }
+    if (typeof data === 'object' && options !== null) {
```
trs, line 179: Be aware that `data` maybe not be a record since other types are "objects".

```
> typeof null
'object'
> typeof []
'object'
```

runtime/lib/Handlers/api/utils.ts
```diff
@@ -51,16 +66,59 @@ export const ReduceKeyValue = (values: { key: string; val: string }[]) =>
...
+
+  const badIP = ips.find((ip) => PROHIBITED_IP_RANGES.some((range) => ipRangeCheck(ip, range)));
+  if (badIP) {
+    throw new VError(`url resolves to IP: ${badIP} in prohibited range`, VError.HTTP_STATUS.BAD_REQUEST);
+  }
+
+  if (await runtime.outgoingApiLimiter.addHostnameUseAndShouldThrottle(hostname)) {
+    // if the use of the hostname is high, delay the api call but let it happen
+    await new Promise((resolve) => setTimeout(resolve, THROTTLE_DELAY));
+  }
```
trs, line 98: Instead of checking this here, we could consider this function a pure type validator, and move it out to the `makeAPICall` function. Since this isn't so much validating as looking up and delaying. That way this function doesn't have to be async too.
What do you think?

ConnorByers: Taking it out makes sense, the function will remain async because of the code block inside:
```
// DNS.resolve returns an array of ips
  const ips = validator.isIP(hostname)
    ? [hostname]
    : await DNS.resolve(hostname).catch(() => {
        throw new VError(`cannot resolve hostname: ${hostname}`, VError.HTTP_STATUS.BAD_REQUEST);
      });
```

trs: Good point.
Maybe we should move that out too? It kinda seems like a separate concern.

Steps seem to be:

- Get IP from hostname
- Validate IP
- Make request to IP
  - Rate limit request to IP

Really, each could be their own function.

trs (approved):

runtime/lib/Runtime/OutgoingApiLimiter/index.ts
```diff
@@ -0,0 +1,42 @@
...
+  private EXPIRY_LENGTH = 60;
+
+  constructor(private runtime: Runtime) {}
+
+  private isPublic(): boolean {
+    return !this.runtime.authorization;
+  }
+
+  private makeRedisHostnameHash(hostname: string): string {
+    return `${this.REDIS_PREFIX}_${this.isPublic() ? 'public' : this.runtime.authorization!}_${hostname}`;
```
DecathectZero, line 19: I think maybe we just use `versionID` and have the same number for both public and private?

PR #238, opened 2021-12-14
feat: dialog manager api key support (VF-2429)

trs:
**Fixes or implements VF-2429**

Adds support for the new API key format. New routes to interact with the API without providing a `versionID`. This is supported by using the API key to look up the latest version.

### DIFF
--- a/lib/middlewares/project.ts
+++ b/lib/middlewares/project.ts
@@ -46,7 +46,7 @@ class Project extends AbstractMiddleware {
 
         const project = await api.getProjectUsingAuthorization(req.headers.authorization);
         req.headers.project_id = project._id.toString();
-        req.params.versionID = project.liveVersion?.toString() ?? project.devVersion?.toString();
+        req.params.versionID = project.devVersion!.toString();
         return next();
       }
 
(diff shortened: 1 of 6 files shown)


lib/middlewares/project.ts
```diff
@@ -22,10 +26,30 @@ const VALIDATIONS = {
...
+      req.params.versionID = (req.headers.versionID as string) ?? req.params.versionID;
+
+      // Facilitate supporting routes that require a versionID but do not have to supply one.
+      // We can use the provided API key to look up the project and grab the latest version.
+      if (!req.params.versionID && typeof req.headers.authorization === 'string') {
+        if (!(api instanceof CreatorDataApi)) {
+          throw new VError('Only supported via Creator Data API');
+        }
+
+        const project = await api.getProjectUsingAuthorization(req.headers.authorization);
```
DecathectZero, line 47: I wonder if we need to pass the entire project object in every time if we're only looking for the `devVersion` or specific fields in the project. But that's something we can think about later

lib/middlewares/project.ts
```diff
@@ -22,10 +26,30 @@ const VALIDATIONS = {
...
+        req.headers.project_id = project._id.toString();
+        req.params.versionID = project.liveVersion?.toString() ?? project.devVersion?.toString();
```
DecathectZero, line 49: I think always `devVersion` by default

lib/middlewares/project.ts
trs: We could definitely optimize later to have a specific route that only returns the dev version instead of the entire project.

DecathectZero (approved):

PR #248, opened 2021-12-19
feat: auto delegate go-to (VF-000)

DecathectZero:

So what's going on here.
Right now, when we send a `GOTO` trace with a request, we expect the client calling the webhook to interpret/parse the `GOTO` and send a subsequent follow-up request to the webhook.

Now instead, with `autoDelegate`, the runtime can self-consume the GOTO trace and call itself again.

This removes the onus on the developer using the API to create special logic for the GOTO, although they still can by setting `config.selfDelegate = true`! So it's a win-win.

### Related PRs

### Checklist


lib/services/filter/index.ts
```diff
@@ -37,7 +37,7 @@ class Filter extends AbstractManager<{ utils: typeof utils }> implements Context
               ...trace,
               payload: {
                 ...trace.payload,
-                message: sanitizeSSML(trace.payload.message),
+                message: this.services.utils.sanitizeSSML(trace.payload.message),
```
DecathectZero, line 40: not really related to the PR but we should've actually be doing dependency injection here

lib/services/interact/index.ts
```diff
@@ -18,35 +31,30 @@ class Interact extends AbstractManager {
...
-    } = req;
-    // `request` prop is deprecated, replaced with `action`
-    // Internally the name request is still used
-    request = action ?? request;
-
-    if (request?.type === Request.RequestType.LAUNCH && state) {
-      state.stack = [];
-      state.storage = {};
-      request = null;
-    }
```
DecathectZero, line 42: moved this to `/lib/services/state`

lib/services/interact/index.ts
```diff
@@ -18,35 +31,30 @@ class Interact extends AbstractManager {
...
       query: { locale },
       headers: { authorization, origin, sessionid: sessionId },
     } = req;
 
-    let {
-      body: { request = null },
-    } = req;
-    // `request` prop is deprecated, replaced with `action`
-    // Internally the name request is still used
-    request = action ?? request;
```
DecathectZero, line 36: unified this to line 51 on the updated code.

lib/services/interact/autoDelegate.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (!context.trace) break;
+
+    const [filteredTrace, goToTrace] = filterAndGetLastRemovedValue(context.trace, (frame) => !isGoToTrace(frame));
+    trace.push(...filteredTrace);
+
+    if (isGoToTrace(goToTrace)) {
+      context.request = goToTrace.payload.request;
+    } else {
+      break;
+    }
```
DecathectZero, line 41: if there is a `GOTO` trace, keep iterating, otherwise, stop.

lib/services/interact/autoDelegate.ts
```diff
@@ -0,0 +1,46 @@
...
+  let lastItem: A | null = null;
+  const filteredList = list.filter((a) => {
+    if (filterFunc(a)) return true;
+    lastItem = a;
+    return false;
+  });
+  return [filteredList, lastItem];
+};
+
+const isGoToTrace = (frame?: Trace.AnyTrace | null): frame is Trace.GoToTrace => frame?.type === Trace.TraceType.GOTO && !!frame.payload.request;
```
jonahsnider, line 19: `Trace.AnyTrace | null | undefined` is kinda ick. is there any difference between `null` and `undefined` here?

lib/services/interact/autoDelegate.ts
```diff
@@ -0,0 +1,46 @@
...
+    trace.push(...filteredTrace);
+
+    if (isGoToTrace(goToTrace)) {
+      context.request = goToTrace.payload.request;
+    } else {
+      break;
+    }
+
+    iterations++;
+  } while (iterations < MAX_DELEGATION_TURNS);
```
jonahsnider, line 41: this seems like it's just a DIY for loop

lib/services/interact/autoDelegate.ts
```diff
@@ -0,0 +1,46 @@
+import { Trace } from '@voiceflow/base-types';
+
+import { PartialContext, TurnBuilder } from '@/runtime';
+import { Context } from '@/types';
+
+export const MAX_DELEGATION_TURNS = 3;
+
+// mostly just saves us needing to traverse an array twice
+const filterAndGetLastRemovedValue = <A>(list: A[], filterFunc: (a: A) => boolean): [A[], A | null] => {
```
jonahsnider, line 9: not important but typescript convention is to use `T` as the name of generic types

also,
```suggestion
// mostly just saves us needing to traverse an array twice
// TODO: move to @voiceflow/common
const filterAndGetLastRemovedValue = <A>(list: A[], filterFunc: (a: A) => boolean): [A[], A | null] => {
```

or add a jira ticket

jonahsnider (approved):

lib/services/interact/autoDelegate.ts
DecathectZero: it is, but I think given the context a `do...while` makes more sense, because `iterations` is not the main focus. It could easily `break` on the first iteration.

I could convert it to a for loop, wdyt readability wise.

jonahsnider: could you convert it to a `for` loop please?
nothing wrong with having it `break` early 😛 

it's nice to keep `iterations` local to the loop instead of declaring it on L22 in the function body.
seeing that + a `do ... while` makes me expect some kind of complex logic involving the iteration count

lib/services/interact/autoDelegate.ts
DecathectZero: https://github.com/voiceflow/libs/commit/ee55d9963b2f491696ab8643e78f166b9c6134e3

lib/services/interact/autoDelegate.ts
DecathectZero: yeah `undefined` is not needed here

lib/services/interact/autoDelegate.ts
DecathectZero: one thing is that in the dowhile loop, typescript resolves `context = ....` while in the for loop format it is still nullish and causes an error
Screen Shot 2022-01-03 at 3 58 35 PM
:

jonahsnider: yeah that's unfortunate. seems to be since the `while` block allows type narrowing like `if` statements

DecathectZero: I can either add `!` in the for loop or stick to the `do while` wdyt

jonahsnider: keep the non-null assertion & for loop imo

jonahsnider: btw, you can get a similar type narrowing feature if you do [this](https://www.typescriptlang.org/play?#code/DYUwLgBAdgrgtgLmvARiAThAPs4wIC8uwA3AFBkBmA9pgBSiQCWhEADCRCwDwQDM2HLDiECRWHk5MA1NICUEAN5kIq5CKIBGcgF8KAY2pQAztVAA6YNQDmdYRGkRNckkA)

you should not do that though it's very ugly

package.json
```diff
@@ -14,12 +14,12 @@
   "dependencies": {
     "@types/mathjs": "^6.0.4",
     "@types/workerpool": "^5.0.1",
-    "@voiceflow/api-sdk": "3.3.5",
-    "@voiceflow/backend-utils": "3.3.0",
+    "@voiceflow/api-sdk": "3.4.5",
+    "@voiceflow/backend-utils": "4.2.2",
```
jonahsnider, line 18: CI is failing because of the breaking change in backend-utils

fix it by reverting or by updating the ratelimit code in this repo to be compatible

trs (approved):

PR #251, opened 2022-01-03
feat: refactor rate limit (VF-000)

DecathectZero:

there's only like a million places this could go wrong 😅
`project_id` => `projectID`
`project_id` was not something expected from the actual request, it is just gets added through a middleware.

Outwith `req.params.versionID` it's legacy now and uses the `unifyVersionID` middleware to translate it. 
All hail `req.headers.versionID`. 

I've done some manual testing and it seems to work.

remove rate limiter service, use the new class instead.

I do think that this should get removed out of backend-utils:
https://github.com/voiceflow/backend-utils/blob/master/src/middlewares/rateLimit.ts#L41-L48
It's too specific to `general-runtime`. Each service should have their own consumption model.

### Related PRs

### Checklist


lib/controllers/interact.ts
```diff
@@ -4,23 +4,23 @@
  */
 
 import { Request as BaseRequest } from '@voiceflow/base-types';
-import { Request } from 'express';
 
 import { RuntimeRequest } from '@/lib/services/runtime/types';
 import { State } from '@/runtime';
+import { Request } from '@/types';
```
DecathectZero, line 10: the `Request` from `@/types` allows us to define the header

backend/docs/openapi.yaml
```diff
@@ -1,6 +1,6 @@
 openapi: '3.0.3'
 info:
-  version: 1.0.0
+  version: 1.0.1
```
DecathectZero, line 3: this should've been added on @trs 's docs update PR:
https://github.com/voiceflow/general-runtime/commit/3388b1702c794b5a8b33b8d27965cd325a1eef58

jonahsnider (approved):

trs (commented):
I don't _love_ using the nouns "version" and "project" to reference the IDs. To me the names imply the object, not just the identifier.

We're already coercing the version ID with the `unifyVersionID` middleware, we could just coerce into `versionID` instead of `version`. Same could be done with project.

Express has [`req.get`](https://expressjs.com/en/api.html#req.get) to case-insensitively get a header, maybe we should be using that throughout instead of referencing the key directly? Otherwise `req.headers` keys are lower-cased (see https://nodejs.org/api/http.html#messageheaders), so can be referenced with `req.headers.versionid` regardless of what the user set.

Another option is to have a global middleware that overrides `req.headers` that uses a Proxy to case-insensitively get headers, eg:
```js
req.headers = new Proxy(req.headers, {
  get(_, name) {
    return req.get(name);
  }
});
```

DecathectZero:
> I don't _love_ using the nouns "version" and "project" to reference the IDs. To me the names imply the object, not just the identifier.
> 
> We're already coercing the version ID with the `unifyVersionID` middleware, we could just coerce into `versionID` instead of `version`. Same could be done with project.
> 
> Express has [`req.get`](https://expressjs.com/en/api.html#req.get) to case-insensitively get a header, maybe we should be using that throughout instead of referencing the key directly? Otherwise `req.headers` keys are lower-cased (see https://nodejs.org/api/http.html#messageheaders), so can be referenced with `req.headers.versionid` regardless of what the user set.
> 
> Another option is to have a global middleware that overrides `req.headers` that uses a Proxy to case-insensitively get headers, eg:
> 
> ```js
> req.headers = new Proxy(req.headers, {
>   get(_, name) {
>     return req.get(name);
>   }
> });
> ```

The other issue is that express-validator only seems to force everything lower-case as well. https://github.com/voiceflow/general-runtime/blob/aefa17b99457781899100a6d9a39a41438551780/lib/controllers/stateManagement/index.ts#L17

trs:
> The other issue is that express-validator only seems to force everything lower-case as well.
> 
> https://github.com/voiceflow/general-runtime/blob/aefa17b99457781899100a6d9a39a41438551780/lib/controllers/stateManagement/index.ts#L17

Looks like it [references `req.headers`](https://express-validator.github.io/docs/check-api.html#headerfields-message), so the proxy would work here.

backend/docs/openapi.yaml
```diff
@@ -130,7 +130,7 @@ tags:
 
       By default, requests will target the dev version of your project. To specify a version, such as the live version, provide a `{versionID}` header.
 
-      ### versionID
+      ### version
```
trs, line 133:
```suggestion
      ### versionID
```

lib/middlewares/project.ts
```diff
@@ -1,21 +1,17 @@
...
+const { header } = Validator;
 const VALIDATIONS = {
-  PARAMS: {
-    VERSION_ID: param('versionID')
-      .optional()
-      .isString(),
-  },
   HEADERS: {
-    VERSION_ID: header('versionID')
+    VERSION: header('versionID')
```
trs, line 14:
```suggestion
    VERSION_ID: header('versionID')
```

Would have to update references too.

trs (approved):

PR #255, opened 2022-01-05
feat: add config env vars for API block (VF-000)

Matthew-Smith:

Because you provide the runtime as a service to run for partners, it is ideal to be able to configure aspects of it without making code changes. This change allows us to modify the API block configuration via environment variables instead of directly in the code.

Add environment variables to the config.ts and use in utils.ts

### Checklist


config.ts
```diff
@@ -32,6 +32,9 @@ const CONFIG: Config = {
   INTEGRATIONS_HANDLER_ENDPOINT: getOptionalProcessEnv('INTEGRATIONS_HANDLER_ENDPOINT') || 'none',
   // api-block
   API_HANDLER_ENDPOINT: getOptionalProcessEnv('API_HANDLER_ENDPOINT'),
+  API_MAX_TIMEOUT_MS: Number(getOptionalProcessEnv('API_MAX_TIMEOUT_MS', (20 * 1000).toString())),
+  API_MAX_CONTENT_LENGTH_BYTES: Number(getOptionalProcessEnv('API_MAX_CONTENT_LENGTH_BYTES', '1000000')),
+  API_MAX_BODY_LENGTH_BYTES: Number(getOptionalProcessEnv('API_MAX_BODY_LENGTH_BYTES', '1000000')),
```
DecathectZero, line 37:
```suggestion
  API_MAX_TIMEOUT_MS: Number(getOptionalProcessEnv('API_MAX_TIMEOUT_MS')) || 20000,
  API_MAX_CONTENT_LENGTH_BYTES: Number(getOptionalProcessEnv('API_MAX_CONTENT_LENGTH_BYTES') || 1000000,
  API_MAX_BODY_LENGTH_BYTES: Number(getOptionalProcessEnv('API_MAX_BODY_LENGTH_BYTES') || 1000000,
```

DecathectZero (approved):
makes sense!

config.ts
Matthew-Smith: **note:** I did it this way because further up in the file there is 

```typescript
  ERROR_RESPONSE_MS: Number(getOptionalProcessEnv('ERROR_RESPONSE_MS', (10 * 1000).toString())),
```
and I just wanted to follow any of your conventions in your codebase but I'm happy to change it.

One thing to note is that `||` doesn't handle `'0'` so if I were to set `API_MAX_TIMEOUT_MS="0"` in my env it would use the fallback. That is probably okay but just want to make sure you're aware!

DecathectZero: we can just switch over to `??` in that case. Yeah I noticed our convention as well

config.ts
```diff
@@ -32,6 +32,9 @@ const CONFIG: Config = {
   INTEGRATIONS_HANDLER_ENDPOINT: getOptionalProcessEnv('INTEGRATIONS_HANDLER_ENDPOINT') || 'none',
   // api-block
   API_HANDLER_ENDPOINT: getOptionalProcessEnv('API_HANDLER_ENDPOINT'),
+  API_MAX_TIMEOUT_MS: Number(getOptionalProcessEnv('API_MAX_TIMEOUT_MS')) || 20000,
+  API_MAX_CONTENT_LENGTH_BYTES: Number(getOptionalProcessEnv('API_MAX_CONTENT_LENGTH_BYTES') || 1000000,
+  API_MAX_BODY_LENGTH_BYTES: Number(getOptionalProcessEnv('API_MAX_BODY_LENGTH_BYTES') || 1000000,
```
DecathectZero, line 37: sorry my commit suggestion was malformed
```suggestion
  API_MAX_CONTENT_LENGTH_BYTES: Number(getOptionalProcessEnv('API_MAX_CONTENT_LENGTH_BYTES')) ?? 1000000,
  API_MAX_BODY_LENGTH_BYTES: Number(getOptionalProcessEnv('API_MAX_BODY_LENGTH_BYTES')) ?? 1000000,
```

DecathectZero:
sorry, I tried again on my end and I think your original PR using the original convention worked best. 

Because `Number(undefined)` gives `NaN` and `NaN ?? 10` gives `NaN` and you made the point about `||` ofuscating the 0.

PR #260, opened 2022-01-12
feat: entity filling incoming intent (VF-000)

DecathectZero:

After extensive testing with our dialog service (this should be renamed to entity filling), I decided to make some changes that better reflect the user's intentions.

We make too much use of the fallback intent. It should be up to the NLP to resolve something as a "None Intent", otherwise whatever else it resolves to is still valid data that can be made use of.

I've simplified the cases for entity filling. Suppose we are trying to fill out a pizza type for "pizza intent" with `size` & `type` entities: 
```
-> "what type of pizza do you want?" (Sys)
<- "help" (User)
```

* We will first send just "help" to LUIS to match, which will easily result in the "help intent". (incoming intent)
* Next, we send `${dm_prefix_hash} help`, which will more often than not, result in `${dm_prefix_hash} intent` with no entities filled out. The prefix highly biases the NLP model and will have false positives.

With this update, if the `${dm_prefix_hash} intent` contains no entities that the intent needs (i.e. `size` & `type`), we should just disregard it and use the incoming intent instead. 
This is made with the assumption that ALL `${dm_prefix_hash} intent` utterances **should** contain a relevant entity of some kind.

This results in simpler logic and more concise classification, making it easier to break out of slot filling and into other intents.

I've also updated the docs here to reflect the new states better.
https://www.notion.so/voiceflow/LUIS-Dialogue-Management-86549a90f3984f21a48ed6aebbdaa8f8

Tried it on a flow like this and it works quite as intended:
Screen Shot 2022-01-12 at 12 06 36 AM

### Related PRs

### Checklist

### DIFF
--- a/lib/services/dialog/index.ts
+++ b/lib/services/dialog/index.ts
@@ -55,7 +55,7 @@ class DialogManagement extends AbstractManager<{ utils: typeof utils }> implemen
     const dmPrefixedResultName = dmPrefixedResult.payload.intent.name;
     log.trace(`[app] [runtime] [dm] DM-Prefixed inference result ${log.vars({ resultName: dmPrefixedResultName })}`);
 
-    if (dmPrefixedResultName.startsWith(VF_DM_PREFIX)) {
+    if (dmPrefixedResultName.startsWith(VF_DM_PREFIX) || dmPrefixedResultName === dmStateStore.intentRequest?.payload.intent.name) {
       // Remove hash prefix entity from the DM-prefixed result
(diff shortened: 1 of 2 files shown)


frankgu968:
Looks good to me! Is there any way we can get some tracking done on the branches to see what percentage of requests are going through each of the cases? More rigorous approach to pruning the unused branches imo.

frankgu968 (approved):

DecathectZero:
> Looks good to me! Is there any way we can get some tracking done on the branches to see what percentage of requests are going through each of the cases? More rigorous approach to pruning the unused branches imo.

certainly, can add that in as another ticket. Maybe something like segement is good for this.

PR #261, opened 2022-01-13
[Merged by Bors] - ci: new cicd system

xavidop:
**Fixes or implements VF-2657**

Implements [this](https://www.notion.so/voiceflow/CI-CD-System-f80ef4dd66b8407ba23f7a72e097b96d)

### Checklist


xavidop:
bors try

xavidop:
bors try

.circleci/config.yml
```diff
@@ -44,190 +59,158 @@ defaults:
           branch_pattern: master
 
 orbs:
-  vfcommon: voiceflow/common@0.0.270
+  vfcommon: voiceflow/common@dev:8c5ebfe
```
frankgu968, line 62: fix this once the orb is published

Dockerfile
```diff
@@ -1,18 +1,3 @@
...
-
-RUN apk add --no-cache python3 make g++
-
-RUN echo $NPM_TOKEN > .npmrc && \
-  yarn install && \
-  yarn build && \
-  rm -rf build/node_modules && \
-  rm -f .npmrc
-
 FROM node:16-alpine
```
frankgu968, line 1: it just occurred to me: if we are using node:16-alpine as the runtime container, would the built artifacts work? because the build might be done on a different environment

frankgu968 (changes requested):

Dockerfile
xavidop: the docker image we are using for build it's using the same docker image: 168387678261.dkr.ecr.us-east-1.amazonaws.com/ci-node-build-image:v1

https://github.com/voiceflow/infrastructure/blob/master/utils/circleci/Dockerfile.node.build

.circleci/config.yml
xavidop: yup

frankgu968 (approved):

PR #264, opened 2022-01-13
fix: api injection (VF-000)

DecathectZero:

I saw this error after bumping the general-runtime package for alexa:
```
Error: env var: GENERAL_SERVICE_ENDPOINT not found
```
the alexa-runtime definitely doesn't need the `GENERAL_SERVICE_ENDPOINT` to work. Because we referenced `@/config` directly, the downstream runtimes now need the same environment variables.

Stems from: https://github.com/voiceflow/general-runtime/pull/255

Instead, they should be instantiated as the handlers get being added.
This is blocking some critical bug fixes for alexa.

`API_HANDLER_ENDPOINT` is no longer needed, this was a reference to the API microservice. It is done locally now.

### Related PRs

### Checklist

### DIFF
--- a/config.ts
+++ b/config.ts
@@ -31,10 +31,9 @@ const CONFIG: Config = {
   // integrations block
   INTEGRATIONS_HANDLER_ENDPOINT: getOptionalProcessEnv('INTEGRATIONS_HANDLER_ENDPOINT') || 'none',
   // api-block
-  API_HANDLER_ENDPOINT: getOptionalProcessEnv('API_HANDLER_ENDPOINT'),
-  API_MAX_TIMEOUT_MS: Number(getOptionalProcessEnv('API_MAX_TIMEOUT_MS', '20000')),
-  API_MAX_CONTENT_LENGTH_BYTES: Number(getOptionalProcessEnv('API_MAX_CONTENT_LENGTH_BYTES', '1000000')),
-  API_MAX_BODY_LENGTH_BYTES: Number(getOptionalProcessEnv('API_MAX_BODY_LENGTH_BYTES', '1000000')),
+  API_MAX_TIMEOUT_MS: Number(getOptionalProcessEnv('API_MAX_TIMEOUT_MS')) || null,
+  API_MAX_CONTENT_LENGTH_BYTES: Number(getOptionalProcessEnv('API_MAX_CONTENT_LENGTH_BYTES')) || null,
+  API_MAX_BODY_LENGTH_BYTES: Number(getOptionalProcessEnv('API_MAX_BODY_LENGTH_BYTES')) || null,
 
   PROJECT_SOURCE: getOptionalProcessEnv('PROJECT_SOURCE'),
 
--- a/runtime/lib/Handlers/api/index.ts
+++ b/runtime/lib/Handlers/api/index.ts
@@ -5,15 +5,11 @@ import safeJSONStringify from 'safe-json-stringify';
 
 import { HandlerFactory } from '@/runtime/lib/Handler';
 
-import { APINodeData, makeAPICall } from './utils';
-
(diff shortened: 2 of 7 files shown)


config.ts
```diff
@@ -31,10 +31,9 @@ const CONFIG: Config = {
   // integrations block
   INTEGRATIONS_HANDLER_ENDPOINT: getOptionalProcessEnv('INTEGRATIONS_HANDLER_ENDPOINT') || 'none',
   // api-block
-  API_HANDLER_ENDPOINT: getOptionalProcessEnv('API_HANDLER_ENDPOINT'),
-  API_MAX_TIMEOUT_MS: Number(getOptionalProcessEnv('API_MAX_TIMEOUT_MS', '20000')),
-  API_MAX_CONTENT_LENGTH_BYTES: Number(getOptionalProcessEnv('API_MAX_CONTENT_LENGTH_BYTES', '1000000')),
-  API_MAX_BODY_LENGTH_BYTES: Number(getOptionalProcessEnv('API_MAX_BODY_LENGTH_BYTES', '1000000')),
+  API_MAX_TIMEOUT_MS: Number(getOptionalProcessEnv('API_MAX_TIMEOUT_MS')) || null,
```
DecathectZero, line 34: `Number(null | undefined)` results in `NaN` we still want `null`

runtime/lib/Handlers/api/index.ts
```diff
@@ -26,7 +22,7 @@ const APIHandler: HandlerFactory<Node.Integration.Node, IntegrationsOptions | vo
         actionBodyData.headers = [...headers, { key: USER_AGENT_KEY, val: USER_AGENT }];
       }
 
-      const data = await makeAPICall(actionBodyData, runtime);
+      const data = await makeAPICall(actionBodyData, runtime, config as ResponseConfig);
```
DecathectZero, line 25: so it won't resolve `undefined` with `void` so I have to cast it.

jonahsnider, line 25: this type assertion scares me, should `makeAPICall` be updated to allow `Partial<ResponseConfig>`? i think that would fix the types here

tests/runtime/lib/Handlers/api.unit.ts
```diff
@@ -40,8 +38,8 @@ describe('API Handler unit tests', () => {
       sinon.restore();
     });
     it('success', async () => {
-      const customAPIEndpoint = 'https://foo';
-      const apiHandler = APIHandler({ customAPIEndpoint });
+      const config = 'config';
```
jonahsnider, line 41: surely this should be `{}` and not a string

jonahsnider (approved):

tests/runtime/lib/Handlers/api.unit.ts
DecathectZero: it doesn't matter too much, I'm just checking that whatever I set it to is getting passed directly into the function and not getting modified.

runtime/lib/Handlers/api/utils.ts
```diff
@@ -9,7 +9,6 @@ import querystring from 'querystring';
 import safeJSONStringify from 'safe-json-stringify';
 import validator from 'validator';
 
-import CONFIG from '@/config';
```
DecathectZero, line 12: this was breaking downstream runtimes

config.ts
jonahsnider: yeah JS is funny in that sense.
null gets casted to 0, undefined gets casted to NaN

we should really come up with some util for converting optional env vars to numbers

trs: @DecathectZero @jonahsnider This package is super helpful for reading and converting environment variables: https://www.npmjs.com/package/env-var

PR #276, opened 2022-01-28
fix: store the last visited node, check on interact (VF-2489)

trs:
**Fixes or implements VF-2489**

This fixes an infinite loop when a choice entity is the same as the intent it is going to.

We store the last visited node on every runtime loop. This enables us to check that we aren't trying to handle the interaction that we just triggered a GOTO from.

### DIFF
--- a/runtime/lib/Runtime/cycleHandler.ts
+++ b/runtime/lib/Runtime/cycleHandler.ts
@@ -1,5 +1,6 @@
 /* eslint-disable no-await-in-loop */
 
+import { StorageType } from '@/lib/services/runtime/types';
 import { Action } from '@/runtime';
 import { EventType } from '@/runtime/lib/Lifecycle';
 import ProgramModel from '@/runtime/lib/Program';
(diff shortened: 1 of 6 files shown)


runtime/lib/Runtime/cycleHandler.ts
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable no-await-in-loop */
 
+import { StorageType } from '@/lib/services/runtime/types';
```
DecathectZero, line 3: I wouldn't import anything from `@/lib` in `/runtime` since the `/runtime` is meant as an isolated package that gets implemented by other services like alexa or google.
`/lib/services/runtime` is general's implementation of runtime.

This is a better place for global level storage variables:
https://github.com/voiceflow/general-runtime/blob/edf93d2d947cd58985f59d6f39eb396959f55ea8/runtime/lib/Constants/flags.ts#L1-L4

DecathectZero, line 3: I wouldn't import anything from `@/lib` in `/runtime` since the `/runtime` is meant as an isolated package that gets implemented by other services like alexa or google.
`/lib/services/runtime` is general's implementation of runtime.

This is a better place for global level storage variables:
https://github.com/voiceflow/general-runtime/blob/edf93d2d947cd58985f59d6f39eb396959f55ea8/runtime/lib/Constants/flags.ts#L1-L4

DecathectZero (changes requested):

DecathectZero (approved):

PR #279, opened 2022-02-01
feat: track goto node to avoid infinite loops (VF-2489)

trs:
**Fixes or implements VF-2489**

When a GOTO is triggered from an interaction, we keep track of the node that we are originating from. If next interaction is the same node, we don't attempt to process it. This fixes an infinite loop when a choice entity is the same as the intent it is going to.

This works both on the project that originated from this error, as well as the project that was broken after the previous "fix" (#276)

### DIFF
--- a/lib/services/runtime/handlers/interaction.ts
+++ b/lib/services/runtime/handlers/interaction.ts
@@ -22,10 +22,6 @@ const utilsObj = {
   addNoReplyTimeoutIfExists,
 };
 
-const isIntentEvent = (event: BaseNode.Utils.BaseEvent): event is BaseNode.Utils.IntentEvent => {
-  return 'goTo' in event;
-};
-
 export const InteractionHandler: HandlerFactory<GeneralNode.Interaction.Node | ChatNode.Interaction.Node, typeof utilsObj> = (utils) => ({
   canHandle: (node) => !!node.interactions,
   handle: (node, runtime, variables) => {
@@ -55,7 +51,7 @@ export const InteractionHandler: HandlerFactory<GeneralNode.Interaction.Node | C
         // allow handler to apply side effects
         matcher.sideEffect();
 
-        if (isIntentEvent(event) && event.goTo != null) {
+        if (BaseNode.Utils.isIntentEvent(event) && event.goTo != null) {
           const { request } = event.goTo!;
           runtime.trace.addTrace<Trace.GoToTrace>({
             type: BaseNode.Utils.TraceType.GOTO,
--- a/lib/services/runtime/handlers/interaction.ts
+++ b/lib/services/runtime/handlers/interaction.ts
@@ -11,6 +11,7 @@ import { findEventMatcher } from './event';
 import NoMatchHandler from './noMatch';
 import NoReplyHandler, { addNoReplyTimeoutIfExists } from './noReply';
 import RepeatHandler from './repeat';
+import { isIntentEvent } from './utils';
 
 const utilsObj = {
   repeatHandler: RepeatHandler(),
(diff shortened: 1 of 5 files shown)


lib/services/runtime/handlers/interaction.ts
```diff
@@ -41,21 +41,25 @@ export const InteractionHandler: HandlerFactory<GeneralNode.Interaction.Node | C
       return utils.noReplyHandler.handle(node, runtime, variables);
     }
 
-    for (let i = 0; i < node.interactions.length; i++) {
-      const { event, nextId } = node.interactions[i];
+    if (runtime.storage.get(StorageType.GOTO_FROM_NODE_ID) !== node.id) {
+      for (let i = 0; i < node.interactions.length; i++) {
+        const { event, nextId } = node.interactions[i];
```
jonahsnider, line 46: this could be a `.forEach()`

lib/services/runtime/handlers/interaction.ts
```diff
@@ -41,21 +41,25 @@ export const InteractionHandler: HandlerFactory<GeneralNode.Interaction.Node | C
       return utils.noReplyHandler.handle(node, runtime, variables);
     }
 
-    for (let i = 0; i < node.interactions.length; i++) {
-      const { event, nextId } = node.interactions[i];
+    if (runtime.storage.get(StorageType.GOTO_FROM_NODE_ID) !== node.id) {
```
jonahsnider, line 44: this `if` statement is a [negated condition](https://eslint.org/docs/rules/no-negated-condition)

lib/services/runtime/handlers/interaction.ts
```diff
@@ -41,21 +41,25 @@ export const InteractionHandler: HandlerFactory<GeneralNode.Interaction.Node | C
...
-      const matcher = utils.findEventMatcher({ event, runtime, variables });
+        const matcher = utils.findEventMatcher({ event, runtime, variables });
+        if (!matcher) continue;
 
-      if (matcher) {
         // allow handler to apply side effects
         matcher.sideEffect();
 
         if ((event as BaseNode.Utils.IntentEvent).goTo) {
+          const { request } = (event as BaseNode.Utils.IntentEvent).goTo!;
```
jonahsnider, line 55: instead of repeating this assertion twice i would recommend doing something like this at the beginning of the loop:

```ts
const { event: _event } = node.interactions[i];

const event = _event as BaseNode.Utils.IntentEvent;
```

jonahsnider (approved):

lib/services/runtime/handlers/interaction.ts
```diff
@@ -67,6 +71,8 @@ export const InteractionHandler: HandlerFactory<GeneralNode.Interaction.Node | C
 
         return nextId || null;
       }
+    } else {
+      runtime.storage.delete(StorageType.GOTO_FROM_NODE_ID);
```
DecathectZero, line 75: we don't need the `else` right.
`runtime.storage.delete(StorageType.GOTO_FROM_NODE_ID);` is fine to run every single time?

trs: Right 👍🏻

lib/services/runtime/handlers/interaction.ts
trs: Can't exit early from a `forEach` unfortunately.

lib/services/runtime/handlers/interaction.ts
trs: I think a type guard is actually a better pattern here. I'll add one.

lib/services/runtime/handlers/interaction.ts
jonahsnider: oop, forgot about that

lib/services/runtime/handlers/interaction.ts
trs: No longer the case as Ty pointed out we don't need the `else`

lib/services/runtime/handlers/interaction.ts
trs: It's okay, I switched it to a `for of` loop until I realized we needed the index 😆

DecathectZero (approved):

PR #284, opened 2022-02-09
[Merged by Bors] - feat: goto with diagram and refactor commands (VF-2887)

DecathectZero:

Making commands a little more readable


runtime/lib/Util/index.ts
```diff
@@ -1,26 +0,0 @@
-import { BaseModels } from '@voiceflow/base-types';
-
-import Stack from '@/runtime/lib/Runtime/Stack';
-
-export type CommandMatcher<C extends BaseModels.BaseCommand = BaseModels.BaseCommand> = (command: C, match?: any) => boolean;
-
-// eslint-disable-next-line import/prefer-default-export
-export const extractFrameCommand = <C extends BaseModels.BaseCommand = BaseModels.BaseCommand>(
```
DecathectZero, line 8: moved all this into `getCommand` to make things more readable

lib/services/runtime/handlers/goTo.ts
```diff
@@ -0,0 +1,36 @@
...
+import { Action } from '@/runtime/lib/Runtime';
+
+import CommandHandler from './command';
+
+const utilsObj = {
+  commandHandler: CommandHandler(),
+};
+
+const GoToHandler: HandlerFactory<BaseNode.GoTo.Node, typeof utilsObj> = (utils) => ({
+  canHandle: (node) => node.type === BaseNode.NodeType.GOTO,
```
z4o4z, line 13: should we add `isGoToNode` typeguard to libs and use it here?

lib/services/runtime/handlers/event/index.ts
```diff
@@ -7,55 +7,77 @@ import { mapEntities } from '../../utils';
...
+export interface Matcher<
+  Request extends BaseRequest.BaseRequest<any> = BaseRequest.BaseRequest<any>,
+  Event extends BaseNode.Utils.BaseEvent = BaseNode.Utils.BaseEvent
+> {
+  match: (context: MatchContext) => context is SideEffectContext<Request, Event>;
+  sideEffect: (context: SideEffectContext<Request, Event>) => (variables: Store) => void;
+}
+
+export const intentEventMatcher: Matcher<BaseRequest.IntentRequest, BaseNode.Utils.IntentEvent> = {
+  match: (context): context is SideEffectContext<BaseRequest.IntentRequest, BaseNode.Utils.IntentEvent> => {
```
trs, line 32: Do we need this assertion here since it is declared in the `Matcher` interface?

trs (approved):
Having a goto node clarifies this behaviour so much better. Nice job!

lib/services/runtime/handlers/goTo.ts
DecathectZero: I think we need to do this for every node, I wouldn't do it as part of this PR

lib/services/runtime/handlers/event/index.ts
DecathectZero: yeah we don't it will be a lot more readable. I will update this :)

DecathectZero: Screen Shot 2022-02-10 at 4 56 53 PM

@trs turns our typescript doesn't like this and won't resolve `: boolean` to typeguard interfaces?

PR #295, opened 2022-02-22
[Merged by Bors] - fix: remove state from transcript (VF-2873)

DecathectZero:
**Fixes or implements VF-2873**

As part of the test of removing the `state` column from `transcripts_turns` I'm deleting the entire column on my dev postgres, removing this write reference, and testing that everything still works fine.


xavidop (approved):
it should be good, you have to remove it from alexa and google runtime

DecathectZero:
> it should be good, you have to remove it from alexa and google runtime

yup need to merge this first because the other services depend on the types defined here.

PR #303, opened 2022-03-10
[Merged by Bors] - feat: add exception handler middleware (VF-000)

trs:

Add the new exception handler middleware for better and more accurate error responses.

Depends on: https://github.com/voiceflow/backend-utils/pull/70

### DIFF
--- a/lib/middlewares/project.ts
+++ b/lib/middlewares/project.ts
@@ -56,7 +56,7 @@ class Project extends AbstractMiddleware {
       }
 
       if (!req.headers.versionID) {
-        throw new Error();
+        throw new VError('Missing version-id header', 400);
       }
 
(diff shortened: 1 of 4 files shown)


lib/middlewares/project.ts
```diff
@@ -56,7 +56,7 @@ class Project extends AbstractMiddleware {
       }
 
       if (!req.headers.versionID) {
-        throw new Error();
+        throw new VError('Missing version-id header', 400);
```
trs, line 59: There are entire pages in datadog with `no permissions for this version at Project.attachID` 500 errors. This should quiet most of those.

jonahsnider: why would anyone throw a plain Error with no message...😖

jonahsnider (approved):

PR #310, opened 2022-04-18
[Merged by Bors] - fix: invalid api-key error (VF-2840)

DecathectZero:
**Fixes or implements VF-2840**

If the API key is not a valid JWT, don't throw a 500 error, instead a 401 unauthorized.

### DIFF
--- a/lib/middlewares/project.ts
+++ b/lib/middlewares/project.ts
@@ -34,14 +34,16 @@ class Project extends AbstractMiddleware {
     HEADERS_AUTHORIZATION: VALIDATIONS.HEADERS.AUTHORIZATION,
   })
   async attachID(req: Request<Record<string, unknown>, unknown, { versionID?: string }>, _res: Response, next: NextFunction): Promise<void> {
-    const api = await this.services.dataAPI.get(req.headers.authorization);
+    const api = await this.services.dataAPI.get(req.headers.authorization).catch((error) => {
+      throw new VError(`invalid API key: ${error}`, VError.HTTP_STATUS.UNAUTHORIZED);
+    });
 
     try {
       // Facilitate supporting routes that require a versionID but do not have to supply one.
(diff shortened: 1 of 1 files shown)


lib/middlewares/project.ts
```diff
@@ -34,14 +34,16 @@ class Project extends AbstractMiddleware {
     HEADERS_AUTHORIZATION: VALIDATIONS.HEADERS.AUTHORIZATION,
   })
   async attachID(req: Request<Record<string, unknown>, unknown, { versionID?: string }>, _res: Response, next: NextFunction): Promise<void> {
-    const api = await this.services.dataAPI.get(req.headers.authorization);
+    const api = await this.services.dataAPI.get(req.headers.authorization).catch((error) => {
+      throw new VError(`invalid API key: ${error}`, VError.HTTP_STATUS.UNAUTHORIZED);
```
trs, line 38: Since this callback isn't async, it'll throw an exception, not a rejected promise.

trs (changes requested):

lib/middlewares/project.ts
DecathectZero: Which is the intended behavior right? The exception will stop further execution
Screen Shot 2022-04-18 at 10 36 33 AM

trs: I suppose the express chain will catch it. It just means this function could reject or throw, which might be confusing outside of an express context.

trs (approved):

lib/middlewares/project.ts
DecathectZero: it will always throw, but yeah it's a standard pattern for us to just throw a `VError` to trigger the desired HTTP error code. 
The other alternative would've been an even larger `try` `catch` but that leads to nested `try` `catch`es which isn't prefered.

PR #311, opened 2022-04-22
[Merged by Bors] - fix: replace variables [bugfix] (VF-3384)

DecathectZero:
**Fixes or implements VF-3384**

This is quite terrible and is meant to be a temporary.
There should be no APL in the base types or as part of the prototype tool. The Alexa project should convert it into a standard visual step.

for example this is how the stream step references a MP3 URL: https://github.com/voiceflow/general-runtime/blob/4e455014623d48d788339953af62c4de2260c367/lib/services/runtime/handlers/stream.ts#L18

Once this is merged I'll come up with a larger refactor to consoliate the visual types in general

### DIFF
--- a/lib/services/runtime/handlers/visual.ts
+++ b/lib/services/runtime/handlers/visual.ts
@@ -13,7 +13,7 @@ const VisualHandler: HandlerFactory<BaseNode.Visual.Node, typeof handlerUtils> =
   handle: (node, runtime, variables) => {
     runtime.trace.debug('__visual__ - entered', BaseNode.NodeType.VISUAL);
 
-    if (node.data.visualType === BaseNode.Visual.VisualType.APL) {
+    if (node.data.visualType === BaseNode.Visual.VisualType.APL && node.data.imageURL) {
       node.data.imageURL = utils.replaceVariables(node.data.imageURL, variables.getState());
     }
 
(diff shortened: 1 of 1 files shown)


lib/services/runtime/handlers/visual.ts
```diff
@@ -1,13 +1,22 @@
...
+
+const VisualHandler: HandlerFactory<BaseNode.Visual.Node, typeof handlerUtils> = (utils) => ({
   canHandle: (node) => node.type === BaseNode.NodeType.VISUAL && !!node.data,
 
-  handle: (node, runtime) => {
+  handle: (node, runtime, variables) => {
     runtime.trace.debug('__visual__ - entered', BaseNode.NodeType.VISUAL);
 
+    if (node.data.visualType === BaseNode.Visual.VisualType.APL) {
+      node.data.imageURL = utils.replaceVariables(node.data.imageURL, variables.getState());
```
trs, line 17: Should we check if the `imageURL` is a variable before calling this? Or default back to `imageURL` if this returns nothing.

trs: I suppose the replace should just return the original if it's not matched. So should be a non-issue.

trs (approved):

PR #314, opened 2022-05-12
[Merged by Bors] - feat: modify analytics service call to new ingest service [bugfix] (VF-3505)

pmvrmc:
**Fixes or implements VF-3505**

We need to change the general-runtime to make the calls go to the new event-ingestion service, modifying the payload to support it

Adapting the analytics client to call the new endpoint with the correct payload

### Deployment Notes

we need to set the INGEST_WEBHOOK_ENDPOINT properly to the new service

### Related PRs

- https://github.com/voiceflow/database-cli/pull/193
- https://github.com/voiceflow/alexa-runtime/pull/290
- https://github.com/voiceflow/google-runtime/pull/209
- https://github.com/voiceflow/creator-api/pull/988
- https://github.com/voiceflow/creator-app/pull/5454

### Checklist
- [ ] this is a breaking change and should publish a new major version ??


pmvrmc:
If we are not using a feature_flag, I guess this might be is a breaking change no? The current ingest service would not be able to handle this new payload forma

lib/clients/analytics.ts
```diff
@@ -25,130 +22,87 @@ export class AnalyticsSystem extends AbstractClient {
...
+    // add milliseconds to maintain order
+    const unixTime = timestamp.getTime() + 1;
 
-        type: (trace ?? request)?.type ?? 'launch',
-        payload: trace ?? request ?? {},
-        format,
-        timestamp: timestamp.toISOString(),
-      },
-    } as GeneralInteractBody;
+    return fullTrace.map((trace, idx) => ({
```
pmvrmc, line 37: simpler logic because launch is set directly before this runs

lib/clients/analytics.ts
```diff
@@ -25,130 +22,87 @@ export class AnalyticsSystem extends AbstractClient {
...
+      metadata: {
+        end: metadata.end,
+        locale: metadata.data.locale,
       },
+      traces: [
+        // launch trace
+        {
+          type: 'launch',
+          payload: {},
+          format: metadata.request ? 'request' : 'launch',
```
pmvrmc, line 74: not entirely sure, but looking at previous logic this seem like a valid case. is it?

lib/services/analytics.ts
```diff
@@ -8,9 +10,21 @@ class Analytics extends AbstractManager implements ContextHandler {
...
 
-    this.services.analyticsClient?.track({ versionID, event: Event.TURN, metadata: context, timestamp: new Date() }).catch((error) => {
-      log.error(`[analytics] failed to track ${log.vars({ versionID, error })}`);
-    });
+    this.services.mongo?.db
+      .collection('versions')
+      .findOne<{ projectID: string }>({ _id: new ObjectId(versionID) })
+      .then((version) =>
+        this.services.analyticsClient?.track({
+          projectID: version!.projectID,
```
pmvrmc, line 18: we should be safe here because a version shoudl always have a projectID, right?

lib/clients/analytics.ts
trs: I don't think it makes sense to hard-code the launch type. An interaction could skip the launch all together.
It might make more sense to have a function that formats the `request` similar to the `createTraceBody`.

Do we want to include the `request` in the traces at all? My understanding is that traces are for the actions the project took based on the steps. Maybe it should be a separate item in the interaction body?

lib/services/analytics.ts
trs: Ideally database lookups should be abstracted to some sort of model. The business logic shouldn't depend on the database implemented. But, doesn't look like we have any patterns set out in this repository, this kind of separation of concerns is something to keep in mind.

A cool pattern would be to have an event queue for these logs, where we can fire a log event and have a service listen for them for formatting and persisting. It would remove the need for the ingest client here, making responses faster since it wouldn't have to wait for the ingest.

Nothing actionable here, just some thoughts that spurred.

pmvrmc: yeah I agree with you on the separation of concers, but since there was no model pattern on the project (and in other places the db was being used like this) I went for it.

Regarding the queue side of ir. This pattern sort of acts that way, we no longer wait on the ingest client response to continue processing the response. The runtime builds the trace/log/transcripts and sends it in a fire and forget manner (we only log in case of an error). Since it's plain http we dont have any persistent thing, but if we wanted its a matter of chaning the ingestClient and ingest service to support for it.

lib/clients/analytics.ts
pmvrmc: from what I could tell from the code, we always hardcoded a launch interaction with an empty payload. it doesnt matter what the actual trace is, the first item is alway a launch one.

The current logic is to fallback to pushing the `request` as the trace payload wthen there is no trace. tbh I dont even know how the `request` format could be trigered. My thought was that we always have a `trace` but to handle some legacy thing we still look for `request`. Not sure, will need to dig in.

trs (approved):

lib/services/interact/index.ts
```diff
@@ -30,7 +30,7 @@ class Interact extends AbstractManager<{ utils: typeof utils }> {
     params: { userID?: string };
     body: { state?: State; action?: RuntimeRequest; request?: RuntimeRequest; config?: BaseRequest.RequestConfig };
     query: { locale?: string };
-    headers: { authorization?: string; origin?: string; sessionid?: string; versionID: string; platform?: string };
+    headers: { authorization?: string; origin?: string; sessionid?: string; versionID: string; projectID: string; platform?: string };
```
trs, line 33: We can't rely on the `projectID` being on the request. We only look it up when given a DM API key (`middlewares.project.attachID`), and we don't specify that consumers send it.

trs: Can we go forward without the `projectID`?

lib/clients/analytics.ts
DecathectZero: I think we should just have a straight-up separate `action` property in the document for requests. So then it would just be:
```
{
  ...
  action: {
    type: "launch"
  },
  traces: [
   ....
  ],
  ...
}
```

lib/services/interact/index.ts
DecathectZero: otherwise I think we just extract the `projectID` from the` versionID` when we fetch the version. We ALWAYS fetch the version every turn in `general-runtime`, and it is cached for the entire request:
https://github.com/voiceflow/general-runtime/blob/master/lib/services/state/index.ts#L69

pmvrmc: I removed the projectID from the headers and kept in only on the handle(context) bit

trs (approved):

PR #330, opened 2022-06-09
feat: add verification for google-adapter calls (VF-3675)

pmvrmc:
**Fixes or implements VF-3675**

Handles validation of google-adapter calls

### Related PRs
- https://github.com/voiceflow/adapters/pull/15


trs (changes requested):
I think we should outline a system for this instead of just passing it as a config. This may require some overarching auth/api key/permission changes.

There is nothing stopping a request from just _not_ sending the `config.adapter` and avoiding this all together.

We should come up with a technical design for this first.

trs:
As discussed, we can have the channel pass the api key to the adapter using a query parameter. That key can be used with the adapter instead of an env variable.

PR #335, opened 2022-06-14
[Merged by Bors] - feat: create carousel handler (VF-3748)

viniciusdacal:
**Fixes or implements VF-3748**

Added a new runtime handler to support carousel step.

With this particular step, we need to support a blocking flow, when the doesn't have a next port, and also a regular flow.
Each button inside the carousel, can trigger a command, and it should also have the possibility to add a no reply and no match.


lib/services/runtime/handlers/carousel.ts
```diff
@@ -0,0 +1,91 @@
...
+          title: replaceVariables(card.title, variablesMap),
+          description: {
+            slate,
+            text,
+          },
+          buttons: card.buttons.map((button) => ({
+            ...button,
+            name: replaceVariables(button.name, variablesMap),
+          })),
+        } as BaseNode.Carousel.TraceCarouselCard;
```
jonahsnider, line 52: why is this assertion necessary?

lib/services/runtime/handlers/carousel.ts
```diff
@@ -0,0 +1,91 @@
...
+
+import { StorageType } from '../types';
+import { slateInjectVariables, slateToPlaintext } from '../utils';
+import CommandHandler from './command';
+import { findEventMatcher } from './event';
+import NoMatchHandler from './noMatch';
+import NoReplyHandler, { addNoReplyTimeoutIfExists } from './noReply';
+
+const handlerUtils = {
+  _sample,
```
jonahsnider, line 16: this is unused

jonahsnider (approved):
looks good. needs some tests but this should be good to merge otherwise

lib/services/runtime/handlers/carousel.ts
viniciusdacal: hmm, not needed actually, I added it to get cleaner error messages while implementing it.
Removing...

z4o4z (approved):

trs (approved):

PR #338, opened 2022-06-16
feat: second batch of steps for runtime logging (VF-3554)

jonahsnider:
**Fixes or implements VF-3554**
**Fixes or implements VF-3824**
**Fixes or implements VF-3825**
**Fixes or implements VF-3828**
**Fixes or implements VF-3832**

adds runtime logging for the following steps:

- condition step (if v1)
- condition step (if v2)
- flow step
- text step

### Related PRs

- https://github.com/voiceflow/libs/pull/305

### Checklist

### DIFF
--- a/runtime/lib/Handlers/flow.ts
+++ b/runtime/lib/Handlers/flow.ts
@@ -1,14 +1,15 @@
-import { BaseNode } from '@voiceflow/base-types';
+import { BaseNode, RuntimeLogs } from '@voiceflow/base-types';
 
 import { S } from '@/runtime/lib/Constants';
 import { HandlerFactory } from '@/runtime/lib/Handler';
 import Frame from '@/runtime/lib/Runtime/Stack/Frame';
 
+import DebugLogging from '../Runtime/DebugLogging';
 import { mapStores } from '../Runtime/utils/variables';
 
 const FlowHandler: HandlerFactory<BaseNode.Flow.Node> = () => ({
   canHandle: (node) => !!node.diagram_id,
-  handle: (node, runtime, variables) => {
+  handle: (node, runtime, variables, program) => {
     if (!node.diagram_id) {
       return node.nextId || null;
     }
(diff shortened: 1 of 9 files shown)


runtime/lib/Handlers/flow.ts
```diff
@@ -1,15 +1,19 @@
...
+import DebugLogging from '../Runtime/DebugLogging';
 import { mapStores } from '../Runtime/utils/variables';
 
 const FlowHandler: HandlerFactory<BaseNode.Flow.Node> = () => ({
   canHandle: (node) => !!node.diagram_id,
   handle: (node, runtime, variables) => {
     if (!node.diagram_id) {
+      // TODO: Check if this runs when you run a flow node where you haven't set the linked flow
+      // Update the log type to support logging a "before: {current flow} after: null" or something
+      // Maybe just log null entirely instead of a ValueChange
```
trs, line 16: We can explore these options after some testing. I'm sure we will be adding and removing log traces.

tests/runtime/lib/Handlers/flow.unit.ts
```diff
@@ -58,6 +72,30 @@ describe('flowHandler unit tests', () => {
       expect(runtime.trace.debug.args).to.eql([
         [`entering flow \`${newFrame.getProgramID()}\``, BaseNode.NodeType.FLOW],
       ]);
+      expect(runtime.trace.addTrace.args).to.eql([
```
trs, line 75: I don't love checking this way, but I understand you're working in the constraints of this test already.

Alternatively, could we just create an actual `Runtime` instance and pass that in? That way we could check the actual trace array that it has a `log` trace.
The focus is less on "what is this function calling" and more on "given these arguments, what response do I get?". It's a _big_ different with these since the side effects are the thing we want to test, but the approach remains.

trs (approved):

tests/runtime/lib/Handlers/flow.unit.ts
jonahsnider: i am also not a huge fan of this testing method. currently the way we test traces in general-runtime is by checking what `addTrace` gets called with which is why i decided to go this route.

as an alternative, creating a whole `Runtime` might be a bit tricky (lots of things to stub/mock), but creating a real `Trace` would work.

if it's alright with you, i'd like to stick with this style of testing for now, and then we can come back and totally redo all these tests in the future

trs: Yeah that's fine. Gotta strike a balance here

PR #340, opened 2022-06-16
refactor: change custom action trace type (VF-3846)

MrPicklePinosaur:
**Fixes or implements VF-3846**

Currently, custom action trace types use the name the user enters as the type. This can cause issues if they use a name of an already existing trace type.

Instead we can change the type of a custom action to "custom" and set the name in the payload.

```
{
  type: 'custom',
  payload: {
    name: string,
    body: string
  }
}
```

### Related PRs

- https://github.com/voiceflow/libs/pull/307
- https://github.com/voiceflow/general-service/pull/359

### Checklist


DecathectZero:
closing for now - there might be too many dependencies downstream that this might break

PR #342, opened 2022-06-20
feat: add custom action isBlocking checkbox support (VF-3675)

pmvrmc:
**Fixes or implements VF-3675**

If a custom action sets the isBlocking checkbox that is now propagated to the runtime

### Related PRs

- https://github.com/voiceflow/general-service/pull/358
- https://github.com/voiceflow/creator-app/pull/5626
- https://github.com/voiceflow/libs/pull/308
- https://github.com/voiceflow/adapters/pull/29


lib/services/runtime/handlers/_v1.ts
```diff
@@ -43,16 +43,18 @@ export const _V1Handler: HandlerFactory<BaseNode._v1.Node, typeof utilsObj> = (u
...
+    const stopTypes = runtime.turn.get<string[]>(TurnType.STOP_TYPES) || [];
+
+    const stop: boolean = !!runtime.turn.get(TurnType.STOP_ALL) || stopTypes.includes(node.type) || node.stop;
+
     runtime.trace.addTrace<BaseNode.Utils.BaseTraceFrame<unknown>>({
       type,
       payload,
       defaultPath: node.defaultPath,
       paths: node.paths.map((path) => ({ label: path.label, event: path.event! })),
+      stop,
```
trs, line 55: Again, I don't see why we'd need this. The runtime won't send any more traces if `stop` is true.

trs (changes requested):
Let's discuss this

lib/services/runtime/handlers/_v1.ts
pmvrmc: it wont send any more traces, but how the client/adapter handles it is different.

If the custom action is `stop` (or we send it on the stopTypes) the client needs to call interact again to proceed with the flow. Otherwise it can just handle that specific trace and the also the other ones.

This is basically syntatic sugar for not needing to add/check the stopTypes. The trace itself should hold all the information needed to handle it.

PR #348, opened 2022-06-23
refactor: versioned API routes (VF-0000)

MrPicklePinosaur:
**Fixes or implements VF-0000**

add versioned api routes support. refactor api structure from:
```
├── interact                                                                                                                                            
│   ├── defaultState                                                                                                                                    
│   └── interact                                                                                                                                        
└── state                                                                                                                                               
    ├── deleteState                                                                                                                                     
    ├── fetchState                                                                                                                                      
    ├── interact                                                                                                                                        
    ├── updateState                                                                                                                                     
    └── updateVariables 
```
to
```
├── v1
│   ├── stateful
│   │   └── ...
│   └── stateless
│       └── ...
└── v2alpha
    ├── stateful
    │   └── ...
    └── stateless
        └── ...
```

currently `v2alpha` is just a duplicated version of `v1`. this PR is just a prereq for merging some API changes from VF-3846 (https://github.com/voiceflow/general-runtime/pull/340)

### Checklist


lib/controllers/v2alpha/stateless/index.ts
```diff
@@ -0,0 +1,36 @@
+/**
+ * [[include:contextHandlers.md]]
+ * @packageDocumentation
+ */
```
trs, line 4: I've never seen this before, is it some sort of documentation generation?

backend/api/routers/v2alpha/stateful/index.ts
```diff
@@ -0,0 +1,105 @@
...
+  );
+
+  router.patch(
+    '/user/:userID/variables',
+    middlewares.rateLimit.versionConsume,
+    middlewares.project.attachID,
+    controllers.stateful.v2alpha.updateVariables
+  );
+
+  // Legacy 1.0.0 routes with versionID in params
```
trs, line 55: Why should we have legacy routes in a v2?

trs (changes requested):
If we're introducing a v2 we should put some improvements in place:
- A single "way" of calling interact. No more "legacy" routes.
- Improving the request
  - Remove some odd configs
  - No more stateful + stateless routes: single method
- Improving the response
  - Remove `stack`, this is a runtime implementation detail

We should also organize these better. We should group by domain and not by version. So have interact routes together declared with their version and path together, not all v1 together, etc. It's much harder to find a route when organized by version.

These new version were brought up with the new custom action format right? But I don't see where we are introducing that or adding the transformation back in the legacy routes.

backend/api/routers/v2alpha/stateful/index.ts
MrPicklePinosaur: seems like i messed something up when refactoring - whoops

MrPicklePinosaur:
> 

this PR was meant to get the basis of the versioning set up - so `v2alpha` at the moment is just a duplicate of `v1`. [https://github.com/voiceflow/general-runtime/pull/340] is meant to follow up and actually apply all the `v2alpha` changes

lib/controllers/v2alpha/stateless/index.ts
MrPicklePinosaur: hm im not sure, it seems to have been present in the old code

MrPicklePinosaur:
fyi: the tests for general-runtime routes are not very good, could use some rewrite in the future

PR #355, opened 2022-06-24
feat: third batch of steps for runtime logging (VF-3554)

jonahsnider:
**Fixes or implements VF-3554**
**Fixes or implements VF-3821**
**Fixes or implements VF-3822**
**Fixes or implements VF-3823**
**Fixes or implements VF-3828**
**Fixes or implements VF-3833**
**Fixes or implements VF-3834**
**Fixes or implements VF-4018**

- improves logs for the flow step
- adds logs for the intent step (ok, not really. but it does add logs for `nlu.intent_resolved` which is effectively the same thing)
- adds logs for the API step
- adds logs for the code step
- adds logs for the capture step (v1 & v2)
- adds logs for the `conversation_start` global event
- fixes a bug in how the `logs` query parameter was parsed when the runtime logging feature flag was enabled

most of the changes are pretty straightforward, though there were some more interesting ones.
specifically, i had to make some changes to the `Context` interface to support `nlu.intent_resolved`.

### Related PRs

- https://github.com/voiceflow/libs/pull/313

### Checklist


lib/services/runtime/index.ts
```diff
@@ -100,8 +103,31 @@ class RuntimeManager extends AbstractManager<{ utils: typeof utils }> implements
...
+      context.request
+    );
+    this.contextIDToRuntimeMap.set(context.id, runtime);
+
+    return runtime;
+  }
+
+  private garbageCollectRuntimeCache(): void {
+    [...this.contextIDToRuntimeMap.entries()]
+      .filter(([, runtime]) => runtime.hasEnded())
```
jonahsnider, line 127: now that i think about it, i'm pretty sure the context could enter a state where it's not ended, but will also never be ended. so, this would leak memory

lib/services/runtime/index.ts
```diff
@@ -100,8 +103,31 @@ class RuntimeManager extends AbstractManager<{ utils: typeof utils }> implements
...
+      .forEach(([id]) => {
+        this.contextIDToRuntimeMap.delete(id);
+      });
```
trs, line 130: Can we move this to a separate PR as it is more perf than log related? That way it can be tested and released in isolation.

jonahsnider: this is only performance related because of the `contextIDToRuntimeMap` map which prevents V8 from freeing `Context` objects that are no longer in use (since they're being referenced in the `Map`). imo it makes sense to keep the logic here in this PR

DecathectZero:
If its not being used I think we can just remove the debug logging service for now to not cause more confusion down the line - it shouldn't be hard to add it back in

lib/services/runtime/handlers/capture.ts
```diff
@@ -67,20 +67,42 @@ export const CaptureHandler: HandlerFactory<VoiceflowNode.Capture.Node | ChatNod
...
       if (!node.variable && node.slots?.length && request.payload.entities) {
-        variables.merge(
-          mapEntities(
-            node.slots.map((slot) => ({ slot, variable: slot })),
-            request.payload.entities
-          )
+        const variablesBefore: Record<string, RuntimeLogs.VariableValue | null> = Object.fromEntries(
+          node.slots.map((entity) => [entity, variables.get<RuntimeLogs.VariableValue>(entity) ?? null])
         );
+        const variablesAfter = mapEntities(
```
DecathectZero, line 73: I think we might not even need logging for this legacy capture step, we could flag it as deprecated

lib/services/runtime/handlers/captureV2.ts
```diff
@@ -94,18 +94,38 @@ export const CaptureV2Handler: HandlerFactory<VoiceflowNode.CaptureV2.Node, type
       };
 
       if (isNodeCapturingEntity(node) && intent.name === node.intent?.name) {
-        variables.merge(
-          mapEntities(
-            node.intent.entities.map((slot) => ({ slot, variable: slot })),
-            request.payload.entities
-          )
+        const variablesBefore: Record<string, RuntimeLogs.VariableValue | null> = Object.fromEntries(
+          node.intent.entities.map((entity) => [entity, variables.get<RuntimeLogs.VariableValue>(entity) ?? null])
```
DecathectZero, line 98: we can probably make a utility function for this kind of pattern in the `Store`

runtime/lib/Handlers/code/index.ts
```diff
@@ -42,34 +43,69 @@ const CodeHandler: HandlerFactory<BaseNode.Code.Node, CodeOptions | void> = ({
...
 
       return node.success_id ?? null;
     } catch (error) {
+      const serializedError = error.response?.data || error.toString();
       runtime.trace.debug(
-        `unable to resolve code  \n\`${safeJSONStringify(error.response?.data || error.toString())}\``,
+        `unable to resolve code  \n\`${safeJSONStringify(serializedError)}\``,
         BaseNode.NodeType.CODE
       );
+      if (node.id !== 'PROGRAMMATICALLY-GENERATED-CODE-NODE') {
```
DecathectZero, line 98: we should probably make this a `const`

lib/services/runtime/index.ts
```diff
@@ -101,7 +94,18 @@ class RuntimeManager extends AbstractManager<{ utils: typeof utils }> implements
...
 
   private getRuntimeForContext(context: Context): Runtime {
-    return this.createClient(context.data.api).createRuntime(context.versionID, context.state, context.request);
+    const runtime = this.createClient(context.data.api).createRuntime(
+      context.versionID,
+      context.state,
+      context.request
+    );
+
+    runtime.debugLogging.refreshContext(context);
```
DecathectZero, line 103: we don't really need a `refreshContext` function at all if we just pass in the logging level as a param during the creation of the runtime?

DecathectZero (approved):

lib/services/runtime/handlers/captureV2.ts
```diff
@@ -94,18 +94,38 @@ export const CaptureV2Handler: HandlerFactory<VoiceflowNode.CaptureV2.Node, type
...
+        });
 
         return handleCapturePath();
       }
 
       if (isNodeCapturingEntireResponse(node)) {
+        const variableBefore = variables.get<RuntimeLogs.VariableValue>(node.variable);
         variables.set(node.variable, query);
+        runtime.debugLogging.recordStepLog(RuntimeLogs.Kinds.StepLogKind.CAPTURE, node, {
+          changedVariables: {
```
DecathectZero, line 122: I wonder if we can build a utility function for generating the `changedVariables` notation

lib/services/runtime/index.ts
jonahsnider: the `.refreshContext()` method is used in other handlers where we can't access the runtime

lib/services/runtime/handlers/captureV2.ts
jonahsnider: so i experimented with trying to build a util function for this, and it is possible but it doesn't seem to help with boilerplate much. you end up with basically the same amount of code with either option. it's also tricky to get typesafety to work properly with the util function, but i bet it's possible to fix that

PR #356, opened 2022-06-24
[Merged by Bors] - feat: gotonode handler (CT-252)

z4o4z:
**Fixes or implements VF-252**

### Checklist


pmvrmc (commented):
curious, seems lib/services/runtime/handlers/gotoNode.ts and runtime/lib/Handlers/gotoNode.ts have the same exact content, do we really need both?

lib/services/runtime/handlers/gotoNode.ts
```diff
@@ -0,0 +1,39 @@
+import { BaseNode, BaseTrace } from '@voiceflow/base-types';
```
trs, line 1: We shouldn't be duplicating these handlers. We can just import the other handler from `@/runtime` in [the handler setup](https://github.com/voiceflow/general-runtime/blob/master/lib/services/runtime/handlers/index.ts).

trs (changes requested):

lib/services/runtime/handlers/gotoNode.ts
DecathectZero: what handlers are we duplicating here

lib/services/runtime/handlers/gotoNode.ts
```diff
@@ -0,0 +1,39 @@
...
+    if (!node.diagramID || node.diagramID === runtime.stack.top().getProgramID()) return node.nextId ?? null;
+
+    runtime.trace.addTrace<BaseTrace.PathTrace>({
+      type: BaseNode.Utils.TraceType.PATH,
+      payload: { path: 'jump' },
+    });
+
+    const frameIndex = runtime.stack.getFrames().findIndex((frame) => frame.getProgramID() === node.diagramID);
+
+    runtime.stack.popTo(frameIndex + 1);
```
DecathectZero, line 23: if the diagram doesn't exist in the stack `findIndex` is going to pop it all the way down to -1 for `frameIndex`.

If it goes to 0 we will be removing the `baseProgram` (which ties together all the topics with global intents)

I guess it's fine because we inject it back in every single update cycle? So by the time the next response comes in it should be okay.
https://github.com/voiceflow/general-runtime/blob/master/runtime/lib/Runtime/index.ts#L160

DecathectZero (approved):
Let's combine
`lib/services/runtime/handlers/gotoNode.ts`
`runtime/lib/Handlers/gotoNode.ts`
together so we can reuse the code?

z4o4z:
> curious, seems lib/services/runtime/handlers/gotoNode.ts and runtime/lib/Handlers/gotoNode.ts have the same exact content, do we really need both?

yeah, I just moved it from services to the platform but forgot to remove the original file, thanks! :)

lib/services/runtime/handlers/gotoNode.ts
z4o4z: yeah, just forgot to remove service one after moving the handler to platform folder :)

PR #358, opened 2022-06-24
[Merged by Bors] - feat: re-refactor API step fetch implementation code (VF-2744)

jonahsnider:
**Fixes or implements VF-2744**
**Fixes or implements VF-3919**

fixes a bug in config parsing for the API step that would cause every HTTP request to be aborted when the `API_MAX_TIMEOUT_MS` environment variable was not specified.

use lodash `merge()` to treat `{ a: undefined }` the same as ` {}`

previously the logic did not do this, and the default config of `{ a: 123 }` would be overwritten by `{ a: undefined }`, resulting in a config of `{ a: undefined }`.
this is an example config, but in the actual code this meant that the maximum duration before timeout was set to `undefined` milliseconds.

### Related PRs

- originally implemented in #334

### Checklist


runtime/lib/Handlers/api/utils.ts
```diff
@@ -110,76 +113,73 @@ const validateIP = async (hostname: string) => {
...
+      // Filter out variables that are undefined
+      .filter((keyValuePair): keyValuePair is [string, Variable] => keyValuePair[1] !== undefined)
+  );
+
+  return {
+    newVariables,
+    responseJSON:
+      // If response body is a JSON object then add in the `VF_` helpers
+      object.isObject(responseJSON) && !Array.isArray(responseJSON)
+        ? { ...responseJSON, ...responseBodyAdditions }
```
pmvrmc, line 180: do we have the same merge issue here? should we favor _merge?

jonahsnider: no, in this case we want to always overwrite those keys unconditionally. the `_merge` function is useful if you only want to merge in a property if it was `undefined` or missing from the object

trs (approved):

PR #364, opened 2022-06-30
feat: remove creator-api dependency (VF-3982)

trs:
**Fixes or implements VF-3982**

Remove the dependency on `creator-api` all together, rely instead on the `server-data-api`.

Environment variables no longer needed:
- `CREATOR_API_AUTHORIZATION`
- `CREATOR_API_ENDPOINT`

### Setup information

Run both this and https://github.com/voiceflow/server-data-api/pull/128 locally. Make any interact requests.

### Deployment Notes

The release in order needs to be:
1. https://github.com/voiceflow/server-data-api/pull/128
2. https://github.com/voiceflow/general-runtime/pull/364
3. https://github.com/voiceflow/creator-api/pull/1020

### Related PRs

- https://github.com/voiceflow/server-data-api/pull/128
- https://github.com/voiceflow/creator-api/pull/1020


runtime/lib/DataAPI/serverDataAPI.ts
```diff
@@ -85,6 +85,14 @@ class ServerDataAPI<
 
     return data;
   };
+
+  public getProjectUsingAuthorization = async (apiKey: string): Promise<PJ> => {
+    const split = apiKey.split('.');
+    const [, , apiKeyID] = split.length === 4 ? split : ['', ...split];
```
zhihil, line 91: This line of code feels a little arcane. Although less concise, it might be worth using `isDMAPIKey()` and `isWorkspaceAPIKey()` to spell out for the reader why `split.length === 4` or `split.length !== 4` are different cases.

runtime/lib/DataAPI/serverDataAPI.ts
```diff
@@ -85,6 +85,14 @@ class ServerDataAPI<
...
+
```
zhihil, line 92: How does this handle the case where `authorization` is `undefined`? I don't think we ever figured out the black magic of how the prototype tool is able to contact creator without the authorization.

lib/services/state/cacheDataAPI.ts
```diff
@@ -40,6 +40,11 @@ class CacheDataAPI implements DataAPI<VoiceflowProgram.Program, VoiceflowVersion
   async fetchDisplayById(displayId: number) {
     return this.api.fetchDisplayById(displayId);
   }
+
+  async getProjectUsingAuthorization(apiKeyID: string): Promise<VoiceflowProject.Project> {
+    const apiKey = await this.api.getProjectUsingAuthorization(apiKeyID);
+    return apiKey as VoiceflowProject.Project;
```
zhihil, line 46: Is the naming here incorrect? It should be `const project = await` not `const apiKey = `?

zhihil (changes requested):

runtime/lib/DataAPI/serverDataAPI.ts
zhihil: Nevermind, tested on the prototype tool. This appears to work.

runtime/lib/DataAPI/serverDataAPI.ts
trs: Added a helper to use here: https://github.com/voiceflow/libs/pull/316

lib/middlewares/project.ts
```diff
@@ -79,7 +79,7 @@ class Project extends AbstractMiddleware {
         );
       }
 
-      const project = await api.getProjectUsingAuthorization(req.headers.authorization!).catch(() => {
+      const project = await api.getProjectUsingAPIKey(req.headers.authorization!).catch(() => {
```
zhihil, line 82: Don't forget to change the check here `(!Utils.object.hasProperty(api, 'getProjectUsingAuthorization'))` in this file to
```js
(!Utils.object.hasProperty(api, 'getProjectUsingAPIKey'))
```
One of my tests just failed because the check was not updated to this new method name.

zhihil (changes requested):

trs:
This change is a bit premature since there is no authorization in place to restrict which projects/versions a key has access to.
Instead we can retain the creator api call until the authorization revamp is further along.

PR #383, opened 2022-07-19
ci: add renew env label flow

xavidop:
**Fixes or implements IN-331**

Add the renew env flow

New GH ACtion that checks if a new env exists, installs vfcli, and renews an environment if this one exists

### Deployment Notes

This change should be propagated to all repos

### Checklist

### DIFF
--- a/.github/workflows/renew-env.yaml
+++ b/.github/workflows/renew-env.yaml
@@ -1,0 +1,74 @@
+name: PR-label -> renew dev env
+
+on:
+  pull_request:
+    types: [labeled]
+
+concurrency:
+  group: renew-env-${{ github.ref }}
+
+jobs:
+  preview:
+    name: Renew dev environment
+    runs-on: ubuntu-latest
+    continue-on-error: true # Ignore errors
+    if: ${{ github.actor != 'dependabot[bot]' }}
+    steps:
+      - name: Check PR env labels
+        uses: voiceflow/pr-label-match@master
+        with:
+          regex: env-
+        id: envNameLabel
+
+      - name: Check for new env request
+        uses: Dreamcodeio/pr-has-label-action@master
+        id: renewEnvLabel
+        with:
+          label: renew-env
+
+      - name: Remove renew-env label (if exist)
+        uses: buildsville/add-remove-label@v1
+        if: steps.renewEnvLabel.outputs.hasLabel == 'true' && steps.renewEnvLabel.outputs.label == ''
+        with:
+          token: ${{secrets.GH_SA_TOKEN}}
+          label: renew-env
+          type: remove
+
+      - name: Install vfcli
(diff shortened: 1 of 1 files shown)


.github/workflows/renew-env.yaml
```diff
@@ -0,0 +1,75 @@
...
+      - name: Remove renew-env label (if exist)
+        uses: buildsville/add-remove-label@v1
+        if: steps.renewEnvLabel.outputs.hasLabel == 'true' && steps.renewEnvLabel.outputs.label == ''
+        with:
+          token: ${{secrets.GH_SA_TOKEN}}
+          label: renew-env
+          type: remove
+
+      - name: Install vfcli
+        if: (github.event.action == 'labeled' && steps.renewEnvLabel.outputs.hasLabel == 'true' && steps.envNameLabel.outputs.label != '')
```
Xavientois, line 38: The check is redundant because the workflow is only triggered by that event

```suggestion
        if: (steps.renewEnvLabel.outputs.hasLabel == 'true' && steps.envNameLabel.outputs.label != '')
```

.github/workflows/renew-env.yaml
```diff
@@ -0,0 +1,75 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_SA_TOKEN }}
+        run: |
+          API_URL="https://$GITHUB_TOKEN:@api.github.com/repos/voiceflow/vfcli"
+          ASSET_ID=$(curl $API_URL/releases/latest | jq -r '.assets[3].id')
+          curl -J -L -H "Accept: application/octet-stream" "$API_URL/releases/assets/$ASSET_ID" --output vfcli.tar.gz
+          tar -xf vfcli.tar.gz
+
+      - name: Renew the dev env if PR closed not creating and the label exists
```
Xavientois, line 47: I don't understand what `if PR closed not creating` means

.github/workflows/renew-env.yaml
```diff
@@ -0,0 +1,75 @@
...
+        if: (github.event.action == 'labeled' && steps.renewEnvLabel.outputs.hasLabel == 'true' && steps.envNameLabel.outputs.label != '')
```
Xavientois, line 48: The check is redundant because the workflow is only triggered by that event
```suggestion
        if: (steps.renewEnvLabel.outputs.hasLabel == 'true' && steps.envNameLabel.outputs.label != '')
```

xavidop: fixed

.github/workflows/renew-env.yaml
xavidop: that was a typo due to the GH Action I started off

.github/workflows/renew-env.yaml
xavidop: fixed

xavidop:
> For the third-party actions we are using, we should probably not point to their default branch, but rather to a specific released version (or at least the `latest` version). Pointing to the default branch means we might get breaking behaviour.

Pinned versions!

PR #384, opened 2022-07-19
[Merged by Bors] - fix: capture step leaves user hanging if no entity prompt is provided (VF-3977)

DecathectZero:
**Fixes or implements VF-3977**

one important change here is not completely clearing the `dmStateStore` when we don't match something during entity filling - we keep the `priorIntent` for the capture step to act on, so after no match, if the user has already filled the first entity, we still retain their answer for that first entry.

The other part is only set `context.end` to true IF there is an elicit or a prompt

### DIFF
--- a/lib/services/dialog/index.ts
+++ b/lib/services/dialog/index.ts
@@ -53,7 +53,7 @@ class DialogManagement extends AbstractManager<{ utils: typeof utils }> implemen
     dmPrefixedResult: BaseRequest.IntentRequest,
     incomingRequest: BaseRequest.IntentRequest,
     languageModel: BaseModels.PrototypeModel
-  ): boolean => {
+  ): void => {
     const dmPrefixedResultName = dmPrefixedResult.payload.intent.name;
     const incomingRequestName = incomingRequest.payload.intent.name;
     const expectedIntentName = dmStateStore.intentRequest!.payload.intent.name;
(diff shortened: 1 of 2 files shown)


lib/services/dialog/index.ts
```diff
@@ -52,7 +52,7 @@ class DialogManagement extends AbstractManager<{ utils: typeof utils }> implemen
     dmPrefixedResult: BaseRequest.IntentRequest,
     incomingRequest: BaseRequest.IntentRequest,
     languageModel: BaseModels.PrototypeModel
-  ): boolean => {
```
DecathectZero, line 55: this function was really annoying, it was mutating `dmStateStore.intentRequest` AND sending back a boolean signal to act on.

lib/services/dialog/index.ts
```diff
@@ -93,13 +93,9 @@ class DialogManagement extends AbstractManager<{ utils: typeof utils }> implemen
       // Action: Migrate user to the new intent and extract all the available entities
       dmStateStore.intentRequest = incomingRequest;
```
DecathectZero, line 94: case A1 is what was happening most of the time when we typed gibberish. `dmPrefixedResultName` = `incomingRequestName` = `"None"`

lib/services/dialog/index.ts
```diff
@@ -144,14 +140,9 @@ class DialogManagement extends AbstractManager<{ utils: typeof utils }> implemen
...
-        const isFallback = this.handleDMContext(
-          dmStateStore,
-          dmPrefixedResult,
-          incomingRequest,
-          version.prototype.model
-        );
+        this.handleDMContext(dmStateStore, dmPrefixedResult, incomingRequest, version.prototype.model);
 
-        if (isFallback) {
+        if (dmStateStore.intentRequest.payload.intent.name === NONE_INTENT) {
```
DecathectZero, line 145: this catches case A1 where we just straight up get a None intent. This actually mirrors the code on line 159 now:
https://github.com/voiceflow/general-runtime/blob/de5e840943d61773592017efb12670764079b23b/lib/services/dialog/index.ts#L159-L164

lib/services/dialog/index.ts
pmvrmc: good call, makes the function much simpler

pmvrmc (approved):

pmvrmc:
Thank you for improving the solution and getting a fix, I agree with you that this logic is still too convoluted and we might consider a refactor.

However when testing this out I faced an issue. I think it has to deal with my utterances not being correct, but can you check this weird behaviour?
image
image

PR #387, opened 2022-07-26
[Merged by Bors] - docs: updating OpenAPI spec (VF-4096)

zhihil:
**Fixes or implements VF-4096**

Updates the documentation for the Dialog Manager API (`general-runtime`) to be clearer, explain some of our features, and be more comprehensive.

See this [readme instance](https://testvoiceflow.readme.io/reference/overview) for what the updated OpenAPI docs will look like.

Edited the `openapi.yaml`

I will also need to manually edit the OpenAPI docs since the readme sync isn't perfect. I will make those edits by copy-pasting what you see in the [test instance](https://testvoiceflow.readme.io/reference/overview) once this PR is merged.

### Setup information

N/A

### Deployment Notes

N/A

### Related PRs

N/A

### Checklist


backend/docs/openapi.yaml
```diff
@@ -188,23 +337,60 @@ tags:
...
       ```
-      Now whenever we come across a `Pay Credit Card` trace, we can actually trigger a side effect such as the Stripe API and perform the action.
-      Depending on the result of that, we can now select which path we want to go down. In the subsequent request body, you simply have to do this:
+
+      Your client code might extract this last trace, detect that it is a custom "Pay Credit Card" trace, and then perform some logic
+      such as calling the Stripe API to charge your user's credit card. 
+
+      
+      ```js
+
```
jonahsnider, line 348: markdown code blocks shouldn't start or end with blank lines

jonahsnider (approved):

backend/docs/openapi.yaml
```diff
@@ -297,45 +485,134 @@ tags:
         // state, config, and other metadata here
       }
       ```
+  - name: API Key Differences
+    description: >
+      It is **recommended** to use a **Dialog Manager API Key** to access the Dialog Manager API. A DM API key has the prefix
+      `VF.DM.` and it provides the most up-to-date features such as version aliasing.
+
+
+      For backwards compatibility reasons, the DM API key accepts Workspace API keys (`VF.WS`) and legacy Workspace API keys (`VF.`).
```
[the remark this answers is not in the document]
zhihil: This is a quirk of folded mode, which I was using because writing long paragraphs in literal mode is inconvenient. However, this change turned out to be unnecessary anyways because the Readme platform uses plain Markdown and not "Markdown in YAML literal/folded mode". 

Bottomline is that writing plain text in YAML was really tedious and resulted in some awkward decisions here. 

I might change everything back to literal mode next utility sprint, because it turns out to be easier. But I need to prioritize the Adapter work atm.

backend/docs/openapi.yaml
zhihil: In folded mode this is necessary, otherwise the parsing get's messed up. This will be changed to literal mode next utility sprint when I have capacity again.

PR #390, opened 2022-07-29
feat: use auth service (PL-71)

jonahsnider:
# don't merge, the auth service hasn't been deployed yet

**Fixes or implements PL-71**

remove all references to creator-api

use server-data-api directly & do authorization checks with the auth service

### Checklist


lib/clients/auth.ts
```diff
@@ -0,0 +1,57 @@
...
+
+import { Config } from '@/types';
+
+export type AuthType = AuthSDK<Permission>;
+
+const Auth = (config: Config): AuthType => new AuthSDK(config.AUTH_SERVICE_ENDPOINT, undici.fetch);
+
+export default Auth;
+
+export enum Permission {
```
trs, line 12: I feel like just restricting this to what permissions are actually used here is probably better. We can explore places to declare these in a sharable manner in the future.

lib/middlewares/auth.ts
```diff
@@ -0,0 +1,83 @@
...
+  ) {
+    super(services, config);
+  }
+
+  @validate({
+    HEADER_AUTHORIZATION: VALIDATIONS.HEADERS.AUTHORIZATION,
+    HEADER_VERSION_ID: VALIDATIONS.HEADERS.VERSION_ID,
+    HEADER_PROJECT_ID: VALIDATIONS.HEADERS.PROJECT_ID,
+  })
+  async assertPermissionsMiddleware(_req: Request, _res: Response, next: express.NextFunction): Promise<void | never> {
```
trs, line 50: In the future we should explore providing a middleware from the sdk directly.

lib/middlewares/auth.ts
```diff
@@ -0,0 +1,83 @@
...
+        authorization: string;
+      }
+    >;
+
+    const { versionID, projectID } = req.headers;
+
+    try {
+      await this.services.auth.assertAuthorized(
+        // TODO: Pretty sure this is wrong
+        `ApiKey ${req.headers.authorization}`,
```
trs, line 66: Yep, we'd need to check if it looks like an API key (starts with `VF.`), if it doesn't then the auth scheme is `Bearer`.
Can check out [`creator-api` when it does it's verification](https://github.com/voiceflow/creator-api/blob/master/lib/services/authManager/index.ts#L79).

trs (changes requested):

package.json
```diff
@@ -24,6 +24,7 @@
     "@voiceflow/logger": "^1.6.1",
     "@voiceflow/metrics": "1.4.1",
     "@voiceflow/natural-language-commander": "^0.5.1",
+    "@voiceflow/sdk-auth": "../platform/libs/sdk-auth/",
```
trs, line 27: We'll have to wait till we have the CI setup to merge this too.

trs (changes requested):

package.json
```diff
@@ -55,6 +56,7 @@
     "require-from-url": "^3.1.3",
     "slate": "0.72.3",
     "talisman": "^1.1.3",
+    "undici": "^5.8.0",
```
trs, line 59: This includes node-fetch already, should we just use that?

lib/middlewares/auth.ts
jonahsnider: i'd like to if it'd be helpful, but it seems to me that there is very little that can be abstracted here (at least, if we want to reuse the middleware in other services)

types.ts
```diff
@@ -38,9 +38,7 @@ export interface Config extends RateLimitConfig {
 
   GENERAL_SERVICE_ENDPOINT: string | null;
   LUIS_SERVICE_ENDPOINT: string | null;
-
-  CREATOR_API_ENDPOINT: string | null;
-  CREATOR_API_AUTHORIZATION: string | null;
+  AUTH_SERVICE_ENDPOINT: string;
```
Xavientois, line 41: Since this is non-null, will it complain if `AUTH_SERVICE_ENDPOINT` is not set in the environment?

jonahsnider: yes, because this value is populated with `getRequiredProcessEnv()` (which throws an error if the env var isn't defined)

which, now that i think about it, is a mistake most likely. this will break the runtime for selfhosters

jonahsnider: i've updated the PR to make `AUTH_SERVICE_ENDPOINT` optional

xavidop:
bors ping

xavidop:
bors ping

xavidop:
bors ping

xavidop:
bors ping

xavidop:
LETS GOOO POOOOOOONG

PR #397, opened 2022-08-10
feat: create a dedicated handler for CardV2 (CT-749)

filipemerker:
CardV2 was formerly using the Carousel handler, but now has its own.


viniciusdacal (approved):

viniciusdacal (approved):

z4o4z (approved):

lib/services/runtime/handlers/_v1.ts
```diff
@@ -13,7 +13,20 @@ const utilsObj = {
   findEventMatcher,
 };
 
-export const _V1Handler: HandlerFactory<BaseNode._v1.Node, typeof utilsObj> = (utils) => ({
+const getNodeType = (node: BaseNode._v1.Node, variablesMap: Readonly<AnyRecord>) => {
```
pmvrmc, line 16: why is this PR touching the _V1 handler?

lib/services/runtime/handlers/card.ts
```diff
@@ -0,0 +1,38 @@
+import { BaseNode, BaseTrace } from '@voiceflow/base-types';
```
pmvrmc, line 1: why do we need a cardV1 handler on general-service? the cardV1 is just for alexa/google right? general will just use cardv2 or not?

lib/services/runtime/handlers/channelAction.ts
```diff
@@ -0,0 +1,33 @@
+import { BaseNode } from '@voiceflow/base-types';
```
pmvrmc, line 1: why is this PR touching channelAction?

PR #398, opened 2022-08-11
[Merged by Bors] - feat: add channel action runtime handler (PL-38)

pmvrmc:
**Fixes or implements PL-38**

Adds a new runtime handlers that handles CHANNEL_ACTION nodes. It will extract variables and passdown the payload to the adapter

Since this is just adding the runtime handler, we are good to merge when ready since nothing will consume it

### DIFF
--- a/lib/services/runtime/handlers/channelAction.ts
+++ b/lib/services/runtime/handlers/channelAction.ts
@@ -1,0 +1,64 @@
+/* eslint-disable no-restricted-syntax */
+import { BaseNode } from '@voiceflow/base-types';
+import { object, replaceVariables } from '@voiceflow/common';
+
+import { Action, HandlerFactory } from '@/runtime';
+
+import { TurnType } from '../types';
+import CommandHandler from './command';
+import { findEventMatcher } from './event';
+
+const utilsObj = {
+  commandHandler: CommandHandler(),
+  findEventMatcher,
+};
+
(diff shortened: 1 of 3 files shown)


DecathectZero (approved):

lib/services/runtime/handlers/channelAction.ts
```diff
@@ -0,0 +1,64 @@
...
+import { findEventMatcher } from './event';
+
+const utilsObj = {
+  commandHandler: CommandHandler(),
+  findEventMatcher,
+};
+
+export const ChannelActionHandler: HandlerFactory<BaseNode.ChannelAction.Node, typeof utilsObj> = (utils) => ({
+  canHandle: (node) => node.type === BaseNode.NodeType.CHANNEL_ACTION,
+  handle: (node, runtime, variables) => {
```
DecathectZero, line 18: could we just call `_V1Handler.handle()` from this handle? We would need to modify the `node` so it conforms with the _v1 format but it avoids duplicating code like this

pmvrmc: reafactored it, do you agree with it?

DecathectZero (approved):

lib/services/runtime/handlers/channelAction.ts
DecathectZero: yup! Sorry about the late reply - looks good

PR #406, opened 2022-08-22
[Merged by Bors] - fix: cleaner channelaction implementation (PL-38)

DecathectZero:
**Fixes or implements PL-38**

Wanted to clean up the channel action logic more so it just extends the _v1

### DIFF
--- a/lib/services/runtime/handlers/_v1.ts
+++ b/lib/services/runtime/handlers/_v1.ts
@@ -1,6 +1,6 @@
 /* eslint-disable no-restricted-syntax */
 import { BaseNode } from '@voiceflow/base-types';
-import { AnyRecord, replaceVariables } from '@voiceflow/common';
+import { object, replaceVariables } from '@voiceflow/common';
 
 import { Action, HandlerFactory } from '@/runtime';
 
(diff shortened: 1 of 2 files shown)


lib/services/runtime/handlers/_v1.ts
```diff
@@ -53,10 +40,10 @@ export const _V1Handler: HandlerFactory<BaseNode._v1.Node, _v1Utils> = (utils) =
     }
 
     const variablesMap = variables.getState();
-    const type = utils.getNodeType ? utils.getNodeType(node, variablesMap) : getNodeType(node, variablesMap);
-    const payload = utils.getNodePayload
-      ? utils.getNodePayload(node, variablesMap)
-      : getNodePayload(node, variablesMap);
+    const type = replaceVariables(node.type, variablesMap);
```
DecathectZero, line 43: I think this is fine because the channelAction is never going to have a `{variable}` syntax inside it.

lib/services/runtime/handlers/_v1.ts
```diff
@@ -53,10 +40,10 @@ export const _V1Handler: HandlerFactory<BaseNode._v1.Node, _v1Utils> = (utils) =
...
+    const payload = object.deepMap(node.payload, (value) =>
```
DecathectZero, line 44: it's okay to deeply replace variables in the payload - this would be a great feature for many people.

`object.deepMap(string)` is fine too.

lib/services/runtime/handlers/channelAction.ts
```diff
@@ -1,33 +1,23 @@
...
+  handle: (node, ...args) => {
+    const _v1Node: BaseNode._v1.Node = {
+      ...node,
+      _v: 1,
+      type: node.data.name,
+      payload: node.data.payload,
+    };
+
+    return _v1Handler.handle(_v1Node, ...args);
+  },
```
pmvrmc, line 20: good idea to instead of passing the utils fns around just map the object to another strucutre, seems simpler 👍

pmvrmc (approved):
awesome work! ty

PR #407, opened 2022-08-22
fix: small tweaks for adapters

DecathectZero:
**Fixes or implements PL-102**

Yet to be tested. reference comments


lib/services/runtime/handlers/capture/capture.alexa.ts
```diff
@@ -40,10 +39,12 @@ export const CaptureAlexaHandler: HandlerFactory<VoiceflowNode.Capture.Node, typ
...
       utils.addRepromptIfExists({ node, runtime, variables });
 
-      // TODO add elicit trace?
-      // if (node.intent && node.slots?.[0]) {
-      //   runtime.turn.set<TurnElicitSlot>(Turn.ELICIT_SLOT, createElicitSlot(node.intent, node.slots));
-      // }
+      if (node.intent && node.slots?.[0]) {
+        runtime.trace.addTrace<BaseTrace.GoToTrace>({
+          type: BaseTrace.TraceType.GOTO,
+          payload: { request: setElicit(entityFillingRequest(node.intent, node.slots), true) },
```
DecathectZero, line 45: `GOTO + setElicit(, true)` === `Turn.ELICIT_SLOT` (the prototype tool recognizes this format)

lib/services/runtime/handlers/capture/capture.alexa.ts
```diff
@@ -69,9 +70,6 @@ export const CaptureAlexaHandler: HandlerFactory<VoiceflowNode.Capture.Node, typ
 
     ({ nextId = null } = node);
 
-    // request for this turn has been processed, delete request
-    runtime.turn.delete(Turn.REQUEST);
-
```
DecathectZero, line 74: this `Turn.REQUEST` system is replaced by `runtime.getAction`:
https://github.com/voiceflow/general-runtime/blob/master/runtime/lib/Runtime/cycleHandler.ts#L47-L50

lib/services/runtime/handlers/capture/capture.google.ts
```diff
@@ -2,15 +2,13 @@ import { VoiceflowConstants, VoiceflowNode } from '@voiceflow/voiceflow-types';
...
 import { Action, HandlerFactory } from '@/runtime';
-import { Turn } from '@/runtime/lib/constants/flags.google';
 
 import { addButtonsIfExists } from '../../utils';
 import { addRepromptIfExists, isGooglePlatform } from '../../utils.google';
 import CommandHandler from '../command/command';
 import NoInputHandler from '../noReply/noReply.google';
 
 const utilsObj = {
-  v: '',
```
DecathectZero, line 13: we can just assume its ALWAYS `v2`

lib/services/runtime/utils.alexa.ts
```diff
@@ -63,6 +63,9 @@ export const addRepromptIfExists = <B extends VoiceflowNode.Utils.NoReplyNode>({
   const noReplyNode = convertDeprecatedReprompt(node);
   const prompt = _.sample(noReplyNode.noReply.prompts);
   if (prompt && typeof prompt === 'string') {
-    runtime.turn.set(Turn.REPROMPT, replaceVariables(prompt, variables.getState()));
+    runtime.trace.addTrace<BaseNode.Utils.BaseTraceFrame<unknown>>({
+      type: Turn.REPROMPT,
+      payload: replaceVariables(prompt, variables.getState()),
+    });
```
DecathectZero, line 69: I made a custom trace type here because Alexa requires the reprompt to be sent along with the output - it handles the reprompt logic.

on voiceflow, when we get a mismatch, the next turn we just send the reprompt as the output

lib/services/runtime/handlers/utils/entity.ts
```diff
@@ -8,9 +8,13 @@ import { StorageType } from '../../types';
...
 export const VF_ELICIT = 'ELICIT';
 
-export const entityFillingRequest = (name: string): BaseRequest.IntentRequest => ({
+export const entityFillingRequest = (
+  name: string,
+  requiredEntities?: string[]
+): BaseRequest.IntentRequest & { requiredEntities?: string[] } => ({
   type: BaseRequest.RequestType.INTENT,
   payload: { intent: { name }, query: '', entities: [] },
+  requiredEntities,
```
DecathectZero, line 17: alexa has this thing where all the required slots to be defined:
https://github.com/voiceflow/alexa-runtime/blob/0f25e89b842b768efc40ad634e4bf9097099da5a/lib/services/runtime/handlers/utils/directives.ts#L9-L17

lib/services/runtime/handlers/text.ts
```diff
@@ -5,8 +5,8 @@ import _sample from 'lodash/sample';
 import log from '@/logger';
```
DecathectZero, line 5: I don't think theres any difference between the `text.google` and general text?

lib/services/runtime/handlers/command/command.alexa.ts
```diff
@@ -32,7 +32,6 @@ export const getCommand = (runtime: Runtime) => {
     if (!found) {
       intentName = IntentName.STOP;
       _.set(request, 'payload.intent.name', intentName);
```
DecathectZero, line 34: I think we're mutating the request object already? `Turn.REQUEST` is not a thing in general.
https://github.com/voiceflow/general-runtime/blob/95fb393680d2a94651ebcc16999f57d2a43a99fb/runtime/lib/Runtime/index.ts#L110-L112

lib/services/runtime/handlers/captureV2/captureV2.alexa.ts
```diff
@@ -46,25 +49,24 @@ export const CaptureV2Handler: HandlerFactory<VoiceflowNode.CaptureV2.Node, type
...
     }
 
     // handle noMatch
     const noMatchPath = utils.noMatchHandler.handle(node, runtime, variables);
     if (noMatchPath === node.id && node.intent?.entities) {
-      // TODO add delegate trace?
-      // runtime.turn.set<Intent>(Turn.DELEGATE, createDelegateIntent(node.intent.name, node.intent.entities));
+      runtime.trace.addTrace<BaseTrace.GoToTrace>({
+        type: BaseTrace.TraceType.GOTO,
+        payload: { request: entityFillingRequest(node.intent.name, node.intent.entities) },
```
DecathectZero, line 68: think of `GOTO` without `setElicit(, true)` as  `Turn.DELEGATE`

lib/services/runtime/handlers/capture/capture.alexa.ts
pmvrmc: I was curious on what this was about, thanks for clarifying

lib/services/runtime/handlers/command/command.alexa.ts
pmvrmc: got it, makes sense!

lib/services/runtime/handlers/text.ts
pmvrmc: Diff was `runtime.turn.set(Turn.DF_ES_TEXT_ENABLED, true);` done in google but not general. But as you pointed, it should be the adapter doing that logic everytime it gets a text trace. Since this logic is not under an if clause we can just remove it.

pmvrmc (approved):

PR #412, opened 2022-09-02
[Merged by Bors] - fix: dont display empty card (CT-882)

amandasteinhauer:
**Fixes or implements CT-882**

check if card has any content to prevent displaying a line for an empty card in the prototype tool


z4o4z (approved):

viniciusdacal:
if we extract this bit to a `getDescription` function, should be enough to remove the error.

```ts
if (typeof node.description === 'string') {
  const parsedDescription = replaceVariables(node.description, variablesMap);
  description = {
    text: parsedDescription,
    slate: [{ text: parsedDescription }],
  };
} else {
  const slateValue = utils.slateInjectVariables(
    node.description as BaseText.SlateTextValue,
    utils.sanitizeVariables(variables.getState())
  );
  description = {
    slate: slateValue,
    text: utils.slateToPlaintext(slateValue),
  };
}
```

PR #420, opened 2022-09-21
feat: public api route (PL-170)

DecathectZero:
**Fixes or implements PL-170**

This PR depends on https://github.com/voiceflow/libs/pull/369 and https://github.com/voiceflow/database-cli/pull/260

The setup for a public api route is based on the `projectID` where we do not expose any sensitive information or allow the client to manipulate the state. This is still rate-limited by `versionID` (maybe we should do `projectID` in the future?).

We check that `project.apiPrivacy` is `BaseModels.Project.Privacy.PUBLIC` to allow access. To test, you have to manipulate this field directly on your development mongo:
Screen Shot 2022-09-21 at 2 02 25 AM

The key to all of this working is here: 
https://github.com/voiceflow/general-runtime/blob/6b8d14ba1efb1c0dceebbd346adeb30c992c4f9e/lib/clients/dataAPI.ts#L77-L93
we see that if `authorization` is `undefined` we end up using `this.remoteDataApi` which is the `server-data-api`, which allows us to call any project/version/program resource.

This is how the FE creator-app allows for publicly shared prototypes. The other endpoints use `middlewares.rateLimit.verify` which ensures only requests from the `creator-app` origin can call it - which again, isn't great. 
https://github.com/voiceflow/general-runtime/blob/626aafcb7da3c9cf0cd87c56a7e7900e890dba69/lib/middlewares/rateLimit.ts#L11-L22

We should absolutely refactor this whole system to use `mongoApi` over `server-data-api` soon and fix all the brittle middleware layers that have a bunch of dependent side effects.


lib/controllers/stateManagement/index.ts
```diff
@@ -38,12 +38,32 @@ class StateManagementController extends AbstractController {
...
+      body: { action },
+    } = req;
+
+    // only pass in select properties to avoid any potential security issues
+    return this.services.stateManagement.interact({
+      headers: { projectID, versionID },
+      params: { userID },
+      body: { action },
+      query: {},
+    });
```
DecathectZero, line 64: this is to ensure we carefully control what parameters/config can be used. i.e. no logs, verbose mode, etc.

lib/controllers/stateManagement/index.ts
```diff
@@ -38,12 +39,38 @@ class StateManagementController extends AbstractController {
...
+    HEADERS_VERSION_ID: VALIDATIONS.HEADERS.VERSION_ID,
+  })
+  async publicInteract(
+    req: Request<
+      { userID: string },
+      { action?: RuntimeRequest; config?: BaseRequest.RequestConfig },
+      { projectID: string; versionID: string }
+    >
+  ) {
+    // only pass in select properties to avoid any potential security issues
```
zhihil, line 59: I think there's a `customAJV` utility that we use in some backend services, which allows us to define validation schemas, which could detect extraneous filters for us. This might be a more foolproof approach to filtering out properties.

lib/middlewares/project.ts
```diff
@@ -128,6 +129,46 @@ class Project extends AbstractMiddleware {
...
+
+    const project = await api?.getProject(req.params.projectID).catch(() => {
+      throw new VError('no access to project', VError.HTTP_STATUS.UNAUTHORIZED);
+    });
+
+    if (project.apiPrivacy !== BaseModels.Project.Privacy.PUBLIC) {
+      throw new VError('no access to project', VError.HTTP_STATUS.UNAUTHORIZED);
+    }
+
+    req.headers.projectID = req.params.projectID;
```
zhihil, line 160: I think this line is unnecessary no? I think we only did this for the DM API middlewares because the new format should pass `projectID` in the headers, but we needed to adapt the old format where we passed it in the params. 

Since this is a new endpoint, I think we can just mandate that the `projectID` is passed into `headers` OR just use `req.params.projectID` in the controller.

DecathectZero: it's not just the `controllers` but a bunch of things downstream (services) also rely on `projectID` being in the headers. We should probably figure out ways to clean it up
https://github.com/voiceflow/general-runtime/blob/25f105fb3ecfec11f86045fc4120626d54817437/lib/services/stateManagement.ts#L10-L19

lib/controllers/stateManagement/index.ts
DecathectZero: yeah I agree, I can make an AJV for it

DecathectZero: The `customAJV` pattern we have only applies to the body object of the request, however we do have to filter out additional things from the header, params. So we still might have to do the parsing there.

DecathectZero: yeah we still need to ensure `authorization` and other properties are not passed in the headers etc

backend/api/routers/public.ts
```diff
@@ -0,0 +1,22 @@
...
+export default (middlewares: MiddlewareMap, controllers: ControllerMap) => {
+  const router = express.Router();
+
+  router.use(bodyParser.json({ limit: BODY_PARSER_SIZE_LIMIT }));
+
+  const interactMiddleware = [middlewares.project.resolvePublicProjectID, middlewares.rateLimit.versionConsume];
+
+  // full route: /public/:projectID/state/user/:userID/interact
+  router.post(
+    '/:projectID/state/user/:userID/interact',
```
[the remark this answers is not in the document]
DecathectZero: This is because `projectID` is replacing the responsibility of the API key itself, which maps to a projectID. Also the other routes with `versionID` in them are considered legacy and that's not the pattern we promote today.

The only way to enable aliases `production` `development` and have a streamlined publishing flow, is for the client to reference the `projectID`

astopo (approved):

DecathectZero:
the e2e failed because of an `npm` server issue, I'll save it from having to run again

PR #423, opened 2022-10-01
[Merged by Bors] - fix: public interact response shape (CT-1046)

DecathectZero:
**Fixes or implements CT-1046**

we have this very annoying/bad pattern determined here:
https://github.com/voiceflow/general-runtime/blob/4ee8ab709790bf7dadb4fe337e83923cfccd00ab/lib/services/stateManagement.ts#L27

where it might give back an array of traces OR an object with `{ state, trace: Trace[], request }` inside.
Since the public endpoint isn't used yet, it would be much better for it to return `{ trace: Trace[] }` instead of just `Trace[]`.

This gives us opportunities to add additional return details in the future.

### DIFF
--- a/lib/controllers/stateManagement/index.ts
+++ b/lib/controllers/stateManagement/index.ts
@@ -58,12 +58,14 @@ class StateManagementController extends AbstractController {
       { projectID: string; versionID: string }
     >
   ) {
-    return this.services.stateManagement.interact({
+    const trace = await this.services.stateManagement.interact({
       ...req,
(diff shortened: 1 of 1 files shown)


pmvrmc (approved):
I really like this pattern, ideally it should be applied to all endpoints.

I think that even better would be to have a `{ meta, data }` on all responses, then traces would live inside data.

But that requires a bigger change and discussion. I'm 100% in favor of merging this as is 👍

PR #424, opened 2022-10-06
[Merged by Bors] - feat: upload custom tls certificate for api calls (PL-120)

AhmedVF:
**Fixes or implements PL-120**

This PR does two things:
1. During runtime, when an API request needs to be made, we check to see if the corresponding node as any tls info. If it does, we use it to read the certificate and key from AWS S3, which we then use to create an agent which is included in the request. This allows us to use the certificate and key someone uploads to engage in a mutual tls connection with their server.
2. When editing their diagram, there is an option for someone to make a request to an API. Previously, `creator-app` made a request to the `creator-api`, which in turn made a request to `custom-api` which finally made a request to the user's API. Now, `creator-app` makes a request to `general-runtime` which follows the exact same steps to make the request as it does at runtime.

- Add a createAgent function which is responsible for creating the agent with the tls info. This agent is then passed as a parameter to the request.
- Added some helper functions to create the S3 client and read from the bucket. These are used in the createAgent function.
- Added a new endpoint on `general-runtime` that `creator-app` calls when making the request from the editor.

### Setup information

Add the following env var. `S3_TLS_BUCKET="com.voiceflow.dev.api.images"`

### Deployment Notes

This requires the env var above to also be added for the production environment, but that has already been done in coordination with the infra team.

### Related PRs

https://github.com/voiceflow/general-service/pull/416

### Checklist

- [x] any new env vars have been added to the [notion doc](https://www.notion.so/voiceflow/Add-Environment-Variables-be1b0136479f45f1adece7995a7adbfb) and infra has been notified


.env.e2e
```diff
@@ -7,6 +7,7 @@ AWS_ACCESS_KEY_ID='null'
 AWS_SECRET_ACCESS_KEY='null'
 AWS_ENDPOINT='http://localstack.test.e2e:8000'
 AWS_REGION='localhost'
+S3_TLS_BUCKET='com.voiceflow.dev.api.images'
```
trs, line 10: I see a [request to create `API_IMAGE_BUCKET`](https://www.notion.so/voiceflow/Add-env-var-general-runtime-API_IMAGE_BUCKET-92cc2130af1e4402afbed34b71fbf2c5), but nothing about `S3_TLS_BUCKET`.
I don't think the general runtime needs `API_IMAGE_BUCKET`, but it does need `S3_TLS_BUCKET`.

runtime/lib/Handlers/api/utils.ts
```diff
@@ -3,15 +3,28 @@ import { object } from '@voiceflow/common';
...
 import FormData from 'form-data';
+import { Agent } from 'https';
 import ipRangeCheck from 'ip-range-check';
 import safeJSONStringify from 'json-stringify-safe';
-import _merge from 'lodash/merge';
 import fetch, { BodyInit, Headers, Request, Response } from 'node-fetch';
 import { setTimeout as sleep } from 'timers/promises';
 import validator from 'validator';
 
+import CONFIG from '@/config';
```
trs, line 13: This cannot be accessed here, since everything in the `runtime` folder is a library that other runtimes use. This is the reason for [the pattern of passing the values though when initializing the handler](https://github.com/voiceflow/general-runtime/pull/424/files#diff-de87f9769e26c15dc082d4959f9a1bfd5fe811a2a8d1e691b66e67d70b90ad66L35).

This will have to be changed to follow the existing patterns.

trs (changes requested):

runtime/lib/Handlers/api/utils.ts
jonahsnider: note: i came up with this config code while pairing with ahmed. and totally forgot about where we're allowed to access `CONFIG`. sorry about that ahmed

how would you recommend properly passing the config here? make it a parameter to all the exported functions? and then pass `CONFIG` to the handler and then to the util functions, or, in the case of the new HTTP routes (which don't use the handler), just access `CONFIG` directly and pass it through

.env.e2e
AhmedVF: Sorry, you raise a good point. I originally asked infra to call the bucket `API_IMAGE_BUCKET` but we collectively decided on a new name. I didn't bother updating the request though since it had already been created, but I will do that for completeness.
Here's the slack convo if you're interested. https://voiceflowhq.slack.com/archives/C0411RY00BY/p1665072638157869

runtime/lib/Handlers/api/utils.ts
AhmedVF: Okay, let me try to get back to how it was before.

.env.e2e
AhmedVF: These docs have been updated.

lib/utils.ts
```diff
@@ -74,3 +75,15 @@ export const isConstructor = <T extends AnyClass>(clazz: T | unknown): clazz is
     return false;
   }
 };
+
+export const createResponseConfig = (config: Config): ResponseConfig => {
```
trs, line 79: This is not a generic function, it is very specific to the api block, so it should not live in a generic utils place. This should probably just be applied in the api handler itself.

runtime/lib/Handlers/api/utils.ts
```diff
@@ -3,15 +3,17 @@ import { object } from '@voiceflow/common';
 import VError from '@voiceflow/verror';
```
trs, line 3: Outside of this PR, this whole file needs a revamp.

lib/utils.ts
jonahsnider: is the API handler allowed to import the `Config` type?

backend/api/index.ts
```diff
@@ -13,6 +14,7 @@ export default (middlewares: MiddlewareMap, controllers: ControllerMap) => {
   router.use('/interact', InteractRouter(middlewares, controllers));
   router.use('/state', StateRouter(middlewares, controllers));
   router.use('/public', PublicRouter(middlewares, controllers));
+  router.use('/api', APIRouter(middlewares));
```
trs, line 17: We should be a bit more specific with the route, `/api` is very generic.

trs (approved):

lib/utils.ts
trs: It is not, the `Config` type is at the app level, not the library level.

backend/api/index.ts
AhmedVF: Updated to use `/test/api`

PR #425, opened 2022-10-07
refactor: pass config in directly (PL-120)

trs:
**Fixes or implements PL-120**

The `createResponseConfig` was not a generic utility, so should not have lived where it was.
By removing the defaults from the exported library package portion, it would mean each runtime that is dependant on this library portion would also have to add defaults themselves. I've attempted to retain these defaults.

### Related PRs

### Checklist


trs:
I realize this will have impacts on the other runtimes as well. Those would have to change to pass in the defaults as well. They get the defaults "for free" right now since the runtime library portion is in control of that currently. These changes move out the defaults to the app side.

AhmedVF (approved):

PR #426, opened 2022-10-11
[Merged by Bors] - fix: replace carousel button variables (CT-1055)

DecathectZero:
super simple, but nested attributes of carousel buttons need to get `{....}` replaced with the actual value of the variable.

Screen Shot 2022-10-11 at 7 17 36 PM
Screen Shot 2022-10-11 at 7 18 10 PM

otherwise it just resolves as `{url}` or something

### DIFF
--- a/lib/services/runtime/handlers/carousel.ts
+++ b/lib/services/runtime/handlers/carousel.ts
@@ -1,5 +1,5 @@
 import { BaseNode, BaseTrace } from '@voiceflow/base-types';
-import { replaceVariables, sanitizeVariables } from '@voiceflow/common';
+import { deepVariableSubstitution, replaceVariables, sanitizeVariables } from '@voiceflow/common';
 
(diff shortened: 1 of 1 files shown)


viniciusdacal (approved):

lib/services/runtime/handlers/carousel.ts
```diff
@@ -45,8 +45,7 @@ export const CarouselHandler: HandlerFactory<BaseNode.Carousel.Node, typeof hand
             text,
           },
           buttons: card.buttons.map((button) => ({
-            ...button,
-            name: replaceVariables(button.name, variablesMap),
+            ...deepVariableSubstitution(button, variablesMap),
```
viniciusdacal, line 48: any reason to keep spread?
 we could just return 
```ts 
deepVariableSubstitution(button, variablesMap)
```

viniciusdacal (approved):

trs (approved):

PR #427, opened 2022-10-12
[Merged by Bors] - feat: remap vf intents to channel intents when predicting intents (PL-125)

pmvrmc:
**Fixes or implements PL-125**

Whatever we predict needs to match what the version.prototype.model.intents.name (which is used to compile the prototype-program)

for google/dfes it’s easy. We actually never mapped intent names (all prototype-programs store intents like actions.intent.YES), so regardless what LUIS returns, we always want need to map from VF to Google intents (falling back to whatever the intent name was if no match on map). Worst case scenario we do an extra operation to search the map, because the model has been rerun and it already returns actions.intent.YES

for alexa we do map intent names, but once we stop calling alexa-service/renderSync and start calling general-service/renderSync  that mapping stops. So we store a variable under `project.platformData.hasChannelIntents` whenever general-service/renderSync is called for alexa projects. If said variable is true, then when predicting we map from VF to Alexa Intents. Again, worst case scenario we do an extra operation to search the map, because the model has been rerun and it already returns AMAZON.YesIntent.

### Related PRs

- https://github.com/voiceflow/creator-api/pull/1095
- https://github.com/voiceflow/libs/pull/376
- https://github.com/voiceflow/general-service/pull/418

### Checklist


lib/services/nlu/utils.ts
```diff
@@ -11,3 +15,30 @@ export const getNoneIntentRequest = (query = ''): BaseRequest.IntentRequest => (
     entities: [],
   },
 });
+
+export const mapChannelData = (data: any, platform: VoiceflowConstants.PlatformType, hasChannelIntents?: boolean) => {
+  // google/dfes intents were never given meaningful examples untill https://github.com/voiceflow/general-service/pull/379 was merged
+  // this means that sometimes we might predict a VF intent when it should be a google one
+
+  // alexa intents were given some but not exhaustive examples untill https://github.com/voiceflow/general-service/pull/379 was merged
+  // this means old programs will hold VF intents, new ones wil hold channel intents
```
zhihil, line 24: My memory of the Adapters work is a bit hazy, so I'm just double-checking if this assumption is right: "this means old programs will hold VF intents, new ones wil hold channel intents" 

I'm assuming before Adapters work started, we were mapping channel intents to VF intents for a prototype program. I guess we removed this code during the Adapters work.

Do you remember where this code was?

lib/services/nlu/utils.ts
```diff
@@ -11,3 +15,30 @@ export const getNoneIntentRequest = (query = ''): BaseRequest.IntentRequest => (
...
+  const mapToUse = match(platform)
+    .with(VoiceflowConstants.PlatformType.GOOGLE, () => GoogleConstants.VOICEFLOW_TO_GOOGLE_INTENT_MAP)
+    .with(VoiceflowConstants.PlatformType.DIALOGFLOW_ES, () => GoogleConstants.VOICEFLOW_TO_GOOGLE_INTENT_MAP)
```
zhihil, line 27: Using `GoogleConstants.VOICEFLOW_TO_GOOGLE_INTENT_MAP` will miss the mapping from `actions.intent.fallback` into whatever our `VF.fallback` is. 

I changed the DFES templates a while ago so that they include `actions.intent.fallback`, so all recent DFES projects include only this intent. 

See`libs/packages/google-dfes-types/src/constants/intents.ts` for the DFES intents.

We'll need to create a "merged map" consisting of both the Google intents used in the Google Actions project type and all Google intents used in the DFES project types.

lib/services/nlu/utils.ts
pmvrmc: we still map to VF intents for alexa only. creator-app calls alexa-service which maps and thens calls general-service (https://github.com/voiceflow/alexa-service/blob/master/lib/services/prototype/index.ts)

for google/dfes we never mapped the intents (maybe buggy implementation, fn is there but doesnt do a thing https://github.com/voiceflow/google-service/blob/master/lib/services/prototype/index.ts AND https://github.com/voiceflow/google-service/blob/master/lib/services/dialogflow/es/prototype/index.ts)

when we swicth the creator-app to not call alexa/google-service anymore when rendering but calling the general-service, these programs will hold different names

lib/services/nlu/utils.ts
pmvrmc: I dont think Fallback from DFES is the same thing as no intent.

idea here with google maps is to improve the prototype-experience for google, atm for a google project the prototype-tool wont recognize "yes" as "actions.intent.Yes" but as "VF.YES" which breaks the flow.

Also afaik you cant trigger a "actions.intent.fallback" from the prototype-tool so I dont think we need to add it. do you agree?

pmvrmc: after some more testing, we cant map `none` into `fallback` otherwise if we have a choice block the no match wont work. Instead of following the no match path, it will jump to the fallback.

lib/services/nlu/utils.ts
```diff
@@ -11,3 +15,33 @@ export const getNoneIntentRequest = (query = ''): BaseRequest.IntentRequest => (
...
+    .with(VoiceflowConstants.PlatformType.DIALOGFLOW_ES_CHAT, () => GoogleConstants.VOICEFLOW_TO_GOOGLE_INTENT_MAP)
+    .with(VoiceflowConstants.PlatformType.DIALOGFLOW_ES_VOICE, () => GoogleConstants.VOICEFLOW_TO_GOOGLE_INTENT_MAP)
+    .with(VoiceflowConstants.PlatformType.ALEXA, () => {
+      if (hasChannelIntents) return AlexaConstants.VoiceflowToAmazonIntentMap;
```
pmvrmc, line 32: consider fallback<->none inside choice block and/or intent capture to see if they are really mappable

check /next/pause/etc (stream actions)

zhihil (approved):
Screen Shot 2022-10-19 at 10 48 47 AM

I would just make sure to update the strings in `general-runtime/tests/lib/services/nlu/index.unit.ts`, but code-wise looks good to me.

PR #428, opened 2022-10-17
[Merged by Bors] - feat: add get publishing endpoint (CT-000)

DecathectZero:

Added a new endpoint: 
`GET /public/:projectID/publishing` - this fetches the `version.platformData.publishing` field of a PUBLIC project

because of the `resolvePublicProjectID` middleware, the project's `project.apiPrivacy` must be public to get this information. This is important to expose because it allows us to fetch project settings and metadata from the adapters/clients.

### DIFF
--- a/backend/api/routers/public.ts
+++ b/backend/api/routers/public.ts
@@ -9,12 +9,16 @@ export default (middlewares: MiddlewareMap, controllers: ControllerMap) => {
 
   router.use(bodyParser.json({ limit: BODY_PARSER_SIZE_LIMIT }));
 
-  router.use([middlewares.project.resolvePublicProjectID, middlewares.rateLimit.versionConsume]);
+  const interactMiddleware = [middlewares.project.resolvePublicProjectID, middlewares.rateLimit.versionConsume];
 
   // full route: /public/:projectID/state/user/:userID/interact
-  router.post('/:projectID/state/user/:userID/interact', controllers.stateManagement.publicInteract);
+  router.post(
+    '/:projectID/state/user/:userID/interact',
+    interactMiddleware,
+    controllers.stateManagement.publicInteract
+  );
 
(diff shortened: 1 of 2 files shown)


backend/api/routers/public.ts
```diff
@@ -9,14 +9,12 @@ export default (middlewares: MiddlewareMap, controllers: ControllerMap) => {
 
   router.use(bodyParser.json({ limit: BODY_PARSER_SIZE_LIMIT }));
 
-  const interactMiddleware = [middlewares.project.resolvePublicProjectID, middlewares.rateLimit.versionConsume];
```
pmvrmc, line 12: question: why use router.use instead of always passing the middleware array?

While more verbose, having the array makes it more explicit which middlewares run on the route.

lib/controllers/stateManagement/index.ts
```diff
@@ -68,6 +68,17 @@ class StateManagementController extends AbstractController {
...
 
+  @validate({
+    HEADERS_VERSION_ID: VALIDATIONS.HEADERS.VERSION_ID,
+  })
+  async getPublicPublishing(req: Request<{ userID: string }, never, { versionID: string }>) {
+    const api = await this.services.dataAPI.get();
+
+    const version = await api.getVersion(req.headers.versionID);
+
+    return version.platformData.publishing || {};
```
pmvrmc, line 79: maybe using the null coalescing operator (??) instead of OR (||) ?

pmvrmc (approved):
looks good! just some minor code style questions

lib/controllers/stateManagement/index.ts
DecathectZero: @pmvrmc I think it's important that this is always an object. `false` or `null` might trigger errors because it is a `Record<string, unknown>` type on the adapter.

PR #433, opened 2022-10-31
[Merged by Bors] - feat: use user id as transcript sessionID [bugfix] (CT-1050)

DecathectZero:
**Fixes or implements CT-1050**

Now logs are distinct based on `projectID` + `sessionID` and not just `sessionID`. 

The current problem is on the logs, all sessionIDs are saved as this opaque format:
```(metadata.state?.variables ? `${versionID}.${metadata.state.variables.user_id}` : versionID);```

And services calling this would have no idea how to retrieve the logs for a session. All `general-runtime` stateful API calls fills the `userID` field in the `interaction` context.
Instead of saving all our logs as this `${versionID}.${metadata.state.variables.user_id}` format we should just save it as `metadata.userID`

This doesn't affect the only place we expose transcripts to users: the prototype tool, because `metadata.data.reqHeaders?.sessionid` always gets declared there.

It just means all log database entries going forward will just directly user the userID instead of a compound entry with the `versionID` as well.

### DIFF
--- a/lib/clients/analytics.ts
+++ b/lib/clients/analytics.ts
@@ -50,10 +50,7 @@ export class AnalyticsSystem extends AbstractClient {
     metadata: Context;
     timestamp: Date;
   }): GeneralInteractionBody {
-    const sessionID =
-      metadata.data.reqHeaders?.sessionid ??
-      metadata.userID ??
-      (metadata.state?.variables ? `${versionID}.${metadata.state.variables.user_id}` : versionID);
+    const sessionID = metadata.data.reqHeaders?.sessionid ?? metadata.userID ?? versionID;
 
     return {
       projectID,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: lib/clients/analytics.ts [hunk 1]
```diff
@@ -50,9 +50,7 @@ export class AnalyticsSystem extends AbstractClient {
     metadata: Context;
     timestamp: Date;
   }): GeneralInteractionBody {
-    const sessionID =
-      metadata.data.reqHeaders?.sessionid ??
-      (metadata.state?.variables ? `${versionID}.${metadata.state.variables.user_id}` : versionID);
```
DecathectZero, line 55: `metadata.state.variables.user_id` is declared here: https://github.com/voiceflow/general-runtime/blob/64ffdc86bed3ac11b8f9b28ec45fdcdcdd5de1ae/lib/services/runtime/index.ts#L75-L78

It is the same as `metadata.userID`

jonahsnider (approved):

PR #439, opened 2022-11-04
[Merged by Bors] - feat: implement support to global no-match no-reply (CT-1144)

viniciusdacal:
**Fixes or implements CT-1144**

 - Inject version into runtime, to be accessible from no-match and no-reply handlers.
 - Update no match handler to support global no-match
 - Update no reply handler to support global no-reply
 - Update add no reply timeout function, to account for global no-reply delay.


runtime/lib/Runtime/index.ts
```diff
@@ -60,12 +68,14 @@ class Runtime<R = any, DA extends DataAPI = DataAPI, S extends AnyRecord = AnyRe
...
   constructor(
     public versionID: string,
     state: State,
-    private request: R | null = null,
-    { services = {} as S, handlers = [], api }: Options<DA, S>,
-    events: Lifecycle
+    private request: Request | null = null,
+    { services = {} as Services, handlers = [], api }: Options<DataAPI, Services>,
+    events: Lifecycle,
+    public version?: Version
```
DecathectZero, line 78: maybe we just turn it into one big single param that is an object?

lib/services/runtime/handlers/noMatch.ts
```diff
@@ -57,9 +66,15 @@ export const NoMatchHandler = (utils: typeof utilsObj) => ({
       payload: { path: 'reprompt' },
     });
 
-    const output = node.noMatch?.randomize
-      ? _.sample<string | BaseText.SlateTextValue>(nonEmptyNoMatches)
-      : nonEmptyNoMatches?.[noMatchCounter];
+    let output: string | BaseText.SlateTextValue | undefined;
```
DecathectZero, line 69: if output is undefined we shouldn't be adding the trace

DecathectZero (approved):

DecathectZero (approved):

PR #446, opened 2022-11-20
[Merged by Bors] - fix: add default variables (PL-000)

DecathectZero:

This is a quality-of-life fix to populate default variables we already expose on Voiceflow's creator tool: (brought up by @Diophontine)
`sessions`: number of new sessions the user has started (empty stack)
`timestamp`: UNIX time in seconds when the turn started
`user_id`: stateful API user ID in the URL param

Currently, for prototyping they aren't made use of - but it is important for production purposes.

Screen Shot 2022-11-20 at 3 15 20 PM

The timestamp mirrors the functionality in the other runtimes, where it is the UNIX timestamp down to the second, not the millisecond:
https://github.com/voiceflow/alexa-runtime/blob/c93882892df5207e5eaef1629f33c800b5147163/lib/services/alexa/request/lifecycle/update.ts#L18

### DIFF
--- a/lib/services/state/index.ts
+++ b/lib/services/state/index.ts
@@ -1,4 +1,5 @@
 import { BaseModels, BaseRequest, BaseTrace } from '@voiceflow/base-types';
+import _ from 'lodash';
 
 import { PartialContext, State } from '@/runtime';
 import { Context, InitContextHandler } from '@/types';
@@ -20,7 +21,7 @@ class StateManager extends AbstractManager<{ utils: typeof utils }> implements I
    * generate a context for a new session
    * @param versionID - project version to generate the context for
    */
-  generate({ prototype, rootDiagramID }: BaseModels.Version.Model<any>, state?: State): State {
+  generate({ prototype, rootDiagramID }: BaseModels.Version.Model<any>, state?: State, userID?: string): State {
     const DEFAULT_STACK = [{ programID: rootDiagramID, storage: {}, variables: {} }];
 
     const stack =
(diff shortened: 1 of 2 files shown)


trs (approved):

lib/services/state/index.ts
```diff
@@ -30,12 +33,18 @@ class StateManager extends AbstractManager<{ utils: typeof utils }> implements I
         variables: frame.variables || {},
       })) || DEFAULT_STACK;
 
+    const variables = {
+      ...prototype?.context.variables,
+      ...state?.variables,
+    };
+
+    // new session default variables
+    variables.sessions = (_.isNumber(variables.sessions) ? variables.sessions : 0) + 1;
```
pmvrmc, line 42: why not move this part to the variables initialization? just to tidy up the code

lib/services/state/index.ts
```diff
@@ -44,16 +53,21 @@ class StateManager extends AbstractManager<{ utils: typeof utils }> implements I
...
+    const entities = version.prototype?.model.slots.map(({ name }) => name) || [];
+
+    const variables = {
+      ...initializeStore(entities),
+      ...initializeStore(version.variables),
+      ...state.variables,
+    };
+
+    // new turn default variables
+    variables.timestamp = this.services.utils.getTime(); // unix time in seconds
```
pmvrmc, line 66: very minor, but again, maybe move this to the variables initialization?

pmvrmc (approved):

lib/services/state/index.ts
DecathectZero: this needs to happen under `generate` because `generate` gets called on new sessions (stack is empty)

`initializeVariables` gets called every turn and it's important that we don't update the session count every turn.

lib/services/state/index.ts
DecathectZero: this is already under `initializeVariables`. I can clean it up a little but more perhaps.

PR #448, opened 2022-11-28
[Merged by Bors] - feat: allow updating a transcripts without overriding all of its previous properties (PL-296)

pmvrmc:
**Fixes or implements PL-296**

For threaded platforms (like whatsapp) a transcript is everlasting and since we dont have a "conversation start" event, we have to resort to always calling create transcript on every interaction.

That means on the creat transcript endpoint we need to not override the `unread`, `reportsTags` and `createdAt`.

### Related PRs

- https://github.com/voiceflow/database-cli/pull/295
- https://github.com/voiceflow/platform/pull/309
- 
### Checklist

### DIFF
--- a/lib/controllers/public/index.ts
+++ b/lib/controllers/public/index.ts
@@ -75,10 +75,19 @@ class PublicController extends AbstractController {
     const { mongo } = this.services;
     if (!mongo) throw new Error('mongo not initialized');
 
-    const data: {
-      projectID: ObjectId;
-      sessionID: string;
-      createdAt?: Date;
+    const filter = {
+      projectID: new ObjectId(projectID),
+      sessionID,
+    };
+
+    const insertData = {
+      ...filter,
+      createdAt: new Date(),
+      unread: true,
+      reportTags: [],
+    };
+
+    const updateData: {
       updatedAt: Date;
       os?: string | undefined;
       device?: string | undefined;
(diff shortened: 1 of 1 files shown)


lib/controllers/public/index.ts
```diff
@@ -89,19 +102,26 @@ class PublicController extends AbstractController {
           ...(user.image && { image: user.image }),
         },
       }),
-      unread: true,
-      reportTags: [],
     };
 
-    const { value } = await mongo.db
+    const oldTranscript = await mongo.db
```
trs, line 107: We should be able to do this with a single `fineOneAndUpdate` query.

```ts
    const filter = {
      projectID: new ObjectId(projectID),
      sessionID,
    };

    const insertData = {
      ...filter,
      createdAt: new Date(),
      unread: true,
      reportTags: []
    };

    const updateData: {
      updatedAt: Date;
      os?: string | undefined;
      device?: string | undefined;
      browser?: string | undefined;
      user?: {
        name?: string | undefined;
        image?: string | undefined;
      };
    } = {
      updatedAt: new Date(),
      ...(os && { os }),
      ...(device && { device }),
      ...(browser && { browser }),
      ...(user && {
        user: {
          ...(user.name && { name: user.name }),
          ...(user.image && { image: user.image }),
        },
      }),
    };

    const { value: newTranscript } = await mongo.db
      .collection('transcripts')
      .findOneAndUpdate(
        filter,
        {
          $set: updateData,
          $setOnInsert: insertData
        },
        { upsert: true, returnOriginal: false }
      );
```

pmvrmc: oh, I didnt know about $setOnInsert, very handy indeed, thx for the tip!

PR #449, opened 2022-12-05
[Merged by Bors] - feat: nlc entity filling (PL-000)

DecathectZero:

this fix requires a long description, but the code is pretty hard to navigate as well so I'll do my best to explain.
I've noticed terrible performance for the capture step/entity filling feature if the model is untrained. We should have a decent fallback measure if the NLP isn't working.

[NLC](https://github.com/voiceflow/natural-language-commander) is a library we use to handle regex matching against intent utterances and extracting entities when the NLP is untrained. It will only match if the user types everything correctly letter for letter.

In the current implementation of entity filling (called dialog management in the code atm, a bit confusing I know, just assume dialog management = entity filling), if we know the user is trying to fill a SPECIFIC entity of an intent, we first make a match of [their intent against the general model](https://github.com/voiceflow/general-runtime/blob/8e22616890c8c4d011ea12845ae463df788d34e4/lib/services/nlu/index.ts#L85-L115). 

Then we also check it against a `dmPrefix`'ed intent utterances:
Screen Shot 2022-12-05 at 3 56 39 PM
Screen Shot 2022-12-05 at 4 03 47 PM

These are special utterances tied to the intent, the prefix is just a hash of the intent name. When we know the user is trying to fill for a specific intent, we call the NLP and prepend the prefix to their raw text query:
https://github.com/voiceflow/general-runtime/blob/8e22616890c8c4d011ea12845ae463df788d34e4/lib/services/dialog/index.ts#L125-L128

So if they said "quote for life insurance" as `query` we would first run it against the NLP normally and then against "1f79373e6d quote for life insurance" as `query`. The LUIS model already has these utterances with prefixes baked in, the NLC does not. 

Also our [predict](https://github.com/voiceflow/general-runtime/blob/8e22616890c8c4d011ea12845ae463df788d34e4/lib/services/nlu/index.ts#L24) code comes in 3 stages:
1. try against the NLC regex but hard matching for slots. If the utterance is something like "quote for {insuranceType}" only valid forms of the slot are okay. They can't say "quote for toilet paper" and expect NLC to match
2. call the actual `luis-authoring-service` endpoint and check against the actual NLP
3. if the user never trained their model, try against NLC regex but open matching flow slots, If the utterance is something like "quote for {insuranceType}" then  "quote for toilet paper" will match the intent and insuranceType = "toilet paper"

What this PR does is if we KNOW that we're looking for the entity of a specific intent, we pass that through when generating the NLC model, and inject that intent's slots' (`intent.slots[].dialog.utterances`) utterances, with prefixes, into the actual utterances of the intent.

[we have this `handleDialog` function for NLC]( "quote for toilet paper"), but looked through the source code, it never worked. It relies on `slot.dialog.utterances` which was never getting passed in:
https://github.com/voiceflow/general-runtime/blob/15a8cfbad416a14af7c194bc59b1516325b50ea3/lib/services/nlu/nlc.ts#L155-L162
`filledEntities` never contains the `dialog.utterances` required, so this would be `NoneIntent` all the time anyways.

TLDR, if we are looking for a specific intent's slots' utterances, just add them to the regex model with a prefix. 

In action:
Screen Shot 2022-12-05 at 3 47 43 PM

### DIFF
--- a/tests/lib/services/nlu/nlc.unit.ts
+++ b/tests/lib/services/nlu/nlc.unit.ts
@@ -36,7 +36,7 @@ describe('nlu nlc service unit tests', () => {
       const openSlot = 'openSlot';
       const dmRequest = 'dmRequest';
 
-      expect(createNLC({ model, locale, openSlot } as any)).to.eql(nlcObj);
+      expect(createNLC({ model, locale, openSlot, dmRequest } as any)).to.eql(nlcObj);
       expect(NLCStub.callCount).to.eql(1);
       expect(registerSlotsStub.args).to.eql([[nlcObj, model, openSlot]]);
       expect(registerIntentsStub.args).to.eql([[nlcObj, model, dmRequest]]);
--- a/tests/lib/services/nlu/nlc.unit.ts
+++ b/tests/lib/services/nlu/nlc.unit.ts
@@ -9,7 +9,6 @@ import * as nlc from '@/lib/services/nlu/nlc';
 import {
   createNLC,
   handleNLCCommand,
-  handleNLCDialog,
   nlcToIntent,
   registerBuiltInIntents,
   registerSlots,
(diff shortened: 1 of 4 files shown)


tests/lib/services/nlu/nlc.unit.ts
```diff
@@ -316,4 +250,116 @@ describe('nlu nlc service unit tests', () => {
       ]);
     });
   });
+
+  describe.only('registerIntents', () => {
```
trs, line 254:
```suggestion
  describe('registerIntents', () => {
```

trs (approved):

tests/lib/services/nlu/nlc.unit.ts
DecathectZero: WHOOOPSSSS

PR #458, opened 2022-12-23
feat: hack week runtime auto complete

DecathectZero:


lib/services/runtime/handlers/utils/generativeNoMatch.ts
```diff
@@ -0,0 +1,36 @@
+import axios from 'axios';
+
+import Config from '@/config';
+import Transcript from '@/lib/services/transcript';
+import { Runtime } from '@/runtime';
+
+export const generateNoMatch = async (runtime: Runtime): Promise<string | null> => {
+  if (!Config.ML_GATEWAY_ENDPOINT) return null;
```
frankgu968, line 8: Add a // FIXME to refactor into a client later so we don't forget

lib/services/runtime/handlers/utils/generativeNoMatch.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (input) acc += `User: ${input}\n`;
+    if (output) acc += `AI: ${output}\n`;
+    return acc;
+  }, '');
+
+  if (!parsedTranscript) return null;
+
+  const autoCompleteEndpoint = `${ML_GATEWAY_ENDPOINT}/api/v1/generation/autocomplete`;
+
+  console.log({ parsedTranscript });
```
frankgu968, line 25: We should add in datadog tracking as well. the transcript responses are being recorded by the ingest service right?

DecathectZero: no that ingest service doesn't allow us to fetch transcripts, only write them.
Also there's no efficient way to fetch the 5 most recent transcripts or something

this actually stores it in the user's session state and limits it to 3 turns max.

PR #463, opened 2023-01-05
[Merged by Bors] - feat: added transcript call authorized via api key (PL-363)

zhihil:
**Fixes or implements PL-363**

Adds a new endpoint that behaves similar to the old `controllers.public.createTranscript`, except that it is authorized via DM API key rather than by checking the `apiPrivacy` field on the `project` document.

The `apiPrivacy` field is somewhat tricky to set. The natural place where it should be set is either with a UX toggle or as a part of the channel publish. However,

1. UX toggle requires designs from Mike which is out-of-scope at the moment
2. A custom Teams publish job that sets `apiPrivacy` requires a lot of boilerplate to setup the job on `general-service`, which is far more effort and more error-prone than desired for this small change. 

Added a new endpoint `POST /transcripts/:projectID` 

 - `TranscriptManager` in `lib/services/transcript/index.ts` implements the business logic to upsert a transcript that used to be part of `controllers.public.createTranscript`
 - `controllers.public.upsertTranscript`, `controllers.transcript.upsertTranscript` will now call the `upsertTranscript` method of `TranscriptManager` 
 - `controllers.public.upsertTranscript`, `controllers.transcript.upsertTranscript` are guarded by different middleware. The former checks for `apiPrivacy` to have value `'public'` whereas the latter will check the `Authorization` header for a valid DM API key. 

Renamed a different, existing `Transcript` service (that was created for the AI Assist feature to add the transcript to the `context`) to avoid naming conflict with the `TranscriptManager` introduced by this PR. 

### Setup information

None

### Deployment Notes

None

### Related PRs

- https://github.com/voiceflow/platform/pull/344
- https://github.com/voiceflow/general-runtime/pull/463
- https://github.com/voiceflow/creator-app/pull/6531

### Checklist

- [x] changes have been validated in an ephemeral environment
- [ ] ~~this is a breaking change and should publish a new major version~~
- [ ] ~~any new env vars have been added to the [notion doc](https://www.notion.so/voiceflow/Add-Environment-Variables-be1b0136479f45f1adece7995a7adbfb) and infra has been notified~~

### DIFF
--- a/backend/api/routers/transcript.ts
+++ b/backend/api/routers/transcript.ts
@@ -1,0 +1,21 @@
+import express from 'express';
+
+import { BODY_PARSER_SIZE_LIMIT } from '@/backend/constants';
+import { ControllerMap, MiddlewareMap } from '@/lib';
+
+// stateful API routes
+export default (middlewares: MiddlewareMap, controllers: ControllerMap) => {
+  const router = express.Router();
+
+  router.use(express.json({ limit: BODY_PARSER_SIZE_LIMIT }));
+
+  const middleware = [
+    middlewares.project.resolveVersionAlias,
+    middlewares.project.attachProjectID,
+    middlewares.rateLimit.versionConsume,
+  ];
+
+  router.post('/:projectID', middleware, controllers.transcript.createTranscript);
+
+  return router;
+};
--- a/backend/api/routers/transcript.ts
+++ b/backend/api/routers/transcript.ts
@@ -15,7 +15,7 @@ export default (middlewares: MiddlewareMap, controllers: ControllerMap) => {
     middlewares.rateLimit.versionConsume,
   ];
 
-  router.post('/:projectID', middleware, controllers.transcript.createTranscript);
+  router.post('/:projectID', middleware, controllers.transcript.upsertTranscript);
 
   return router;
 };
--- a/lib/services/aiAssistContext.ts
+++ b/lib/services/aiAssistContext.ts
@@ -10,7 +10,7 @@ const MAX_TURNS = 3;
 
 // writes a primative string transcript into the context state storage
 class AIAssistContext extends AbstractManager implements ContextHandler {
-  static StorageKey = 'transcript';
+  static StorageKey = 'aiAssistTranscript';
 
   static getInput(request: RuntimeRequest) {
     return (isIntentRequest(request) && request.payload.query) || (isTextRequest(request) && request.payload) || null;
--- a/lib/services/aiAssistContext.ts
+++ b/lib/services/aiAssistContext.ts
@@ -9,7 +9,7 @@ import { AbstractManager } from './utils';
 const MAX_TURNS = 3;
 
 // writes a primative string transcript into the context state storage
-class Transcript extends AbstractManager implements ContextHandler {
+class AIAssistContext extends AbstractManager implements ContextHandler {
   static StorageKey = 'transcript';
 
   static getInput(request: RuntimeRequest) {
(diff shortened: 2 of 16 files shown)


lib/services/aiAssistContext.ts
```diff
@@ -9,8 +9,8 @@ import { AbstractManager } from './utils';
 const MAX_TURNS = 3;
```
zhihil, line 9: Renamed this file because this `transcript.ts` was built over the holidays for the AI Assist project and the naming conflicts with my own transcript manager service.

backend/api/routers/transcript.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  router.use(express.json({ limit: BODY_PARSER_SIZE_LIMIT }));
+
+  const middleware = [
+    middlewares.project.resolveVersionAlias,
+    middlewares.project.attachProjectID,
+    middlewares.rateLimit.versionConsume,
+  ];
+
+  router.post('/:projectID', middleware, controllers.transcript.upsertTranscript);
```
zhihil, line 18: Underlying logic is identical to the endpoint in backend/api/routers/public.ts except that we authorize the request using an API key rather than the apiPrivacy property in the project document, which is somewhat involved to set.

lib/services/transcript/index.ts
```diff
@@ -0,0 +1,56 @@
...
+import { AbstractManager } from '../utils';
+import { TranscriptClientInfo, TranscriptUpdatePayload } from './types';
+
+class TranscriptManager extends AbstractManager {
+  async upsertTranscript(
+    projectID: string,
+    sessionID: string,
+    clientInfo: DeepPartial<TranscriptClientInfo>,
+    unread?: boolean
+  ) {
```
zhihil, line 13: This method is identical to `createTranscript` from the original public controller, except that:

1. Naming is changed to `upsert` to reflect its behaviour (function as create for discrete chat sessions, upsert for continuing chat histories)
2. An `unread` boolean is added to allow the caller to set `project.unread` (e.g. transcripts for continuing chat histories need to set `unread: true` each time it gets updated asynchronously due to recent user activity).

zhihil (commented):
Comments for context:

Added a new endpoint that calls logic to upsert a transcript. The previous endpoint on the public controller was awkward to use because you need to set the `apiPrivacy` property on `project` which is more involved than necessary.

There were other approaches considered:
1. Set `apiPrivacy` by creating a custom Teams publish job in the `general-service` (this involved too much boilerplate code setting up the Teams job)
2. Set `apiPrivacy` by making an HTTP call to update the project in the frontend (this places responsibility on the frontend to remember to set `apiPrivacy` which makes for poor abstraction, especially considering how deeply nested in the backend that the context for `apiPrivacy` is).

trs (approved):

backend/api/routers/transcript.ts
trs: Let's avoid relying on the api key to give us the project ID. Let's be explicit about it.
Going forward we cannot rely on the api key with the api key system revamp.

trs (approved):

trs:
bors merge

trs:
bors merge

trs:
bors merge

trs:
bors merge

lib/services/index.ts
```diff
@@ -56,6 +58,7 @@ const buildServices = (config: Config, clients: ClientMap): FullServiceMap => {
   services.analytics = new Analytics(services, config);
   services.stateManagement = new StateManagement(services, config);
   services.transcript = new Transcript(services, config);
+  services.aiAssist = new AIAssist(services, config);
   services.interact = new Interact(services, config);
```
pmvrmc, line 62: this was the cause for e2e tests failing!

aiAssist must be "registered" before interact, otherwise it is not available inside the interact service

PR #465, opened 2023-01-10
[Merged by Bors] - fix: allow prototype to start from buttons (BUG-81)

amandasteinhauer:
**Fixes or implements BUG-81**

Fixes bug where user would get a path not connected error if starting a prototype from a buttons step.

### DIFF
--- a/lib/services/runtime/handlers/interaction/interaction.ts
+++ b/lib/services/runtime/handlers/interaction/interaction.ts
@@ -24,9 +24,10 @@ const utilsObj = {
 export const InteractionHandler: HandlerFactory<VoiceflowNode.Interaction.Node, typeof utilsObj> = (utils) => ({
   canHandle: (node) => !!node.interactions,
   handle: (node, runtime, variables) => {
-    const isStartingFromButtonsStep = runtime.getAction() === Action.REQUEST && !runtime.getRequest();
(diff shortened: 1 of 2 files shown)


trs (commented):
Was this not fixed on the front end? https://voiceflowhq.slack.com/archives/C03TDFSLN92/p1672852291590069

lib/services/runtime/handlers/interaction.ts
```diff
@@ -28,7 +28,9 @@ export const InteractionHandler: HandlerFactory<
 > = (utils) => ({
   canHandle: (node) => !!node.interactions,
   handle: (node, runtime, variables) => {
-    if (runtime.getAction() === Action.RUNNING) {
+    const isStartingFromButtonsStep = runtime.getAction() === Action.REQUEST && !runtime.getRequest();
+
+    if (runtime.getAction() === Action.RUNNING || isStartingFromButtonsStep) {
```
trs, line 33: Could just grab the action once and avoid updating tests.

```suggestion
    const runtimeAction = runtime.getAction();
    const isStartingFromButtonsStep = runtimeAction === Action.REQUEST && !runtime.getRequest();

    if (runtimeAction === Action.RUNNING || isStartingFromButtonsStep) {
```

trs (approved):

PR #466, opened 2023-01-11
feat: implement default no-match no-reply message (CT-1384)

viniciusdacal:
**Fixes or implements CT-1384**

If the default global no-reply or no-match message has never been edited by the user, we should use a default message.

### Related PRs

### Checklist


lib/services/runtime/utils.ts
```diff
@@ -282,3 +285,11 @@ export const processOutput = (output: string | Text.SlateTextValue | undefined,
   const content = slateInjectVariables(output, sanitizedVars);
   return slateToPlaintext(content);
 };
+
+export const isSlate = (platform: string | undefined) => {
```
pmvrmc, line 289: can platforms be slate? or the response?

maybe rename this to `isSlateSupportedPlatform` ? Also aren't there more platforms that support slate text?

pmvrmc: also platform shouldnt be a string but `VoiceflowConstants.PlatformType`.

I think it should be the caller that casts to the correct type, not the util fn

DecathectZero (approved):

lib/services/runtime/handlers/noMatch/index.ts
```diff
@@ -66,7 +66,7 @@ export const getOutput = async (
     // this is fairly inefficient to fetch the project object via API every time, find better solution
     const project = await runtime.api.getProject(runtime.version.projectID).catch(() => null);
```
pmvrmc, line 67: shoudlnt we remove fetching the project here?

lib/services/runtime/handlers/utils/output.ts
```diff
@@ -0,0 +1,11 @@
+import { BaseModels } from '@voiceflow/base-types';
+import { VoiceflowConstants } from '@voiceflow/voiceflow-types';
+
+export const generateOutput = (text: string, project?: BaseModels.Project.Model<any, any>) => {
+  // TODO: exclusively use project.type after large scale migration
```
pmvrmc, line 5: didnt the migration run already? we could simplify this code

lib/services/state/index.ts
```diff
@@ -87,6 +87,7 @@ class StateManager extends AbstractManager<{ utils: typeof utils }> implements I
     const dataApi = await this.services.dataAPI.get(context.data?.reqHeaders?.authorization);
     const api = new CacheDataAPI(dataApi);
     const version = await api.getVersion(context.versionID!);
+    const project = await api.getProject(version.projectID).catch(() => undefined);
```
pmvrmc, line 90: why the catch? it feels like an antipattern, why would we be hiding errors here?

pmvrmc (commented):
lets try to address these issues, it looks to me we can improve code still

PR #470, opened 2023-01-17
feat: adapting the runtime to support dfes work (PL-32)

pmvrmc:
**Fixes or implements PL-32**

With the changes to the adapter, we can simplify the runtime code since the DFES adapter will send a payload that better matches the general-runtime one.

The state update is changed to allow saving nodeIDs with null because sometimes the runtime response to the adapter contains those nodeIDs with null. Since DFES adapter has the particularity of needing to update a state storage field on some requests, we need to push back to the runtime whatever was returned previously (so we dont change anything else other than the state storage field).

### Related PRs

- https://github.com/voiceflow/google-service/pull/334
- https://github.com/voiceflow/platform/pull/384


trs (changes requested):

trs (approved):

pmvrmc:
closing this as we are no longer planning on supporting dfes

PR #475, opened 2023-01-27
[Merged by Bors] - feat: support for alexa adapter (PL-30)

pmvrmc:
**Fixes or implements PL-30**

Changes needed for general-runtime in order to support alexa adapter.
Mostly is simplification/fixes, the only real change is adding a new property when sending a speak trace as response of an entity being unfulfilled.

This is important because the Alexa adapter must ignore this kind of prompts. In this case alexa NLU is the one handling dialog/slot filling so it will by itself push the prompt. We need a way to signal this on the trace so that Alexa adapter doesnt send the prompt.

### Related PRs

- https://github.com/voiceflow/platform/pull/346
- https://github.com/voiceflow/alexa-service/pull/318
- https://github.com/voiceflow/creator-app/pull/6773
- https://github.com/voiceflow/general-service/pull/506
- https://github.com/voiceflow/libs/pull/434


lib/services/runtime/handlers/state/stream/stream.alexa.ts
```diff
@@ -1,25 +1,45 @@
...
+// we need to account for the edge case there is a custom directive with an audio (like a stream)
+// we need to treat that as a stream
+const isDirectiveWithAudio = (node: BaseModels.BaseNode, runtime: Runtime) => {
+  const request = runtime.getRequest();
+  const intentName = isIntentRequest(request) ? request.payload.intent.name : null;
+  return (
+    node.type === BaseNode.NodeType.CHANNEL_ACTION &&
+    (node as BaseNode.ChannelAction.Node).data.name === BaseNode.NodeType.DIRECTIVE &&
+    (node as BaseNode.ChannelAction.Node).data.payload.directive.includes('AudioPlayer.Play') &&
+    intentName === AlexaConstants.AmazonIntent.PLAYBACK_NEARLY_FINISHED &&
```
pmvrmc, line 30: remove this

lib/services/runtime/handlers/state/stream/stream.alexa.ts
```diff
@@ -1,25 +1,45 @@
...
+
+// we need to account for the edge case there is a custom directive with an audio (like a stream)
+// we need to treat that as a stream
+const isDirectiveWithAudio = (node: BaseModels.BaseNode, runtime: Runtime) => {
+  const request = runtime.getRequest();
+  const intentName = isIntentRequest(request) ? request.payload.intent.name : null;
+  return (
+    node.type === BaseNode.NodeType.CHANNEL_ACTION &&
+    (node as BaseNode.ChannelAction.Node).data.name === BaseNode.NodeType.DIRECTIVE &&
+    (node as BaseNode.ChannelAction.Node).data.payload.directive.includes('AudioPlayer.Play') &&
```
pmvrmc, line 29: check if directive is set if not do optional chaining

pmvrmc: check how many projects use directive

tests/lib/services/runtime/handlers/state/stream.unit.ts
```diff
@@ -27,7 +33,10 @@ describe('stream state handler unit tests', () => {
     });
 
     it('true', () => {
-      const runtime = { storage: { get: sinon.stub().returns({ action: 'play' }) } };
+      const runtime = {
+        getRequest: sinon.stub().returns({ payload: { name: 'event1' } }),
+        storage: { get: sinon.stub().returns({ action: 'play' }) },
+      };
```
zhihil, line 39: It looks like adding `getRequest` here isn't necessary to make the tests pass.

lib/services/runtime/utils.ts
```diff
@@ -313,43 +342,5 @@ export const getDefaultNoReplyTimeoutSeconds = (platform: string | undefined) =>
...
-  // handle voice string
-  if (typeof output === 'string') {
-    return replaceVariables(output, sanitizedVars);
-  }
-
-  // handle slate text
-  const content = slateInjectVariables(output, sanitizedVars);
-  return slateToPlaintext(content);
-};
-
```
zhihil, line 353: What is the motivation for removing all of this?

zhihil (commented):
Some preliminary comments and questions. I still have 7 files left to review which are the more complex changes of this PR.

zhihil:
The capture step appears to broken if I mesh just `general-runtime` and `general-service` and run the Prototype Tool. I'm not sure if meshing the Alexa adapter matters at all. The `creator-app` changes probably don't need to be meshed either because it only affects the Stream step.

When I run the capture step, it just immediately returns without waiting to capture an utterance:

Screen Shot 2023-05-16 at 4 04 41 PM

This is the correct behaviour when I have nothing meshed:

Screen Shot 2023-05-16 at 4 06 24 PM

lib/services/runtime/handlers/event.ts
```diff
@@ -38,6 +38,7 @@ export const intentEventMatcher: Matcher<BaseRequest.IntentRequest, BaseNode.Uti
     if (request.diagramID && context.diagramID && request.diagramID !== context.diagramID) return false;
     if (!context.event || !BaseNode.Utils.isIntentEvent(context.event)) return false;
     if (context.event.intent !== request.payload.intent.name) return false;
+    if (request.payload.data) return false;
```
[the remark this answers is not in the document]
pmvrmc: since we want to reuse the intent request for both alexa events and regular ones, the main difference between them is that the alexa ones have a `data` property

tests/lib/services/runtime/handlers/noMatch.unit.ts
```diff
@@ -43,6 +43,7 @@ describe('noMatch handler unit tests', async () => {
       };
 
       const noMatchHandler = NoMatchHandler({
+        getOutput,
```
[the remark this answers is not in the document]
pmvrmc: good catch, leftovers from a previous refactor, will remove

lib/services/runtime/init.ts
```diff
@@ -67,7 +67,37 @@ const init = (client: Client) => {
             payload: {
               src,
               token,
-              action: loop ? BaseNode.Stream.TraceStreamAction.LOOP : BaseNode.Stream.TraceStreamAction.PLAY,
+              action: BaseNode.Stream.TraceStreamAction.PLAY,
```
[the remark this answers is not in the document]
pmvrmc: yeah good idea, will refactor it, maybe even with match from ts-patterns as the return of match can be directly assing to a variable

tests/lib/services/runtime/handlers/state/stream.unit.ts
pmvrmc: yeah, will remove. leftovers from previous iteration

lib/services/runtime/utils.ts
pmvrmc: this was used only by specific alexa handlers that we ported to regular handlers. I dont expect this to be missing, from my tests it has been working nicely. But lets make sure on a dev env 👍

pmvrmc:
> The capture step appears to broken if I mesh just `general-runtime` and `general-service` and run the Prototype Tool. I don't have the Alexa adapter meshed, but I don't think that should matter if I'm not actually running the Voiceflow app on Alexa. The `creator-app` changes probably don't need to be meshed either because it only affects the Stream step.
> 
> When I run the capture step, it just immediately returns without waiting to capture an utterance:
> Screen Shot 2023-05-16 at 4 04 41 PM
> 
> This is the correct behaviour when I have nothing meshed:
> Screen Shot 2023-05-16 at 4 06 24 PM
> 
> There may be other regressions like this that we need to test, so let's get the dev environment going.

weird, it worsks for me. Been trying to get the dev env ready, should be close, need some help from infra

lib/services/runtime/handlers/event.ts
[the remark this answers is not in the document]
zhihil: Add a comment like this in that case:
```ts
/**
 * Intent request is being reused for both Alexa events and general events. To distinguish them,
 * if `request.payload.data` exists, then this is an Alexa event, so we don't have an match. 
 */
```

lib/services/runtime/handlers/event.ts
```diff
@@ -50,6 +51,33 @@ export const intentEventMatcher: Matcher<BaseRequest.IntentRequest, BaseNode.Uti
...
+        ...acc,
+        ...(mapping.var && { [mapping.var]: mapVariables(mapping.path, request.payload.data!) }),
+      }),
+      {}
+    );
+
+    variables.merge(variableChanges);
+  },
+};
+
```
zhihil, line 80: Extracting out the middle part of the conditional will reduce duplication.

```ts
export const isIntentEvent = (request: BaseRequest.IntentRequest, context: any) => {
  if (!isIntentRequest(request)) return false;
  if (request.diagramID && context.diagramID && request.diagramID !== context.diagramID) return false;
  if (!context.event || !BaseNode.Utils.isIntentEvent(context.event)) return false;
  if (context.event.intent !== request.payload.intent.name) return false;
  return true;
}

export const isAlexaEvent = (request: BaseRequest.IntentRequest) => !!request.payload.data;

export const intentEventMatcher: Matcher<BaseRequest.IntentRequest, BaseNode.Utils.IntentEvent> = {
  match: (context): context is SideEffectContext<BaseRequest.IntentRequest, BaseNode.Utils.IntentEvent> => {

    if (!isIntentRequest(request)) return false;
    if (isIntentEvent(request, context)) return false;
    if (isAlexaEvent(request)) return false;

    return true;
  },
  sideEffect: (context) => (variables) => {
    // use event slot mappings map request entities to variables
    const entities = request?.payload.entities || [];

    variables.merge(mapEntities(context.event.mappings || entitiesToMappings(entities), entities));
  },
};

export const alexaEventMatcher: Matcher<BaseRequest.IntentRequest, BaseNode.Utils.AlexaEvent> = {
  match: (context): context is SideEffectContext<BaseRequest.IntentRequest, BaseNode.Utils.AlexaEvent> => {

    if (!isIntentRequest(request)) return false;
    if (isIntentEvent(request, context)) return false;
    if (!isAlexaEvent(request)) return false;

    return true;
  },
  sideEffect: (context) => (variables) => {
    const request = context.runtime.getRequest() as BaseRequest.IntentRequest;

    const variableChanges = context.event.mappings?.reduce(
      (acc, mapping) => ({
        ...acc,
        ...(mapping.var && { [mapping.var]: mapVariables(mapping.path, request.payload.data!) }),
      }),
      {}
    );

    variables.merge(variableChanges);
  },
};
```

I can't extract the `isIntentRequest(request)` part unfortunately because it's being used as a type-guard and we'll need that pass type checks.

lib/services/runtime/handlers/state/stream/stream.alexa.ts
pmvrmc: only 1211 use it, and I bet a big chunk of that is our own internal testing

zhihil (approved):

PR #477, opened 2023-02-04
[Merged by Bors] - feat: added NLU gateway env var (PL-430)

zhihil:
**Fixes or implements PL-430**

Updates `general-runtime` to call `nlu-gateway` instead of LUIS.

### Setup information

N/A

### Deployment Notes

N/A

### Related PRs

- https://github.com/voiceflow/platform/pull/407
- https://github.com/voiceflow/general-service/pull/475
- https://github.com/voiceflow/general-runtime/pull/477

### Checklist

- ~~[ ] any new env vars have been added to the [notion doc](https://www.notion.so/voiceflow/Add-Environment-Variables-be1b0136479f45f1adece7995a7adbfb) and infra has been notified~~

### DIFF
--- a/lib/services/nlu/index.ts
+++ b/lib/services/nlu/index.ts
@@ -58,6 +58,7 @@ class NLU extends AbstractManager<{ utils: typeof utils }> implements ContextHan
     hasChannelIntents,
     platform,
     dmRequest,
+    ffOptions,
   }: {
     query: string;
     model?: BaseModels.PrototypeModel;
@@ -69,6 +70,9 @@ class NLU extends AbstractManager<{ utils: typeof utils }> implements ContextHan
     hasChannelIntents: boolean;
     platform: VoiceflowConstants.PlatformType;
     dmRequest?: BaseRequest.IntentRequestPayload;
+    ffOptions: {
+      workspaceID: number;
+    };
   }): Promise<BaseRequest.IntentRequest> {
     // 1. first try restricted regex (no open slots) - exact string match
     if (model && locale) {
@@ -84,6 +88,7 @@ class NLU extends AbstractManager<{ utils: typeof utils }> implements ContextHan
         .post<NLUGatewayPredictResponse>(`${this.getNluGatewayEndpoint()}/v1/predict/${versionID}`, {
           utterance: query,
           tag,
+          ...ffOptions,
         })
         .catch(() => ({ data: null }));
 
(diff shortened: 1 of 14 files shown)


lib/services/nlu/index.ts
```diff
@@ -16,22 +16,49 @@ import { getNoneIntentRequest, mapChannelData } from './utils';
...
+        },
+        entities: prediction.predictedSlots,
+        confidence: prediction.confidence,
+      },
+    };
+  }
+
+  private getNluGatewayEndpoint() {
+    const protocol = this.config.IS_MESHED === 'true' ? 'http' : 'https';
+    return `${protocol}://${this.config.NLU_GATEWAY_SERVICE_HOST}:${this.config.NLU_GATEWAY_SERVICE_PORT_APP}`;
```
zhihil, line 48: This is needed because `NLU_GATEWAY_SERVICE_HOST` doesn't specify `http` or `https` by design. See comment in `general-service` where this code is also used.

lib/services/nlu/index.ts
```diff
@@ -106,9 +133,12 @@ class NLU extends AbstractManager<{ utils: typeof utils }> implements ContextHan
       projectID: version.projectID,
       versionID: context.versionID,
       tag: project.liveVersion === context.versionID ? VersionTag.PRODUCTION : VersionTag.DEVELOPMENT,
-      nlp: project.nlp,
+      nlp: project.prototype?.nlp,
       hasChannelIntents: project?.platformData?.hasChannelIntents,
       platform: version?.prototype?.platform as VoiceflowConstants.PlatformType,
+      ffOptions: {
+        workspaceID: Number(project.teamID),
```
zhihil, line 140: Needed to support feature flagging. We pass in the `workspaceID` (a.k.a. `teamID`) from `general-runtime` for performance and to be consistent with how `general-service` is required to pass in `workspaceID` due to the jobs in `nlu-gateway` are designed.

You could also read the `workspaceID` from `nlu-gateway` but this adds another project read to the inference (which might add latency). 

The way that `project.teamID` is being processed here is intentional despite the odd implementation. This is due to legacy behaviour. 

The `general-service` uses `creator-api` to pull the hashed `workspaceID` as a `string` of possibly non-numeric characters, whereas `general-runtime` uses `server-data-api` to pull the unhashed `workspaceID` as a number. The type of `project.teamID` is `string`, which is misleading, because it's actually a `number`. I've type-casted the `project.teamID` here to fix a type-issue. 

It does not help to pull `teamID` from `nlu-gateway` either because it also returns an unhashed number for the `workspaceID`. Also it adds some latency to do a redundant read when we could perform a read here.

trs (approved):

PR #478, opened 2023-02-06
[Merged by Bors] - feat: generative response node handler (CT-000)

DecathectZero:
basically replace the variables in the text, send it to GPT, and add the response to output.


lib/services/runtime/handlers/utils/output.ts
```diff
@@ -1,11 +1,26 @@
...
     project?.type === VoiceflowConstants.ProjectType.CHAT ||
     project?.platform === VoiceflowConstants.PlatformType.CHATBOT;
 
-  return isChat ? [{ text }] : text;
+  // return chat response
+  if (isChat) {
+    return output
+      .trim()
+      .split('\n')
+      .map((line) => ({ children: [{ text: line }] }));
```
DecathectZero, line 21: doing `children` preserves newlines in slate format.

lib/services/runtime/handlers/speak.ts
```diff
@@ -7,13 +7,8 @@ import { HandlerFactory } from '@/runtime';
 
 import { FrameType, Output } from '../types';
 
-// TODO: probably we can remove it, since prompt is not used in the node handler,
-// and does not exist in general service handler
-const isPromptSpeak = (node: VoiceflowNode.Speak.Node & { prompt?: unknown }) =>
-  typeof node.prompt === 'string' && node.prompt !== 'true';
-
 const SpeakHandler: HandlerFactory<VoiceflowNode.Speak.Node> = () => ({
-  canHandle: (node) => ('random_speak' in node ? !!node.random_speak : !!node.speak) || isPromptSpeak(node),
```
DecathectZero, line 16: so I had to get rid of this, the reason is mentioned in the `TODO` earlier. 
This is because the generative step also contains a `prompt` property.

jonahsnider (approved):

PR #480, opened 2023-02-10
[Merged by Bors] - feat: disable channel handlers (BUG-245)

trs:
**Fixes or implements BUG-245**

The plan is to move alexa and google to running through the general runtime through adapters. We moved over the channel specific handlers to retain functionality, but that actually breaks these projects inside the prototype tool.

Disabling these channel specific handlers for now until we can focus more on the adapters and runtime merging. Ideally we shouldn't even need these.


pmvrmc (approved):
👍

trs:
bors merge

PR #483, opened 2023-02-27
[Merged by Bors] - feat: deprecate server-data-api move to mongo (PL-000)

DecathectZero:
removing all references to `server-data-api` and directly calling the mongo database from within `general-runtime` (and also `alexa-runtime` and `google-runtime` by extension).

I've take a lot of time to to carefully look over the logic in `server-data-api` and mimic it in general-runtime.
For reference:
[projects](https://github.com/voiceflow/server-data-api/blob/fe2aaa363b4e12fa39de4d4705d1bc9897690d3a/lib/controllers/metadata.ts#L35-L40)
[project by API key](https://github.com/voiceflow/server-data-api/blob/fe2aaa363b4e12fa39de4d4705d1bc9897690d3a/lib/controllers/metadata.ts#L49-L54) - this just returns the entire project (same signature as project)
[versions](https://github.com/voiceflow/server-data-api/blob/fe2aaa363b4e12fa39de4d4705d1bc9897690d3a/lib/controllers/metadata.ts#L23-L33)
[programs](https://github.com/voiceflow/server-data-api/blob/5dc7893c4ee2ca6ffbf971a4b10e4a6fb284eff4/lib/controllers/programs.ts#L18-L21)
[prototype-programs](https://github.com/voiceflow/server-data-api/blob/081d9285823b1344e3f0bd14fe6ba2204ffe488c/lib/services/programManager.ts#L62-L69)

The important thing here is that general-runtime does not need to have a postgresSQL connection - I've run a migration to move everything over from SQL to mongo as `legacyID`: https://github.com/voiceflow/database-cli/commit/f7118e3b96a1664df8a9166665dd0dfe334df8bb


runtime/lib/DataAPI/types.ts
```diff
@@ -9,24 +9,9 @@ export interface DataAPI<
   V extends BaseModels.Version.Model<any> = BaseModels.Version.Model<BaseModels.Version.PlatformData>,
   PJ extends BaseModels.Project.Model<any, any> = BaseModels.Project.Model<AnyRecord, AnyRecord>
 > {
-  init(): Promise<void>;
-
-  fetchDisplayById(displayId: number): Promise<null | Display>;
```
DecathectZero, line 14: this is fixed through a migration, now alexa-runtime fetches it directly from the `project.platformData` object

runtime/lib/DataAPI/types.ts
```diff
@@ -9,24 +9,9 @@ export interface DataAPI<
   V extends BaseModels.Version.Model<any> = BaseModels.Version.Model<BaseModels.Version.PlatformData>,
   PJ extends BaseModels.Project.Model<any, any> = BaseModels.Project.Model<AnyRecord, AnyRecord>
 > {
-  init(): Promise<void>;
```
DecathectZero, line 12: `init` is no longer required because it was only used for `server-data-api`'s initial handshake

lib/middlewares/project.ts
```diff
@@ -74,7 +74,8 @@ class Project extends AbstractMiddleware {
         throw new VError('Error setting up data API', VError.HTTP_STATUS.UNAUTHORIZED);
       });
 
-      const project = await api.getProjectUsingAPIKey(req.headers.authorization!).catch(() => {
+      // TODO: replace with auth service in the future
+      const project = await api.getProject(req.headers.authorization!).catch(() => {
```
DecathectZero, line 78: so on `creatorDataAPI` we check if the `projectID` in `getProject` is an API key or not and fetch it with the appropriate method.

lib/services/dialog/index.ts
```diff
@@ -136,7 +136,7 @@ class DialogManagement extends AbstractManager<{ utils: typeof utils }> implemen
               model: version.prototype?.model,
               locale: version.prototype?.data.locales[0] as VoiceflowConstants.Locale,
               tag: project.liveVersion === context.versionID ? VersionTag.PRODUCTION : VersionTag.DEVELOPMENT,
-              nlp: project.prototype!.nlp,
+              nlp: project.prototype?.nlp,
```
DecathectZero, line 139: if you never train this is undefined @zhihil

lib/services/state/cacheDataAPI.ts
```diff
@@ -30,31 +28,6 @@ class CacheDataAPI implements DataAPI<VoiceflowProgram.Program, VoiceflowVersion
...
-    await this.api.init();
-  }
-
-  async unhashVersionID(versionID: string): Promise<string> {
-    return this.api.unhashVersionID(versionID);
-  }
-
-  async fetchDisplayById(displayId: number) {
-    return this.api.fetchDisplayById(displayId);
-  }
```
DecathectZero, line 44: I've run a migration and updated `alexa-runtime` to never need this.

lib/services/state/cacheDataAPI.ts
```diff
@@ -30,31 +28,6 @@ class CacheDataAPI implements DataAPI<VoiceflowProgram.Program, VoiceflowVersion
...
     return this.projects[projectID];
   }
-
-  async init() {
-    await this.api.init();
-  }
-
-  async unhashVersionID(versionID: string): Promise<string> {
-    return this.api.unhashVersionID(versionID);
-  }
```
DecathectZero, line 40: `legacyID` is always hashed, so no need to use unhashed values

lib/services/state/cacheDataAPI.ts
```diff
@@ -30,31 +28,6 @@ class CacheDataAPI implements DataAPI<VoiceflowProgram.Program, VoiceflowVersion
...
-
-  async getProjectNLP(projectID: string) {
-    if (!this.projectNLPCache.has(projectID)) {
-      this.projectNLPCache.set(projectID, await this.api.getProjectNLP(projectID));
-    }
-
-    return this.projectNLPCache.get(projectID)!;
-  }
-
-  async getProjectUsingAPIKey(apiKeyID: string): Promise<VoiceflowProject.Project> {
```
DecathectZero, line 54: replaced with this: https://github.com/voiceflow/general-runtime/blob/0f2349ffcb3566754aaaac1a032360a42a0d89fd/runtime/lib/DataAPI/creatorDataAPI.ts#L47-L52

trs (approved):
As discussed, we need to add the auth service middleware to these routes before moving production off the creator api.

PR #485, opened 2023-02-27
[Merged by Bors] - chore: remove references to workspace API key (PL-450)

astopo:
**Fixes or implements PL-450**

Update docs - removes references to Workspace API Keys, except for errors.

### Checklist

### DIFF
--- a/backend/docs/openapi.yaml
+++ b/backend/docs/openapi.yaml
@@ -16,10 +16,10 @@ info:
 
     <img src="https://uploads-ssl.webflow.com/5bddf05642686caf6d17eb58/6036dc95761e92262c2f4bd6_IVR%20(3)-p-1080.png" alt="Voiceflow IVR Integration" style="zoom:50%;"/>
 
-    
+
     **Note**: The Dialog Manager API can only be used to integrate Voiceflow **general projects**, that is, 
     projects without a pre-existing third-party upload. 
-    
+
 
     Thus, Alexa, Google Actions, and Dialogflow ES projects on Voiceflow are instead accessed through 
     their associated 3rd party platform, e.g, Alexa projects can be accessed by an Amazon Alexa device. 
@@ -172,7 +172,7 @@ tags:
 
       ### Tracking conversation state
 
-    
+
       All endpoints take in a `userID` parameter, which is used to identify the caller and assign them a unique
       conversation state object.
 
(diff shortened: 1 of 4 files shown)


backend/docs/openapi.yaml
```diff
@@ -862,41 +847,41 @@ paths:
             version alias.
 
 
-            3. `'Cannot resolve version alias from Workspace API key'` - Caller provided a Workspace or legacy Workspace API key, which cannot 
+            3. `'Cannot resolve version alias from Workspace API key'` - Caller provided a legacy Workspace API key, which cannot
```
trs, line 850: I think we could remove this outright. I'd say we treat a WS key in the authorization like any other invalid string, and not have a special error for it.

trs (approved):

backend/docs/openapi.yaml
astopo: I agree - should we remove the error in this PR as well?

trs: Yeah I think so

astopo: @trs I got carried away and removed support for legacy Workspace Keys (along with the error) - but now rethinking that - do we actually want to continue to support them, even if the docs don't say that we do?

lib/middlewares/project.ts
```diff
@@ -54,10 +54,6 @@ class Project extends AbstractMiddleware {
           throw new VError('Missing versionID in request', VError.HTTP_STATUS.BAD_REQUEST);
         }
 
-        if (isVersionTag(req.headers.versionID)) {
```
trs, line 57: I think we still need this, since if they don't give us a DM key but do give us a version tag, we can't resolve the version ID.
Can change the error to just a generic `Cannot resolve version alias` or something.

trs (changes requested):

lib/middlewares/project.ts
astopo: Okay, added back, and updated the error.

backend/docs/openapi.yaml
trs: Yeah I'm thinking we should probably still support the code for now. It'll be unsupported when we move to the new auth service.

runtime/lib/DataAPI/utils.ts
```diff
@@ -4,9 +4,5 @@ export const extractAPIKeyID = (key: unknown): string => {
   if (BaseModels.ApiKey.isWorkspaceAPIKey(key) || BaseModels.ApiKey.isDialogManagerAPIKey(key)) {
     return key.split('.')[2];
   }
-  // Handle legacy workspace API keys: "VF.<id>.<key>"
-  if (typeof key === 'string' && key.startsWith(BaseModels.ApiKey.API_KEY_PREFIX)) {
-    return key.split('.')[1];
-  }
```
trs, line 10: As discussed we should probably continue to support these. This PR should just be the docs

trs (approved):

PR #486, opened 2023-02-28
feat: track requests that are made with deprecated workspace keys (PL-446)

astopo:
**Fixes or implements PL-446**

Add metrics for requests made with deprecated Workspace API Keys.

### Checklist

### DIFF
--- a/lib/services/interact/index.ts
+++ b/lib/services/interact/index.ts
@@ -64,6 +64,9 @@ class Interact extends AbstractManager<{ utils: typeof utils }> {
 
     metrics.generalRequest();
     if (authorization?.startsWith('VF.')) metrics.sdkRequest();
+    // If we are using a non-DM key, assume this is a deprecated Workspace API Key.
+    if (authorization?.startsWith('VF.') && !authorization?.startsWith('VF.DM'))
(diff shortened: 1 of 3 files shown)


lib/services/interact/index.ts
```diff
@@ -64,6 +64,9 @@ class Interact extends AbstractManager<{ utils: typeof utils }> {
 
     metrics.generalRequest();
     if (authorization?.startsWith('VF.')) metrics.sdkRequest();
+    // If we are using a non-DM key, assume this is a deprecated Workspace API Key.
+    if (authorization?.startsWith('VF.') && !authorization?.startsWith('VF.DM'))
+      metrics.deprecatedAPIKey(authorization);
```
trs, line 69: We'd need more info, at least version/project. Otherwise we'd need to correlate metrics data with the database to know who's key this is.

astopo: Good call, was thinking to grab them then run a script, but will be easier if we have the version or project as well. Will update.

astopo:
Moving this to `creator-api` - see PR: https://github.com/voiceflow/creator-api/pull/1194

PR #488, opened 2023-03-02
[Merged by Bors] - fix: generate step last response [bugfix] (BUG-291)

DecathectZero:
the generative step should save its output to `last_response`


lib/services/runtime/handlers/generative.ts
```diff
@@ -46,6 +47,7 @@ const GenerativeHandler: HandlerFactory<VoiceNode.Generative.Node> = () => ({
     );
 
     runtime.stack.top().storage.set<Output>(FrameType.OUTPUT, output);
+    variables.set(VoiceflowConstants.BuiltInVariable.LAST_RESPONSE, output);
```
DecathectZero, line 50: this is the only actual code change

jonahsnider (approved):

PR #495, opened 2023-03-16
[Merged by Bors] - fix(runtime): set as running state if no request (BUG-304)

trs:
**Fixes or implements BUG-304**

When we start from a block in canvas we send a null request with a stack containing the node we want to start from.
The runtime would automatically start in a `REQUEST` state, which causes the capture step to think the request is the user's response to the capture step, causing a no match to trigger.
Instead, if we don't get a request from the user (`null`), then start the runtime in a `RUNNING` state instead. This tells the capture block that we're entering the block and to wait for the user reply instead.


pmvrmc (commented):
looks good, I like the logic of moving this up the chain.

However we should also take the same approach for carousel an cardV2 no?

On those handlers we do the exact same `runtime.getAction() === Action.REQUEST && !runtime.getRequest();` compare logic

PR #497, opened 2023-03-20
fix: capture matching on invalid slot value (PL-313)

zhihil:
**Fixes or implements PL-313**

Changed conditional check of Capture V2 to be more specific. Rather than declaring a successful match on any intent, we check that the intent given to the Capture V2 matches the intent configured for the Capture V2 OR Capture V2 is capturing in the entire response (in which case checking the intent names is unnecessary).

Added a more specific conditional check on Capture V2.

NOTE: The Capture V2 step will now fail on prototyping if it is not trained. Previously, the conditional check was so loose that Capture V2 happened to work on Prototyping even though regex intent classification failed and produced a None intent for a correct utterance. 

Now that the check is more strict, that None intent will be no-matched by the Capture V2 instead of being accepted as a successful match. If we want to fix this and allow limited prototyping even if the NLU model is not trained, we should modify the regex intent classification to produce the Capture step's intent whenever an exact slot value is given. 

### Setup information

N/A

### Deployment Notes

N/A

### Related PRs

N/A

### Checklist

### DIFF
--- a/lib/services/runtime/handlers/captureV2/captureV2.ts
+++ b/lib/services/runtime/handlers/captureV2/captureV2.ts
@@ -3,7 +3,7 @@ import { VoiceflowNode } from '@voiceflow/voiceflow-types';
 
 import { Action, HandlerFactory } from '@/runtime';
 
-import { isIntentRequest, StorageType } from '../../types';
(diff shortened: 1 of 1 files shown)


trs (approved):
Tested a bunch, this just makes sense 👍

PR #498, opened 2023-03-20
[Merged by Bors] - fix: make `response.` prefix for extracting variables from API step responses optional [bugfix] (PL-000)

jonahsnider:

rewrites the function for extracting variables from API step responses.
it makes the function much more readable/maintainable, and also makes it optional to prefix the path with `response.` without breaking reverse compatibility

### Checklist


DecathectZero (approved):

trs (commented):
Can you add a few test cases?
https://github.com/voiceflow/general-runtime/blob/master/tests/runtime/lib/Handlers/api/utils.unit.ts

trs (approved):

PR #499, opened 2023-03-21
[Merged by Bors] - fix(captureV2): check for variable on intent (BUG-313) (BUG-294)

trs:
**Fixes or implements BUG-313, BUG-294**

We only want to `handleCapturePath` if we match the intent we're looking for or it's a full reply capture. The previous fix (https://github.com/voiceflow/general-runtime/commit/b76f63d2a5728611b7f50a534482117a23729130) caused the runtime to think a capture was successful even if it didn't match the intent specified. This change undoes that but also adds an additional `!node.variable` check to the `isNodeCapturingEntity`. This is specifically for Alexa as it's entire user reply has an intent, entity, and variable.

Replaces https://github.com/voiceflow/general-runtime/pull/497 to keep fixed BUG-294 as well.


DecathectZero (approved):

trs:
bors merge

trs:
bors merge

PR #500, opened 2023-03-22
[Merged by Bors] - fix: consolidate all text traces add delay (PL-000)

DecathectZero:
we had this weird pattern of injecting `addTrace` into the `outputTrace` function.

Instead the `outputTrace` fn is now an immutable function called `getOutputTrace` that just returns the trace.
I've added a new function called `addOutputTrace` that emulates the API of the `runtime` so we can add debug traces AFTER the actual text/speak trace.

Added a new `version` param so that we can extract the `messageDelay` settings. The important thing is that this is standardized so ALL places that have an output will do all the peripheral stuff: adding debug traces, saving a message to `{last_response}` etc.

we now use the `delay` field: https://github.com/voiceflow/libs/pull/427

### DIFF
--- a/lib/services/runtime/utils.ts
+++ b/lib/services/runtime/utils.ts
@@ -233,12 +233,16 @@ export function textOutputTrace({
   const richContent = {
     id: cuid.slug(),
     content,
-    messageDelayMilliseconds: delay ?? getVersionMessageDelay(version) ?? DEFAULT_DELAY,
   };
 
   const trace: BaseTrace.Text = {
     type: BaseNode.Utils.TraceType.TEXT,
-    payload: { slate: richContent, message: plainContent, ...(ai && { ai }) },
+    payload: {
+      slate: richContent,
+      message: plainContent,
+      delay: delay ?? getVersionMessageDelay(version) ?? DEFAULT_DELAY,
+      ...(ai && { ai }),
+    },
   };
 
   variables?.set(VoiceflowConstants.BuiltInVariable.LAST_RESPONSE, plainContent);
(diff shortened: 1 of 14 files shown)


lib/services/runtime/utils.ts
```diff
@@ -198,42 +207,88 @@ export const removeEmptyPrompts = (
...
+> => {
+  return !!version?.platformData?.settings?.messageDelay?.durationMilliseconds;
+};
+
+const getVersionMessageDelay = (version?: BaseModels.Version.Model<BaseModels.Version.PlatformData>) => {
+  if (version && hasMessageDelay(version)) {
+    return version.platformData.settings.messageDelay.durationMilliseconds;
+  }
+  return null;
+};
```
trs, line 225: I love the type guard, but this really could just return the optional chaining.

```ts
return version?.platformData?.settings?.messageDelay?.durationMilliseconds ?? nul
```

trs (approved):
Good improvement

lib/services/runtime/utils.ts
```diff
@@ -198,41 +207,90 @@ export const removeEmptyPrompts = (
...
+  const plainContent = slateToPlaintext(content);
+  const richContent = {
+    id: cuid.slug(),
+    content,
+  };
+
+  const trace: BaseTrace.Text = {
+    type: BaseNode.Utils.TraceType.TEXT,
+    payload: {
+      slate: richContent,
```
trs, line 241: Do we want to keep the `durationMilliseconds` on `slate` for compatibility?

trs (approved):

PR #503, opened 2023-04-06
[Merged by Bors] - fix: interaction confidence threshold (BUG-351) [bugfix]

trs:
**Fixes or implements [BUG-351](https://voiceflow.atlassian.net/browse/BUG-351)**

To avoid matching interactions with low confidence, we add a confidence gate similar to the capture step.
Currently set to 60% (same as capture).

Turning this:
image

Into this:
image

[BUG-351]: https://voiceflow.atlassian.net/browse/BUG-351?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/services/runtime/utils.ts
+++ b/lib/services/runtime/utils.ts
@@ -287,3 +287,6 @@ export const processOutput = (output: string | Text.SlateTextValue | undefined, 
   const content = slateInjectVariables(output, sanitizedVars);
   return slateToPlaintext(content);
 };
+
+export const isConfidenceScoreAbove = (threshold: number, confidence: number) =>
+  typeof confidence !== 'number' || confidence > threshold;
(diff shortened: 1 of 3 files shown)


lib/services/runtime/utils.ts
```diff
@@ -287,3 +287,6 @@ export const processOutput = (output: string | Text.SlateTextValue | undefined,
   const content = slateInjectVariables(output, sanitizedVars);
   return slateToPlaintext(content);
 };
+
+export const isConfidenceScoreAbove = (threshold: number, confidence: number) =>
+  typeof confidence !== 'number' || confidence > threshold;
```
zhihil, line 292: Do we know in which situations `request.payload?.confidence` is `undefined`? 

I'm just wondering if its best practice to instead use `type confidence === 'number' && confidence > threshold` here, rather than assuming if confidence is `undefined` then it must be 100.

pmvrmc (approved):

PR #506, opened 2023-04-13
[Merged by Bors] - fix: send credentials to runtime test routes (PL-000)

DecathectZero:
added `@voiceflow/sdk-auth` but had to add `@voiceflow/schema-types` and `nestjs-zod` because they were dependencies. 

Added a middleware layer that just checks to make sure the user is logged into voiceflow and has a valid session.


lib/clients/index.ts
```diff
@@ -30,6 +33,9 @@ const buildClients = (config: Config): ClientMap => {
     ...Static,
     redis,
     mongo,
+    auth: config.AUTH_API_ENDPOINT
+      ? new AuthSDK({ authServiceURI: config.AUTH_API_ENDPOINT, fetchPonyfill: fetch })
+      : null,
```
trs, line 38: With new env vars:

```suggestion
    auth: config.AUTH_API_SERVICE_HOST
      ? new AuthSDK({
        authServiceURI: new URL(
          `${config.NODE_ENV === 'e2e' ? 'https' : 'http'}://${config.AUTH_API_SERVICE_HOST}:${
            config.AUTH_API_SERVICE_PORT_APP
          }`
        ).href,,
        fetchPonyfill: fetch
      })
      : null,
```

lib/middlewares/auth.ts
```diff
@@ -0,0 +1,27 @@
+import { formatAuthorizationToken } from '@voiceflow/sdk-auth/express';
+import { NextFunction, Response } from 'express';
+
+import { Request } from '@/types';
+
+import { AbstractMiddleware } from './utils';
+
+class Auth extends AbstractMiddleware {
+  async verify(req: Request, res: Response, next: NextFunction): Promise<void> {
```
trs, line 9: We can add an identity validation middleware to the new auth sdk too, so you won't have to rebuild this in any service that needs this.

types.ts
```diff
@@ -43,6 +43,7 @@ export interface Config extends RateLimitConfig {
   BUILD_URL: string | null;
 
   GENERAL_SERVICE_ENDPOINT: string | null;
+  AUTH_API_ENDPOINT: string | null;
```
trs, line 46:
```suggestion
  AUTH_API_SERVICE_HOST: string | null;
  AUTH_API_SERVICE_PORT_APP: string | null;
```

trs (approved):

lib/middlewares/auth.ts
DecathectZero: yeah happy to swap this over once that's in place - but if we switch to something auto-gened it might be more difficult I presume?

config.ts
```diff
@@ -43,6 +43,7 @@ const CONFIG: Config = {
   PROJECT_SOURCE: getOptionalProcessEnv('PROJECT_SOURCE'),
 
   GENERAL_SERVICE_ENDPOINT: getOptionalProcessEnv('GENERAL_SERVICE_ENDPOINT'), // voiceflow nlu/tts services
+  AUTH_API_ENDPOINT: getOptionalProcessEnv('AUTH_API_ENDPOINT'),
```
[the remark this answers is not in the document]
DecathectZero: It totally makes sense, it's just that ever other service env var looks like this:
```
  CREATOR_API_ENDPOINT: string;
  BILLING_API_ENDPOINT: string;
  ANALYTICS_API_ENDPOINT: string;
  IDENTITY_API_ENDPOINT: string;
```

it would be a bit weird to deviate or make an exception in this case

lib/clients/index.ts
DecathectZero: in the `.env.test` and `.env.e2e` this case should be covered:
https://github.com/voiceflow/creator-app/blob/7d10d169bc3d168fa8c8f008d6eab5b373616c90/packages/ml-gateway/.env.e2e#L12-L17

lib/middlewares/auth.ts
trs: Nope, we build a middleware/nest layer manually on top of the generated SDK

PR #510, opened 2023-04-19
[Merged by Bors] - feat: code node paths [bugfix] (CT-000)

DecathectZero:


runtime/lib/Handlers/code/index.ts
```diff
@@ -17,6 +18,8 @@ export interface CodeOptions {
 
 export const GENERATED_CODE_NODE_ID = 'PROGRAMMATICALLY-GENERATED-CODE-NODE';
 
+const RESOLVED_PATH = '__RESOLVED_PATH__';
```
matheuspoleza, line 21: is there any reason for those underscores prefix and sufix?

DecathectZero: because it's in the same variable space as variables that the user names.

if we just called it `resolved_path` but the user's project happened to have a variable called that, it would fuck it up.

PR #512, opened 2023-04-29
feat: reset session stack production [bugfix] (PL-000)

DecathectZero:
one major issue we have right now is that existing sessions will never receive new upgrades.

Imagine a project like this, where there is no exit condition:
Screenshot 2023-04-28 at 5 43 37 PM

This means that the user's session's stack will ALWAYS be on that version, because it always references the `programID` from that particular version.

What I've done is take advantage of the fact that we have a base trace frame (bottom of the stack) that is the versionID. What we do now is check if the base frame is in sync with whatever version "production" is, if not we reset the session.

This might be a little jarring because it could mean that the user's session resets in the middle of a conversation - but this is better than people getting really frustrated when they test the DMAPI because they never see their new version.

### DIFF
--- a/lib/services/state/index.ts
+++ b/lib/services/state/index.ts
@@ -1,6 +1,7 @@
 import { BaseModels, BaseRequest, BaseTrace } from '@voiceflow/base-types';
 import _ from 'lodash';
 
+import { FrameType } from '@/lib/services/runtime/types';
 import { PartialContext, State } from '@/runtime';
 import { Context, InitContextHandler } from '@/types';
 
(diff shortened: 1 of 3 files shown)


lib/services/state/index.ts
```diff
@@ -83,6 +84,14 @@ class StateManager extends AbstractManager<{ utils: typeof utils }> implements I
       context.request.payload.entities = [];
     }
 
+    // TODO: this is a hacky way to reset a session's stack after a version upgrade
+    // reset the stack for user if base frameID is not the same as the current version, otherwise they will never update
+    // this is for when the version is labelled as 'production' but can refer to an arbitrary versionID
+    const baseFrame = context.state?.stack?.[0];
+    if (baseFrame?.storage?.[FrameType.IS_BASE] && baseFrame.programID !== context.versionID) {
```
trs, line 91: Could we use the [`null` `nodeID` on that stack](https://github.com/voiceflow/general-runtime/pull/512/files#diff-aee751a6d5f06dfd059c22e66ade0dfc7dc6d56eb89353c51948ac9318433993R196) instead of a new storage item?

trs (approved):

lib/services/state/index.ts
```diff
@@ -83,6 +84,14 @@ class StateManager extends AbstractManager<{ utils: typeof utils }> implements I
       context.request.payload.entities = [];
     }
 
+    // TODO: this is a hacky way to reset a session's stack after a version upgrade
```
trs, line 87: What's the TODO? To find a less hacky way? Otherwise this is more of a NOTE (🚲🏠)

DecathectZero: I think we need some kind of "canonicalID" on programs so we can move people over between the same program but on different versions. 
Granted that it still introduces a large category of errors if the user is on a step that doesn't exist and the functionality of the program means it has no start step.

lib/services/state/index.ts
DecathectZero: `null` is not a consistent reference to guarantee that the frame is a "base" frame.
There "should" always be a base frame but I'm worried there's some legacy stuff, prototype tool stuff where it actually sends in via stateless API the stack but only with the diagram they want to start on.

PR #518, opened 2023-05-19
[Merged by Bors] - feat: static and generative fallback (LLM-001)

DecathectZero:


lib/services/runtime/utils.ts
```diff
@@ -189,7 +193,7 @@ export const slateToPlaintext = (content: Readonly<BaseText.SlateTextValue> = []
     .join('\n')
     .trim();
 
-export const isPromptContentInitialyzed = (content: BaseText.SlateTextValue | string | null | undefined) =>
+export const isPromptContentInitialized = (content: BaseText.SlateTextValue | string | null | undefined) =>
```
viniciusdacal, line 196: this is being used by the other runtimes.
We just need to make sure we update them after we merge this.

lib/services/runtime/utils.ts
```diff
@@ -168,13 +168,17 @@ export const addButtonsIfExists = <N extends BaseRequest.NodeButton>(
...
+};
+
 export const getGenericGlobalNoMatchPrompt =
   ({ isPrompt }: { isPrompt: (prompt: unknown) => prompt is Prompt }) =>
   (runtime: Runtime) => {
-    const { version } = runtime;
-    const prompt = version?.platformData.settings?.globalNoMatch?.prompt;
-    return prompt && isPrompt(prompt) ? prompt : null;
+    const noMatch = getGlobalNoMatch(runtime);
+    return noMatch?.prompt && isPrompt(noMatch?.prompt) ? noMatch?.prompt : null;
```
viniciusdacal, line 179: Oh, I missed that in the past, but the nullish coalescing is not needed after we check the presence of noMatch.

```suggestion
    return noMatch?.prompt && isPrompt(noMatch.prompt) ? noMatch.prompt : null;
```

viniciusdacal (approved):

PR #523, opened 2023-05-29
[Merged by Bors] - fix: llm rate limit fallback (LLM-002)

DecathectZero:
So from my testing, we may get throttled and rate limited on azure at around 20-30 concurrent requests. The 427 error that we get back is nearly instantaneous - this gives us an opportunity to defer the call to the openAI API as a fallback.

Frank is applying for a rate limit increase, but from what we've heard Microsoft is insanely slow on this stuff.

For the time being, I think we only need to set this up on 3.5 - its our default model across the entire tool

### DIFF
--- a/lib/clients/ai/gpt/gpt3_5.ts
+++ b/lib/clients/ai/gpt/gpt3_5.ts
@@ -38,7 +38,7 @@ export class GPT3_5 extends GPTAIModel {
       log.warn(`GPT3.5 completion ${log.vars({ error: error?.response ?? error, messages, params })})}`);
 
       // if we fail on the azure instance due to rate limiting, retry with OpenAI API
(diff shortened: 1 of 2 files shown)


lib/clients/ai/gpt/gpt3_5.ts
```diff
@@ -17,22 +17,32 @@ export class GPT3_5 extends GPTAIModel {
     return this.generateChatCompletion(messages, params);
   }
 
-  async generateChatCompletion(messages: Message[], params: AIModelParams) {
-    const result = await this.client
-      .createChatCompletion(
+  async generateChatCompletion(
+    messages: Message[],
+    params: AIModelParams,
+    client = this.client
```
DecathectZero, line 23: third parameter is the client to use

zhihil (approved):

PR #525, opened 2023-05-31
feat: add anthropic llm models (LLM-004)

DecathectZero:
switching from `openai` `ChatCompletionRequestMessageRoleEnum` to `BaseUtils.ai.Role`, so that it is more generic and shared across all LLM models, not just OpenAI's.

Added SDK client for anthropic, it is now a model that can be leveraged.


lib/clients/ai/anthropic/claude_v1.ts
```diff
@@ -0,0 +1,9 @@
+import { BaseUtils } from '@voiceflow/base-types';
+
+import { AnthropicAIModel } from './utils';
+
+export class ClaudeV1 extends AnthropicAIModel {
+  public modelRef = BaseUtils.ai.GPT_MODEL.CLAUDE_V1;
+
+  protected anthropicModel = 'claude-v1';
```
zhihil, line 8: This is a pretty good abstraction.

The `public`, `protected` quantifiers don't appear to be necessary here. The following seems to build properly:

```ts
export class ClaudeV1 extends AnthropicAIModel {
  modelRef = BaseUtils.ai.GPT_MODEL.CLAUDE_V1;
  anthropicModel = 'claude-v1';
```

One minor point: it may be preferable to use a chain of `super` constructor calls rather than setting the fields in subclasses, since that provides a clearer interface which shows how subclasses should use the parent class. 

Specifically, the constructors could be reworked as follows:

```ts
export abstract class AIModel {
  protected readonly modelRef: BaseUtils.ai.GPT_MODEL;

  constructor(modelRef: BaseUtils.ai.GPT_MODEL) {
    this.modelRef = modelRef;
  }
}

export abstract class AnthropicAIModel extends AIModel {
  protected TIMEOUT = 20000;

  protected client: Client;

  private readonly anthropicModel: string;

  constructor(anthropicModel: string, modelRef: BaseUtils.ai.GPT_MODEL, config: Partial<Config>) {
    super(modelRef);

    this.anthropicModel = anthropicModel;

    if (!config.ANTHROPIC_API_KEY) {
      throw new Error(`Anthropic client not initialized`);
    }

    this.client = new Client(config.ANTHROPIC_API_KEY);
  }
}

export class ClaudeV1 extends AnthropicAIModel {
  constructor(config: Partial<Config>) {
    super(BaseUtils.ai.GPT_MODEL.CLAUDE_V1, 'claude-v1', config)
  }
}
```

Not critical here, but it's definitely clearer if the parent class interface gets more complex, e.g. some fields have default values set by the parent, so the subclass may not need to set the field in its implementation. In this situation, the developer can't tell from a glance what they are required to pass and what is just optional. They'd have to understand read all the subclasses as examples and the implementation to figure this out.

lib/clients/ai/anthropic/utils.ts
```diff
@@ -0,0 +1,66 @@
...
+import { BaseUtils } from '@voiceflow/base-types';
+import { AIModelParams } from '@voiceflow/base-types/build/cjs/utils/ai';
+
+import log from '@/logger';
+import { Config } from '@/types';
+
+import { AIModel } from '../types';
+
+export abstract class AnthropicAIModel extends AIModel {
+  protected TIMEOUT = 20000;
```
zhihil, line 12: Protected isn't necessary here unless want to set different timeouts for different subclasses for whatever reason. Using `private` access is preferable to avoid polluting subclasses with internal parent class data.

Also, setting `TIMEOUT` here in `AnthropicAIModel` is redundant because `TIMEOUT` set is already in `AIModel`

lib/clients/ai/anthropic/utils.ts
```diff
@@ -0,0 +1,66 @@
...
+    const prompt = `${topSystem}\n\n${messages.map(
+      (message) => `${AnthropicAIModel.RoleMap[message.role]} ${message.content}`
+    )}${AI_PROMPT}`;
+
+    const result = await this.client
+      .complete({
+        prompt,
+        model: this.anthropicModel,
+        temperature: params.temperature,
+        max_tokens_to_sample: params.maxTokens || 128,
```
zhihil, line 56: This `128` should probably be a class field.

lib/clients/ai/openai/utils.ts
```diff
@@ -7,10 +8,18 @@ import { AIModel } from '../types';
 export abstract class GPTAIModel extends AIModel {
   protected TIMEOUT = 20000;
```
zhihil, line 9: This `protected TIMEOUT = 20000` is redundant as well. The parent already sets this to `20000` I missed it in my earlier PR review for the GPT models.

zhihil (approved):
Seems good, mainly have some notes about setting parent fields in subclasses.

lib/clients/ai/anthropic/claude_v1.ts
DecathectZero: I think it's fine for now - if the instance variable is an abstracted one, the child classes have to implement it or else get an error. `modelRef` is not defined for them

lib/clients/ai/anthropic/utils.ts
DecathectZero: sure

lib/clients/ai/anthropic/claude_v1.ts
DecathectZero: I think it's fine for now - if the instance variable is an abstracted one, the child classes have to implement it or else get an error. `modelRef` is not defined for them

Xavientois:
bors try

PR #526, opened 2023-05-31
feat: implement new public feedback endpoint (CT-2052)

viniciusdacal:
**Fixes or implements CT-2052**

Add new public endpoint to accept feedback from react-chat

### Related PRs

### Checklist


lib/services/feedback.ts
```diff
@@ -0,0 +1,54 @@
...
+    text: string;
+    projectID: string;
+    userID: string;
+    versionID: string;
+    name: string;
+  }) => {
+    if (!this.analyticsServiceURI) return;
+
+    await this.services.axios
+      .post(
```
pmvrmc, line 33: we should use the analytics-sdk no?

lib/controllers/public/index.ts
```diff
@@ -44,6 +45,28 @@ class PublicController extends AbstractController {
...
+      { userID: string },
+      { text: string; type: 'positive' | 'negative' },
+      { projectID: string; versionID: string }
+    >
+  ) {
+    await this.services.feedback.track({
+      projectID: req.headers.projectID,
+      versionID: req.headers.versionID,
+      text: req.body.text,
+      name: req.body.type === 'positive' ? 'Thumbs up' : 'Thumbs down',
```
pmvrmc, line 64: these should be a constant no?

lib/controllers/public/index.ts
```diff
@@ -44,6 +45,28 @@ class PublicController extends AbstractController {
     return { trace };
   }
 
+  @validate({
+    HEADERS_PROJECT_ID: VALIDATIONS.HEADERS.PROJECT_ID,
+    HEADERS_VERSION_ID: VALIDATIONS.HEADERS.VERSION_ID,
+    BODY_PUBLIC_FEEDBACK: VALIDATIONS.BODY.PUBLIC_FEEDBACK,
+  })
+  async feedback(
```
pmvrmc, line 53: why does this live on the runtime and not creator-api?

pmvrmc (changes requested):
also the new env variables dont seem set according to the checklist, if they are not set we are potentially breaking the server no?

lib/controllers/public/index.ts
viniciusdacal: this will be used by the `react-chat` only. As the react-chat only interacts with the runtime, made sense to include this call to the runtime.

It's mostly `react-chat` -> `runtime-sdk` -> `runtime` -> `analytics`.

Also, we didn't want to expose analytic call publicly, because it could be tracked by adblockers.

lib/services/feedback.ts
viniciusdacal: `analytics-sdk` is a private package. 
We removed all private packages from general-runtime last week, because some clients complained about not being able to use it properly.

lib/controllers/public/index.ts
viniciusdacal: yeah, I can add those to a constant

viniciusdacal:
> also the new env variables don't seem set according to the checklist, if they are not set we are potentially breaking the server no?

I had a chat with infra, those are actually set globally to all projects from what I understood. 

But basically, they were already configured. (I had a slightly different name before and the build was breaking, following the name build is passing and we are able to test it locally without issues)

lib/controllers/public/index.ts
pmvrmc: if you have given some thought into this then I guess its ok.

but definitively feels wrong to add new variables and have the runtime acting as a proxy for the analytics service.

not sure about the adblockers thing, isnt this call potentially being tracked? since it is just like a proxy then wont it be the same thing?

pmvrmc (approved):
tbh Im very conflicted on having this live on the runtime, if we need hacks for adblockers/rate-limiting/etc I dont think runtime should be the place for it.

however I dont want to block deployment of this.

I'll approve but please give it another thought before merging

JoshTheDeveloper:
bors cancel

JoshTheDeveloper:
we'll chat about this tmrw, we can wait a day or two to approve/deploy

backend/api/routers/feedback.ts
```diff
@@ -0,0 +1,18 @@
...
+import { ControllerMap, MiddlewareMap } from '@/lib';
+
+export default (middlewares: MiddlewareMap, controllers: ControllerMap) => {
+  const router = express.Router();
+
+  router.use(bodyParser.json({ limit: BODY_PARSER_SIZE_LIMIT }));
+
+  const feedbackMiddleware = [middlewares.project.resolvePublicProjectID, middlewares.rateLimit.versionConsume];
+
+  // full route: /feedback/:projectID/user/:userID/interact
```
pmvrmc, line 14: typo on the comment here

also since we have feedback at the begining of the route, do we need the last /feedback again?

trs: Yeah, the route currently would be `feedback/:projectID/user/:userID/feedback`

viniciusdacal: Whoops, my bad. I'm removing the final `/feedback` too, great catch

lib/controllers/feedback/index.ts
```diff
@@ -0,0 +1,47 @@
...
+        [key: string]: any;
+      },
+      { projectID: string; versionID: string }
+    >
+  ) {
+    await this.services.feedback.track({
+      projectID: req.headers.projectID,
+      versionID: req.headers.versionID,
+      userID: req.params.userID,
+      ...req.body,
```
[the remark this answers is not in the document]
viniciusdacal: Yeah, tried to make it less aware as possible and possibly future proof

PR #532, opened 2023-06-19
feat: billing and authorization for test endpoints (PL-547)

trs:

Check and consume token quotas in `/test` endpoints. Creates new endpoints with a `workspaceID` param so we can transition the creator app. Once transitioned, the old routes will be removed.

Adds optional auth and billing middleware which will be skipped if the packages are not installed.


lib/middlewares/billing.ts
```diff
@@ -0,0 +1,26 @@
...
+  @factory()
+  checkQuota(quotaName: string, getWorkspaceID: (req: Request) => string | undefined) {
+    return async (req: Request, _res: Response, next: NextFunction) => {
+      try {
+        const workspaceID = getWorkspaceID(req);
+        if (!workspaceID) {
+          return next(new VError('Unauthorized', 401));
+        }
+
+        await this.services.billing.consumeQuota(workspaceID, quotaName, 0);
```
zhihil, line 19: Is this `consumeQuota(..., ..., 0)` intentional? How come we set `count` to `0` here?

lib/middlewares/auth.ts
```diff
@@ -1,77 +1,77 @@
...
-        })
-        .then((res) => res.data);
+  async verifyIdentity(req: Request, _res: Response, next: Next): Promise<void> {
+    const client = await this.getClient();
+    if (!client) return next();
+    const authorization = req.headers.authorization || req.cookies.auth_vf || '';
 
-      if (!identity?.identity?.id) throw new Error();
-    } catch (error) {
-      res.sendStatus(VError.HTTP_STATUS.UNAUTHORIZED);
```
zhihil, line 61: Should we not catch the error and ensure that we send back `HTTP_STATUS_UNAUTHORIZED` to be clearer to the end user?

lib/middlewares/billing.ts
trs: It's a bit odd of a pattern, it's using `0` to check that the quota hasn't been exceeded yet. Then after the tokens are used we consume the quota used.
So since we need to guard, it uses this pattern with `0`.

zhihil: I see, so the `count` is the amount of times that the quota is being deducted. So we improvise a check by deducting 0 uses from the quota and then awaiting the result.

zhihil (approved):

trs:
bors merge

DecathectZero (approved):

PR #533, opened 2023-06-20
fix: add alexa no reply on interaction handler (BUG-453)

pmvrmc:
**Fixes or implements BUG-453**

On interaction calls the no reply prompt was not being sent to the adapter/alexa.

Very similar to what we did on the captureV2 handler, there is a tiny specific logic on the interaction handler to account for this.


zhihil (approved):
The code looks good, however, I'm not able to test because the bug seem to be only reproducible on Alexa and I didn't bring my Alexa device to the office.

trs (approved):

PR #535, opened 2023-06-27
fix: remove Alexa from voice (BUG-456)

trs:

`"Alexa"` is not a valid voice option for Alexa projects. Remove it in that case.


pmvrmc (approved):

trs:
bors merge

PR #541, opened 2023-06-29
feat: check and consume tokens in runtime (PL-000) [bugfix]

trs:

Check and consume the Open AI token quota in runtime handlers.

### DIFF
--- a/lib/services/runtime/handlers/aiResponse.ts
+++ b/lib/services/runtime/handlers/aiResponse.ts
@@ -1,4 +1,5 @@
 import { BaseNode, BaseUtils } from '@voiceflow/base-types';
+import VError from '@voiceflow/verror';
 import { VoiceNode } from '@voiceflow/voice-types';
 
 import { HandlerFactory } from '@/runtime';
@@ -14,6 +15,11 @@ const AIResponseHandler: HandlerFactory<VoiceNode.AIResponse.Node> = () => ({
   canHandle: (node) => node.type === BaseNode.NodeType.AI_RESPONSE,
   handle: async (node, runtime, variables) => {
     const nextID = node.nextId ?? null;
+    const workspaceID = runtime.project?.teamID;
+
+    if (!(await runtime.services.billing.checkQuota(workspaceID, 'OpenAI Tokens'))) {
+      throw new VError('Quota exceeded', VError.HTTP_STATUS.PAYMENT_REQUIRED);
+    }
 
     if (node.source === BaseUtils.ai.DATA_SOURCE.KNOWLEDGE_BASE) {
       const { prompt, mode } = node;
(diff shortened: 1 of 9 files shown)


zhihil:
This change appears to have broken the AI Response step somehow. Once we hit the AI Response step, the Prototype tool just terminates.

Screen Shot 2023-06-30 at 12 20 09 PM

zhihil:
IIRC we also use the knowledge base in the no-match here, to generate a fallback message.

Do we need to guard this quota as well? 

Screen Shot 2023-06-30 at 12 32 46 PM

lib/services/runtime/handlers/aiResponse.ts
```diff
@@ -57,8 +58,18 @@ const AIResponseHandler: HandlerFactory<VoiceNode.AIResponse.Node> = () => ({
       return nextID;
     }
 
+    const workspaceID = runtime.project?.teamID;
+
+    if (!(await runtime.services.billing.checkQuota(workspaceID, 'OpenAI Tokens'))) {
+      throw new VError('Quota exceeded', VError.HTTP_STATUS.PAYMENT_REQUIRED);
+    }
+
     const response = await fetchPrompt(node, variables.getState());
```
zhihil, line 67: We need to check that the model being used is actually from OpenAI, otherwise we'll unfairly charge tokens on the other models. 

In the implementation of `fetchPrompt` below, you can see we're polymorphically retrieving a model using `AI.get(node.model)`

Screen Shot 2023-06-30 at 12 38 34 PM

One of the models is the Claude model, which AFAIK isn't accessed through OpenAI. It belongs to a company called Anthropic. 

We should change the `if` condition to something like this:

```ts
function usesOpenAI(node) {
   return [`GPT_MODEL.DaVinci_003`, `GPT_MODEL.GPT_3_5_turbo`, `GPT_MODEL.GPT_4`].includes(node.model);
}

if (usesOpenAI(node) && !(await runtime.services.billing.checkQuota(workspaceID, 'OpenAI Tokens'))) {
      throw new VError('Quota exceeded', VError.HTTP_STATUS.PAYMENT_REQUIRED);
    }
```

trs: For now we won't worry about this

trs: Token limit applies to all models.

image

zhihil (dismissed):

zhihil (approved):
The quota check appears to be working. Only comment is a minor UX issue. When the quota check causes generative AI responses to fail, the error message is not being properly propagated to the Interact endpoint

This leads to interaction silently failing without a clear indication of why. 

Screen Shot 2023-06-30 at 3 48 55 PM

Screen Shot 2023-06-30 at 3 49 23 PM

trs:
I'll see if we can support these sorts of errors as a fast follow

trs:
bors merge

PR #545, opened 2023-07-07
fix: token limit error message [bugfix] (LLM-000)

DecathectZero:
Right now if you hit a token limit the runtime will just fail silently, the `VError` gets swallowed up by the handler:https://github.com/voiceflow/general-runtime/blob/39ceeb3629fe2867f10940793c4f922a51a82f83/runtime/lib/Runtime/cycleHandler.ts#L33-L45

Also the frontend has no way to visualize this error at the moment, so for the short term, I think it is better to just explicitly surface it as a text or something.

Right now I'm getting a large influx of "bug" reports from users reporting their LLM features aren't working.

Screenshot 2023-07-07 at 3 06 04 PM

### DIFF
--- a/lib/services/runtime/handlers/aiResponse.ts
+++ b/lib/services/runtime/handlers/aiResponse.ts
@@ -1,8 +1,8 @@
 import { BaseNode, BaseUtils } from '@voiceflow/base-types';
-import VError from '@voiceflow/verror';
 import { VoiceNode } from '@voiceflow/voice-types';
 
 import { GPT4_ABLE_PLAN } from '@/lib/clients/ai/types';
+import { QuotaName } from '@/lib/services/billing';
 import { HandlerFactory } from '@/runtime';
 
 import { FrameType, Output } from '../types';
@@ -19,8 +19,17 @@ const AIResponseHandler: HandlerFactory<VoiceNode.AIResponse.Node> = () => ({
     const nextID = node.nextId ?? null;
     const workspaceID = runtime.project?.teamID;
 
-    if (!(await runtime.services.billing.checkQuota(workspaceID, 'OpenAI Tokens'))) {
-      throw new VError('Quota exceeded', VError.HTTP_STATUS.PAYMENT_REQUIRED);
+    if (!(await runtime.services.billing.checkQuota(workspaceID, QuotaName.OPEN_API_TOKENS))) {
+      addOutputTrace(
+        runtime,
+        getOutputTrace({
+          output: generateOutput('[token quota exceeded]', runtime.project),
+          version: runtime.version,
+          ai: true,
+        })
+      );
+      runtime.trace.debug('token quota exceeded', BaseNode.NodeType.AI_RESPONSE);
+      return nextID;
     }
 
     if (node.source === BaseUtils.ai.DATA_SOURCE.KNOWLEDGE_BASE) {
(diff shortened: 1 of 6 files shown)


lib/services/runtime/handlers/aiResponse.ts
```diff
@@ -19,8 +19,17 @@ const AIResponseHandler: HandlerFactory<VoiceNode.AIResponse.Node> = () => ({
...
 
-    if (!(await runtime.services.billing.checkQuota(workspaceID, 'OpenAI Tokens'))) {
-      throw new VError('Quota exceeded', VError.HTTP_STATUS.PAYMENT_REQUIRED);
+    if (!(await runtime.services.billing.checkQuota(workspaceID, QuotaName.OPEN_API_TOKENS))) {
+      addOutputTrace(
+        runtime,
+        getOutputTrace({
+          output: generateOutput('[token quota exceeded]', runtime.project),
+          version: runtime.version,
+          ai: true,
```
pmvrmc, line 28: maybe we had an extra param like `error: true` here to explicit say this is an error?

lib/services/runtime/handlers/aiSet.ts
```diff
@@ -15,8 +15,9 @@ const AISetHandler: HandlerFactory<BaseNode.AISet.Node> = () => ({
 
     if (!node.sets?.length) return nextID;
 
-    if (!(await runtime.services.billing.checkQuota(workspaceID, 'OpenAI Tokens'))) {
-      throw new VError('Quota exceeded', VError.HTTP_STATUS.PAYMENT_REQUIRED);
+    if (!(await runtime.services.billing.checkQuota(workspaceID, QuotaName.OPEN_API_TOKENS))) {
+      runtime.trace.debug('token quota exceeded', BaseNode.NodeType.AI_SET);
```
pmvrmc, line 19: on the ai response we add the debug + outuput trace, in here we just add the debug. Shouldnt we apply the same logic somehow?

lib/services/runtime/handlers/noMatch.ts
```diff
@@ -70,16 +70,28 @@ const getOutput = async (
   }
 
   if (runtime.project?.aiAssistSettings?.aiPlayground) {
+    const workspaceID = runtime.project?.teamID;
+
+    if (!(await runtime.services.billing.checkQuota(workspaceID, QuotaName.OPEN_API_TOKENS))) {
+      runtime.trace.debug('token quota exceeded', BaseNode.NodeType.AI_SET);
+      return { output: generateOutput('global no match [token quota exceeded]', runtime.project), ai: true };
```
pmvrmc, line 77: should we return the actual global no match (from the project config), instead of the string 'global no match'?

lib/services/runtime/handlers/aiResponse.ts
DecathectZero: I think we want to think through it, and what a proper error system should look like instead of jamming it in here.

lib/services/runtime/handlers/aiSet.ts
DecathectZero: `aiSet` technically isn't supposed to actually say anything back to the user, which is why I've just let it be silent for now.

lib/services/runtime/handlers/noMatch.ts
DecathectZero: This is actually the global no match from project config:
Screenshot 2023-07-07 at 3 22 40 PM

Also if we fall back to a static it still gives the user the impression that something is broken on Voiceflow - the most important thing is that they realize it's on their end.

pmvrmc (approved):

trs (approved):

PR #550, opened 2023-07-13
fix: floor token calculation for claude (BUG-490)

trs:

Floors the token calculation for claude so we aren't sending floats to billing.

Also catches errors around consumption as this is a side effect and shouldn't block the user.


DecathectZero (approved):

trs:
bors merge

PR #551, opened 2023-07-18
feat: implement response step handler (PL3-50)

zhihil:
**Fixes or implements PL3-50**

### Related PRs

### Checklist


lib/services/runtime/handlers/v3/response/variant/base.variant.ts
```diff
@@ -0,0 +1,45 @@
...
+  public get type(): Variant['type'] {
+    return this.rawVariant.type;
+  }
+
+  public get condition(): Condition | null {
+    if (this.rawVariant.condition === null) return null;
+    if (!this.cachedCondition) {
+      this.cachedCondition = buildCondition(this.rawVariant.condition, this.varContext);
+    }
+    return this.cachedCondition;
```
zhihil, line 33: This caching might be unnecessary in hindsight? This will essentially just save one object constructor call per variant.

lib/services/runtime/handlers/v3/response/evaluateVariant/evaluateCarousel/evaluateCarousel.ts
```diff
@@ -0,0 +1,42 @@
...
+function convertLayout(cardLayout: CardLayout): BaseNode.Carousel.CarouselLayout {
+  switch (cardLayout) {
+    case CardLayout.CAROUSEL:
+      return BaseNode.Carousel.CarouselLayout.CAROUSEL;
+    case CardLayout.LIST:
+      return BaseNode.Carousel.CarouselLayout.LIST;
+    default:
+      throw new VError('unexpected value for card layout');
+  }
+}
```
zhihil, line 16: Probably could just turn this function into a map.

lib/services/runtime/handlers/v3/response/language-generator/billed.generator.ts
```diff
@@ -0,0 +1,69 @@
...
+import { addOutputTrace, getOutputTrace } from '@/lib/services/runtime/utils';
+import log from '@/logger';
+import { Runtime } from '@/runtime';
+
+import { generateOutput } from '../../../utils/output';
+import { BaseLanguageGenerator } from './base.generator';
+import { BaseLanguageGeneratorReturn } from './base.interface';
+import { AIBillingEvents } from './billed.interface';
+
+export class BilledGenerator<
```
zhihil, line 14: This class is essentially a "Dynamic Decorator" pattern.

The design problem here is that I need to inject billing utilities into `origGenerator` such that:
1. The `origGenerator` is abstracted from the billing calls 
2. If the billing callback fails, then the `.generate()` call fails. There isn't a very clean abstraction for this using events or synchronous hooks. 
3. The billing calls are dependent on a dynamic value, the `runtime`, which may differ with each request.

A decorator like `AuthGuard` provides the right abstraction, but this is a compile-time construct that doesn't really fit the constraints above.

lib/services/runtime/handlers/v3/response/variant/variant.ts
```diff
@@ -0,0 +1,28 @@
...
+    case ResponseVariantType.TEXT:
+      return new TextVariant(rawVariant, varContext);
+    case ResponseVariantType.JSON:
+      return new JSONVariant(rawVariant, varContext);
+    case ResponseVariantType.PROMPT:
+      return new PromptVariant(rawVariant, varContext, langGen, chatHistory);
+    default:
+      throw new VError('unknown variant type encountered');
+  }
+}
```
zhihil, line 28: The response step handler has crazy high latency when testing locally. 

I'm not sure where it's coming from, but I'm a bit concerned it's due to a lot of object allocations here, so I'll have to benchmark the response step handler and see if these object allocations need to be replaced with function calls to avoid touching the heap.

zhihil: Nevermind, the insane latency of `general-runtime` locally is just how it is apparently:

Screen Shot 2023-08-17 at 4 06 51 PM

PR #558, opened 2023-07-27
[Merged by Bors] - feat: add nlu inference endpoint (PL-551)

pmvrmc:
**Fixes or implements PL-551**

Adds a new inference endpoint as requested by ML team.

Trying to add a proper route with params without breaking much of the runtime route flow.
Created new middleware to ensure all params are set also as headers so that compatibility with older middlewares is guaranteed. Hopefully in the future we move out of this old route patterns and can set proper route standards.

### DIFF
--- a/backend/api/routers/nlu.ts
+++ b/backend/api/routers/nlu.ts
@@ -6,9 +6,9 @@ export default (middlewares: MiddlewareMap, controllers: ControllerMap) => {
   const router = express.Router();
 
   router.get(
-    '/inference',
-    middlewares.project.resolveVersionAlias,
-    middlewares.project.attachProjectID,
+    '/project/:projectID/version/:versionID/inference',
+    middlewares.auth.authorize(['project:READ']),
+    middlewares.project.paramsToLegacyHeader,
     middlewares.rateLimit.versionConsume,
     controllers.nlu.inference
   );
--- a/backend/api/routers/nlu.ts
+++ b/backend/api/routers/nlu.ts
@@ -1,7 +1,5 @@
-import bodyParser from '@voiceflow/body-parser';
 import express from 'express';
 
-import { BODY_PARSER_SIZE_LIMIT } from '@/backend/constants';
 import { ControllerMap, MiddlewareMap } from '@/lib';
 
 export default (middlewares: MiddlewareMap, controllers: ControllerMap) => {
--- a/backend/api/routers/nlu.ts
+++ b/backend/api/routers/nlu.ts
@@ -1,0 +1,19 @@
+import bodyParser from '@voiceflow/body-parser';
+import express from 'express';
+
+import { BODY_PARSER_SIZE_LIMIT } from '@/backend/constants';
+import { ControllerMap, MiddlewareMap } from '@/lib';
+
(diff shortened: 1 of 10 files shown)


lib/controllers/nlu.ts
```diff
@@ -0,0 +1,48 @@
...
+  @validate({
+    PARAMS_PROJECT_ID: VALIDATIONS.PARAMS.PROJECT_ID,
+    PARAMS_VERSION_ID: VALIDATIONS.PARAMS.VERSION_ID,
+    QUERY: VALIDATIONS.QUERY.QUERY,
+  })
+  async inference(req: Request) {
+    const { versionID, projectID } = req.params;
+
+    const api = await this.services.dataAPI.get();
+    const [version, project] = await Promise.all([api.getVersion(versionID), api.getProject(projectID)]);
```
trs, line 32: We should ensure that the version's project is the given project ID. Otherwise we could be set totally unrelated IDs.

backend/api/routers/nlu.ts
```diff
@@ -0,0 +1,17 @@
+import express from 'express';
+
+import { ControllerMap, MiddlewareMap } from '@/lib';
+
+export default (middlewares: MiddlewareMap, controllers: ControllerMap) => {
+  const router = express.Router();
+
+  router.get(
+    '/project/:projectID/version/:versionID/inference',
```
trs, line 9: It's a _bit_ redundant to ask for both project and version, since we can get the project ID from the version. But this format does let us support version tags (`development`, `production`) in the future.

lib/middlewares/auth.ts
```diff
@@ -61,7 +61,10 @@ class Auth extends AbstractMiddleware {
         const sdk = await import('@voiceflow/sdk-auth/express').catch(() => null);
         if (!sdk) return next();
 
-        return sdk?.createAuthGuard(client)(actions as any[])(req, res, next);
+        // we cant return immediatly here or the catch is not caught properly
+        // eslint-disable-next-line sonarjs/prefer-immediate-return
+        const authorizedNext = await sdk?.createAuthGuard(client)(actions as any[])(req, res, next);
+        return authorizedNext;
```
trs, line 67:
```suggestion
        return await sdk?.createAuthGuard(client)(actions as any[])(req, res, next);
```

We could just an an `await` here too.

trs: Maybe it's worth updating the SDK to call `next` if it catches an error so we don't have to worry about it during implementation.

lib/controllers/nlu.ts
pmvrmc: ah true! I was assuming auth bit did that, but I think it just ensures the versionID is valid

backend/api/routers/nlu.ts
pmvrmc: that was the idea 😅

lib/middlewares/auth.ts
pmvrmc: yes! maybe something simple we can add to a utility sprint? there was a list of these smaller improvements to do, but cant find it

PR #560, opened 2023-07-31
[Merged by Bors] - feat: better token logging (NLU-740)

e-vandenberg:
**Fixes or implements NLU-740**

Add better tracing to knowledge-base and generative-ai steps, in particular token information which is a requirement for a new testing feature.

The idea behind the change is that whenever a user is charged for tokens, the information should be available in the trace.

knowledge-base trace:
Screenshot 2023-07-31 at 11 38 08 AM

generative ai trace:
Screenshot 2023-07-31 at 11 39 02 AM

### DIFF
--- a/lib/clients/ai/anthropic/utils.ts
+++ b/lib/clients/ai/anthropic/utils.ts
@@ -60,7 +60,6 @@ export abstract class AnthropicAIModel extends AIModel {
     )}${AI_PROMPT}`;
 
     const queryTokens = this.calculateTokenUsage(prompt);
-    let tokens = queryTokens;
 
     const result = await this.client
       .complete({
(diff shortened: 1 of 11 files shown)


lib/clients/ai/anthropic/utils.ts
```diff
@@ -76,11 +77,14 @@ export abstract class AnthropicAIModel extends AIModel {
 
     const output = result?.completion?.trim() ?? null;
 
+    const answerTokens = this.calculateTokenUsage(output ?? '');
     tokens += this.calculateTokenUsage(output ?? '');
 
     return {
       output,
       tokens,
```
trs, line 85: Can we move the total count to here, instead of tracking "two" sets of token counts throughout?

```ts
const queryTokens = ...
...
const answerTokens = ...
...

return {
  tokens: queryTokens + answerTokens,
  ...
}
```

trs (approved):

PR #563, opened 2023-08-01
fix: token cleanup (PL-000)

DecathectZero:
we implemented the whole token system in a massive rush because people were abusing the APIs. 
This is just some clean up work.


lib/services/runtime/handlers/utils/knowledgeBase/answer.ts
```diff
@@ -80,7 +80,7 @@ export const answerSynthesis = async ({
   const output = response.output?.trim().toUpperCase();
 
   if (output?.includes('NOT_FOUND') || output?.startsWith("I'M SORRY,") || output?.includes('AS AN AI'))
-    return { output: null };
+    return { ...response, output: null };
```
DecathectZero, line 83: we were not adding to token count properly before when KB doesn't find anything.

lib/services/runtime/handlers/utils/ai.ts
```diff
@@ -19,17 +22,24 @@ export interface AIResponse {
   output: string | null;
   messages?: BaseUtils.ai.Message[];
   prompt?: string;
-  tokens?: number;
-  queryTokens?: number;
-  answerTokens?: number;
+  tokens: number;
+  queryTokens: number;
+  answerTokens: number;
```
DecathectZero, line 27: make these required just for consistency

e-vandenberg (approved):

PR #569, opened 2023-08-10
feat: add Segment AI requests events

abuyak-vf:

### Related PRs

### Checklist


lib/services/runtime/handlers/noMatch.ts
```diff
@@ -78,14 +129,25 @@ const getOutput = async (
...
     }
 
     // hit global no match if KB wasn't successful
     if (!result?.output && globalNoMatch?.type === BaseVersion.GlobalNoMatchType.GENERATIVE) {
+      const startTime = performance.now();
       result = await generateNoMatch(runtime, globalNoMatch.prompt);
+
+      const action_type = AiRequestActionType.AI_GLOBAL_NO_MATCH;
+
+      await sendSegmentEvent(runtime, action_type, startTime, result).catch(() => null);
```
DecathectZero, line 149: we should do this async, don't need to `await` it. It should run in the background to be non-blocking

DecathectZero (changes requested):
I think for all segment events, we should do this async, don't need to `await` it. It should run in the background to be non-blocking to the response.

If there is an error, we can `.catch` it and log it, but doesn't need to be part of the main interaction thread.

trs:
> I think for all segment events, we should do this async, don't need to `await` it. It should run in the background to be non-blocking to the response.
> 
> If there is an error, we can `.catch` it and log it, but doesn't need to be part of the main interaction thread.

We're actually adding the analytics SDK to this repo, which will make this easier (https://voiceflow.atlassian.net/browse/PL-556)

PR #571, opened 2023-08-10
[Merged by Bors] - feat: adding test endpoint for functions (PL3-90)

zhihil:
**Fixes or implements PL3-90**

Added a test endpoint for Functions to support the Figma designs [here](https://www.figma.com/file/L0eEAZKUWiehEOqpKnhKYI/Functions?node-id=456%3A296234&mode=dev). 

This PR only defines the interface to unblock CxD. The implementation is mocked here and will be completed at a later date for the P0 release.

Defined a new `TestController.testFunction` method, as well as its request interface (`TestFunctionBody`, `TestFunctionParams`) and response interface (`TestFunctionResponse`). These are the main items that require review. 

Defined a new `TestService` that invokes `FunctionRunner` which will contain the main implementation for testing functions. These files will house the implementation in the future and are low-priority items for review.

### Setup information

N/A

### Deployment Notes

N/A

### Related PRs

N/A

### Checklist


backend/api/routers/test.ts
```diff
@@ -11,6 +11,13 @@ export default (middlewares: MiddlewareMap, controllers: ControllerMap) => {
   router.use(bodyParser.json({ limit: BODY_PARSER_SIZE_LIMIT }));
   router.use(middlewares.rateLimit.verify);
 
+  router.post(
+    '/workspaces/:workspaceID/functions/:functionID/',
```
zhihil, line 15: Full path is `/test/workspaces/:workspaceID/functions/:functionID`.

I would prefer `/test/functions/:functionID` or `/functions/:functionID/test` but we don't have granular auth permissions for Functions specifically, so I'm going to use the existing workspace permissions.

pmvrmc (approved):
looks good, but I would doublecheck the naming convention for controller so it doesnt look odd

lib/controllers/test/test.controller.ts
```diff
@@ -127,6 +128,37 @@ class TestController extends AbstractController {
...
     return { output };
   }
+
+  async testFunction(req: Request<TestFunctionParams, TestFunctionBody>): Promise<TestFunctionResponse> {
+    const {
+      params: { functionID },
+      body: inputMapping,
+    } = req;
+
+    await this.services.test.testFunction(functionID, inputMapping);
```
pmvrmc, line 138: shouldnt this call the actual function service and not a bogus test service?

lib/controllers/index.ts
```diff
@@ -7,7 +7,7 @@ import Interact from './interact';
 import NLU from './nlu';
 import Public from './public';
 import StateManagement from './stateManagement';
-import Test from './test';
+import Test from './test/test.controller';
```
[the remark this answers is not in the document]
zhihil: It would better if we move towards this newer naming convention, but naming it all in this one PR would obfuscate what the main changes are. 

I'll change it back

lib/controllers/test/test.controller.ts
zhihil: There isn't a function service though. The goal here is to eventually call the function handler rather than reimplementing the function handler logic. To setup a call to the function handler, this would be done within the `TestService.testFunction()`.

Additionally, I think we should move some business logic into a separate `TestService` rather than putting into this `TestController`. The controller level should only worry about validation, but parts like `testKnowledgeBase` are arranging the business logic which is a service-level concern. 

Similar to my comment about the file naming, I didn't migrate all of the controller methods to the service to avoid getting bogged down in unnecessary changes. We aren't able to implement the test endpoint here until the Content API is done because Tyler S says it's blocking this test endpoint. 

We aren't doing the implementation for the scope of this PR anyways (see the PR description). This is just to define the interface so CxD gets unblocked.

pmvrmc: I get it but atm we could replace most of the changes by a comment like (even if we dont replace, adding this makes sense):
```
// TODO atm we just need to return something for the CxD team, but in the future this would call the actual function code and return the proper response
```

adding a new `TestService` makes 100% sense and even migrating the other test endpoints to use it, but feels a bit weird we are pushing some code that is doing nothing and only limits the final solution.

for example in the future we would need to pull the function by that id, throwing bad request if it doesnt exist which is something we can do at the controller level.

from now to then many things can happen, make this code unviable and it will just sit there.

but this is really just a nitpick if you think otherwise lgtm 💪

PR #573, opened 2023-08-10
[Merged by Bors] - fix: checking quota broken [bugfix] (LLM-000)

DecathectZero:
We had this line on `consumeQuota`: `if (!(typeof count === 'number' && count > 0)) return null;`

and `checkQuota` works by seeing if `await this.consumeQuota(workspaceID, quotaName, 0);` throws and error or not.

Right now it will just always return `null` so `checkQuota` is ALWAYS true. I moved that  filter one level higher and tested.


JoshTheDeveloper (approved):

DecathectZero:
Bors r+

PR #579, opened 2023-08-16
feat: ff for us bank (NLU-000)

e-vandenberg:
US bank wants the KB for all workspaces.

### DIFF
--- a/lib/services/runtime/handlers/utils/knowledgeBase/index.ts
+++ b/lib/services/runtime/handlers/utils/knowledgeBase/index.ts
@@ -32,8 +32,11 @@ const scheme = process.env.NODE_ENV === 'e2e' ? 'https' : 'http';
 export const RETRIEVE_ENDPOINT = host && port ? new URL(`${scheme}://${host}:${port}/retrieve`).href : null;
 export const { KNOWLEDGE_BASE_LAMBDA_ENDPOINT } = Config;
 
-export const getAnswerEndpoint = (workspaceID: string | undefined): string | null => {
-  if (workspaceID && FLAGGED_WORSPACE_IDS.includes(String(workspaceID))) {
+export const getAnswerEndpoint = (cloud_env: string, workspaceID: string | undefined): string | null => {
+  if (cloud_env === 'public' && workspaceID && FLAGGED_WORSPACE_IDS.includes(String(workspaceID))) {
+    return RETRIEVE_ENDPOINT;
+  }
+  if (cloud_env === 'usbank') {
     return RETRIEVE_ENDPOINT;
   }
   if (!KNOWLEDGE_BASE_LAMBDA_ENDPOINT) return null;
@@ -47,7 +50,8 @@ export const fetchKnowledgeBase = async (
   settings?: BaseModels.Project.KnowledgeBaseSettings
 ): Promise<KnowledgeBaseResponse | null> => {
   try {
-    const answerEndpoint = getAnswerEndpoint(workspaceID);
+    const cloud_env = Config.CLOUD_ENV || '';
+    const answerEndpoint = getAnswerEndpoint(cloud_env, workspaceID);
 
     if (!answerEndpoint) return null;
 
--- a/tests/lib/services/runtime/handlers/knowledgeBase.unit.ts
+++ b/tests/lib/services/runtime/handlers/knowledgeBase.unit.ts
@@ -11,15 +11,31 @@ describe('knowledgebase retrieval unit tests', () => {
     it('calls new KB when appropriate', async () => {
       // id for the Python KB Testing workspace created to test this feature
       const flaggedWorkspaceID = '80627';
-      const endpoint = getAnswerEndpoint(flaggedWorkspaceID);
+      const endpoint = getAnswerEndpoint('public', flaggedWorkspaceID);
 
       expect(endpoint).to.eql(RETRIEVE_ENDPOINT);
     });
 
-    it('calls old KB when appropriate', async () => {
+    it('calls new KB for usbank', async () => {
+      // all workspaces are enabled
+      const endpoint = getAnswerEndpoint('usbank', 'anything');
+
+      expect(endpoint).to.eql(RETRIEVE_ENDPOINT);
+    });
+
+    it('calls old KB when appropriate (wrong workspace)', async () => {
       const nonFlaggedWorkspaceID = '123';
       const expectedEndpoint = `${KNOWLEDGE_BASE_LAMBDA_ENDPOINT}/answer`;
-      const endpoint = getAnswerEndpoint(nonFlaggedWorkspaceID);
+      const endpoint = getAnswerEndpoint('public', nonFlaggedWorkspaceID);
+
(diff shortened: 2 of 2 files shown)


lib/services/runtime/handlers/utils/knowledgeBase/index.ts
```diff
@@ -32,8 +32,11 @@ const scheme = process.env.NODE_ENV === 'e2e' ? 'https' : 'http';
 export const RETRIEVE_ENDPOINT = host && port ? new URL(`${scheme}://${host}:${port}/retrieve`).href : null;
 export const { KNOWLEDGE_BASE_LAMBDA_ENDPOINT } = Config;
 
-export const getAnswerEndpoint = (workspaceID: string | undefined): string | null => {
-  if (workspaceID && FLAGGED_WORSPACE_IDS.includes(String(workspaceID))) {
+export const getAnswerEndpoint = (cloud_env: string, workspaceID: string | undefined): string | null => {
+  if (cloud_env === 'public' && workspaceID && FLAGGED_WORSPACE_IDS.includes(String(workspaceID))) {
```
pmvrmc, line 36: can we move the public string to a constant?

lib/services/runtime/handlers/utils/knowledgeBase/index.ts
```diff
@@ -32,8 +32,11 @@ const scheme = process.env.NODE_ENV === 'e2e' ? 'https' : 'http';
...
+    return RETRIEVE_ENDPOINT;
+  }
+  if (cloud_env === 'usbank') {
```
pmvrmc, line 39: can we do instrad something like `if (FLAGGED_CLOUD_ENVS.includes(cloud_env))` ?
hardcoded strings are an antipattern and dont scale

lib/services/runtime/handlers/utils/knowledgeBase/index.ts
zhihil:
```ts
// knowledgeBase/knowledgeBase.types.ts or somewhere else
export enum CloudEnv {
   Public = 'public',
   Private = 'private',
}
```

lib/services/runtime/handlers/utils/knowledgeBase/index.ts
zhihil: Once this PR merged in, `FLAGGED_CLOUD_ENVS` should ideally be replaced with Unleash feature flagging in a follow-up PR as well. 

This how the VFNLU is being locked off in the `nlu-gateway` and it's been key in managing FF complexity.

Screen Shot 2023-08-17 at 9 49 03 AM

Screen Shot 2023-08-17 at 9 50 18 AM

We don't support FF on `general-runtime` so an Unleash client may need to be implemented from scratch.

e-vandenberg:
I reworked the flag definitions a little differently than suggested due to the fact that certain envs have certain workspaces that need to be flagged. 

Let me know what you think of these changes or if you'd rather me define an enum for the envs as suggested.

I agree with the unleash comment, I've been working with it in the creator-app and its much nicer for managing these sorts of things.

lib/services/runtime/handlers/utils/knowledgeBase/index.ts
```diff
@@ -25,17 +25,24 @@ export interface KnowledgeBaseResponse {
...
+flaggedWorkspaces.usbank = [];
 
 const { KL_RETRIEVER_SERVICE_HOST: host, KL_RETRIEVER_SERVICE_PORT: port } = Config;
 const scheme = process.env.NODE_ENV === 'e2e' ? 'https' : 'http';
 export const RETRIEVE_ENDPOINT = host && port ? new URL(`${scheme}://${host}:${port}/retrieve`).href : null;
 export const { KNOWLEDGE_BASE_LAMBDA_ENDPOINT } = Config;
 
-export const getAnswerEndpoint = (workspaceID: string | undefined): string | null => {
-  if (workspaceID && FLAGGED_WORSPACE_IDS.includes(String(workspaceID))) {
+export const getAnswerEndpoint = (cloud_env: string, workspaceID: string): string | null => {
```
pmvrmc, line 37: cloudEnv, not cloud_env

lib/services/runtime/handlers/utils/knowledgeBase/index.ts
```diff
@@ -25,17 +25,24 @@ export interface KnowledgeBaseResponse {
   chunks: KnowledegeBaseChunk[];
 }
 
-export const FLAGGED_WORSPACE_IDS = ['80627', 'Brk8AaGjlQ'];
+const flaggedWorkspaces: { [env: string]: string[] } = {};
+flaggedWorkspaces.public = ['80627', 'Brk8AaGjlQ'];
+flaggedWorkspaces.usbank = [];
```
pmvrmc, line 30: you can simplify it via a map
```
const FLAGGED_WORKSPACES_MAP = new Map<string, string[]>([
  ['public', ['80627', 'Brk8AaGjlQ']],
  ['usbank', []],
])
```

lib/services/runtime/handlers/utils/knowledgeBase/index.ts
```diff
@@ -25,17 +25,24 @@ export interface KnowledgeBaseResponse {
...
 const { KL_RETRIEVER_SERVICE_HOST: host, KL_RETRIEVER_SERVICE_PORT: port } = Config;
 const scheme = process.env.NODE_ENV === 'e2e' ? 'https' : 'http';
 export const RETRIEVE_ENDPOINT = host && port ? new URL(`${scheme}://${host}:${port}/retrieve`).href : null;
 export const { KNOWLEDGE_BASE_LAMBDA_ENDPOINT } = Config;
 
-export const getAnswerEndpoint = (workspaceID: string | undefined): string | null => {
-  if (workspaceID && FLAGGED_WORSPACE_IDS.includes(String(workspaceID))) {
+export const getAnswerEndpoint = (cloud_env: string, workspaceID: string): string | null => {
+  // check if env/workspace pair is flagged, if flagged workspaces list is empty, accept them all
+  if (
```
pmvrmc, line 39: and using the map
```
const flaggedWorkspaces  = FLAGGED_WORKSPACES_MAP.get(cloudEnv);
if (flaggedWorkspaces?.length === 0 || flaggedWorkspaces?.includes(String(workspaceID)) {
  return RETRIEVE_ENDPOINT;
}
```

lib/services/runtime/handlers/utils/knowledgeBase/index.ts
```diff
@@ -47,7 +54,8 @@ export const fetchKnowledgeBase = async (
   settings?: BaseModels.Project.KnowledgeBaseSettings
 ): Promise<KnowledgeBaseResponse | null> => {
   try {
-    const answerEndpoint = getAnswerEndpoint(workspaceID);
+    const cloud_env = Config.CLOUD_ENV || '';
```
pmvrmc, line 57: cloudEnv not cloud_env

pmvrmc (approved):

PR #581, opened 2023-08-22
[Merged by Bors] - feat: intent confidence (NLU-805)

DecathectZero:
check intent confidence, otherwise turn into None Intent.


lib/services/nlu/index.ts
```diff
@@ -90,7 +93,13 @@ class NLU extends AbstractManager<{ utils: typeof utils }> implements ContextHan
         .catch(() => ({ data: null }));
 
       if (data) {
-        const intentRequest = this.adaptNLUPrediction(data);
+        let intentRequest = this.adaptNLUPrediction(data);
+
+        if (!isConfidenceScoreAbove(intentConfidence, intentRequest.payload.confidence)) {
+          // confidence of a none intent is inverse to the confidence of the predicted intent
+          intentRequest = getNoneIntentRequest({ query, confidence: 1 - intentRequest.payload.confidence! });
+        }
```
DecathectZero, line 101: this is basically the CORE logic.

lib/services/nlu/index.ts
```diff
@@ -90,7 +93,13 @@ class NLU extends AbstractManager<{ utils: typeof utils }> implements ContextHan
         .catch(() => ({ data: null }));
 
       if (data) {
-        const intentRequest = this.adaptNLUPrediction(data);
+        let intentRequest = this.adaptNLUPrediction(data);
+
+        if (!isConfidenceScoreAbove(intentConfidence, intentRequest.payload.confidence)) {
+          // confidence of a none intent is inverse to the confidence of the predicted intent
+          intentRequest = getNoneIntentRequest({ query, confidence: 1 - intentRequest.payload.confidence! });
```
zhihil, line 100: Slightly concerned about this expression `1 - intentRequest.payload.confidence`. 

Denys said the VFNLU should always return a confidence if it returns an intent. 

I'm not too sure about LUIS though. 

I would double-check if we accidentally evaluated `1 - undefined` or `1 - null` on LUIS or VFNLU somehow.

DecathectZero: if you look at the logic of `isConfidenceScoreAbove` it will never enter the clause unless `intentRequest.payload.confidence` is defined and a number

DecathectZero: I mean I can make the logic more clear

zhihil: Oh, I see what you mean now.

zhihil (approved):

lib/services/nlu/index.ts
DecathectZero: I've made it a little more redundant but far more clear. It requires some thinking to wrap your head around the confidence score stuff

PR #589, opened 2023-09-11
[Merged by Bors] - feat: add new env names for kl-retriever (NLU-825)

bhuvana-ak:
**Fixes or implements NLU-825**

Add new cloud env names in general-runtime to access kl-retriever.

New Types added for new envs.

### DIFF
--- a/lib/services/runtime/handlers/utils/knowledgeBase/types.ts
+++ b/lib/services/runtime/handlers/utils/knowledgeBase/types.ts
@@ -1,4 +1,7 @@
 export enum CloudEnv {
   Public = 'public',
   USBank = 'usbank',
+  JPMC = 'jpmc',
+  CISCO = 'cisco',
+  VFTEST76 = 'vftest-76',
 }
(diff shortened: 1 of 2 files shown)


lib/services/runtime/handlers/utils/knowledgeBase/types.ts
```diff
@@ -1,4 +1,7 @@
 export enum CloudEnv {
   Public = 'public',
   USBank = 'usbank',
+  JPMC = 'jpmc',
+  CISCO = 'cisco',
```
pmvrmc, line 5: why add these ones if they are not used?

bhuvana-ak: These 2 envs will be enabled after testing the functionality in vf-test.
Soon, I'll have a follow up pr for that..

pmvrmc: i though so! totally fine, but ideally they would be added when they are needed!

lets get it merged 👍

PR #591, opened 2023-09-18
[Merged by Bors] - feat: update general runtime to use versionID and diagramID to fetch programs (CV3-361)

viniciusdacal:
**Fixes or implements CV3-361**

### Related PRs

### Checklist


lib/controllers/stateManagement/requests.ts
```diff
@@ -3,10 +3,10 @@ import { ArrayField, BooleanField, ObjectField, StringField } from '@/lib/contro
 const StackFrame = {
   type: 'object',
   additionalProperties: true,
-  required: ['programID'],
+  required: ['diagramID'],
   properties: {
     nodeID: StringField('nodeID'),
-    programID: StringField('programID'),
+    diagramID: StringField('diagramID'),
```
DecathectZero, line 9: this might break some people calling our API, but I think it should be okay.

viniciusdacal: Yeah, I think the alternative would be defining both properties and making none required, and adding `diagramID` as required a couple days after. 

But those errors should happen only if there's a conversation that started before the release, right? So I guess it's ok to break, because the person can just restart the conversation.

runtime/lib/DataAPI/localDataAPI.ts
```diff
@@ -28,7 +28,7 @@ class LocalDataAPI<
 
   public getVersion = async () => this.version;
 
-  public getProgram = async (programID: string) => this.programs[programID];
+  public getProgram = async (_versionID: string, diagramID: string) => this.programs[diagramID];
```
z4o4z, line 31: I assume `diagramID` is not unique (it's `diagramID` field not `_id`) here, right? if so we should cache programs per version just to be extra safe
```suggestion
  public getProgram = async (versionID: string, diagramID: string) => this.programs[versionID]?.[diagramID] ?? null;
```

runtime/lib/DataAPI/mongoDataAPI.ts
```diff
@@ -61,6 +72,10 @@ class MongoDataAPI<
   };
 
   public getProject = async (projectID: string) => {
```
z4o4z, line 74: looks like we should rename param
```suggestion
  public getProject = async (projectIDOrAuth: string) => {
```

runtime/lib/DataAPI/mongoDataAPI.ts
```diff
@@ -72,6 +87,28 @@ class MongoDataAPI<
     return shallowObjectIdToString(project);
   };
 
+  public getProjectByAPIKey = async (ref: string) => {
```
z4o4z, line 90:
```suggestion
  public getProjectByAPIKey = async (auth: string) => {
```

runtime/lib/Runtime/index.ts
```diff
@@ -186,24 +186,24 @@ class Runtime<
...
-    const program = await this.api.getProgram(this.versionID).catch(() => null);
+    const program = await this.api.getProgram(this.versionID, this.versionID).catch(() => null);
     if (!program) {
       return;
     }
 
     this.stack.unshift(
       new Frame({
-        programID: this.versionID,
+        diagramID: this.versionID,
```
z4o4z, line 206: so we are storing `versionID` into a `diagramID`, probably it makes sense to have a different name in the `Frame` just to show not only `diagramID` is stored in this field...

```suggestion
        programID: this.versionID,
        // or
        frameID: this.versionID,
```

runtime/lib/Runtime/utils/programManager.ts
```diff
@@ -11,27 +11,31 @@ class ProgramManager {
...
       program = this.cachedProgram;
     }
 
     if (!program) {
-      program = new ProgramModel(await this.runtime.api.getProgram(programID));
+      program = new ProgramModel(await this.runtime.api.getProgram(versionID, diagramID));
     }
 
-    this.runtime.callEvent(EventType.programDidFetch, { programID, program });
+    this.runtime.callEvent(EventType.programDidFetch, { programID: diagramID, program });
```
z4o4z, line 38: i guess we should add `versionID` into all vents that have `programID`. should we rename `programID` into `diagramID` here as well?

z4o4z (changes requested):

runtime/lib/Runtime/index.ts
viniciusdacal: No sure if I get the idea.

In this case, the `diagramID` will be the `versionID` because it's the base program. 
During the compilation we create a root program, which its `_id` and `diagramID` will be the same as the `versionID`.

That's why I think it still makes sense to keep the diagramID property.

z4o4z (approved):

DecathectZero (approved):

andreasmartin:
@viniciusdacal and @DecathectZero, I had to revert this commit (7bec1d4681a327ebe184a1eb9c673ff7ddff2bc8) because it was breaking the Dialog API requests such as:
```bash
curl -L '/state/user/userid/interact' `
-H 'Authorization: VF.DM.xx.xx' `
-H 'accept: application/json' `
-H 'content-type: application/json' `
-H 'versionID: development' `
-d '
{
  "action": {
    "type": "text",
    "payload": "Hi"
  }
}
'
```
With your changes, I received the following:
```JSON
{
    "name": "verror",
    "message": "Cannot resolve project version, please verify that your API key is correct.",
    "requestID": "nCfPsSE7ux"
}
```

PR #596, opened 2023-09-27
[Merged by Bors] - feat: use identity guard (PL-625)

trs:
Utilize standard identity guard, remove `verifyDMAPIKey`

### DIFF
--- a/lib/middlewares/auth.ts
+++ b/lib/middlewares/auth.ts
@@ -1,4 +1,3 @@
-import { BaseModels } from '@voiceflow/base-types';
 import VError from '@voiceflow/verror';
 import fetch from 'node-fetch';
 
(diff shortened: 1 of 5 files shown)


lib/middlewares/auth.ts
```diff
@@ -51,49 +38,59 @@ class Auth extends AbstractMiddleware {
...
+  };
 
-      if (!identity?.identity?.id) throw new Error();
+  verifyIdentity = async (req: Request, _res: Response, next: Next): Promise<void> => {
+    try {
+      const guard = await this.getIdentityGuard();
+      if (!guard) return next();
 
-      return next();
+      return guard()(req, res, next);
```
zhihil, line 89: Could we abstract the `getXGuard()` and `verifyX()` methods? Seems their overall logic is similar (e.g. `verifyIdentity()`, `verifyPermissions()`), but they differ only in their strategy, i.e, Identity or Permissions.

zhihil: Nevermind, if this is a temporary pattern, then it's not worth the abstraction.

pmvrmc (approved):
there is a build issue, but code looks good.

theProf (approved):

trs:
bors merge

PR #597, opened 2023-09-28
[Merged by Bors] - feat: moderation llm calls (PL-553)

pmvrmc:
**Fixes or implements PL-553**

We need to add a LLM Moderation call to OpenAI for all models (even the ones that are not from OpenAI). This means that the OpenAI client must always be there for models.

Added a new Unleah client

Tweaked the AI client to be a dynamic instead of static service so that it can support FF via unleash.

Created a new service AISynthesis that pulls some util fns into it so that it can call the new dynamix AI client.

Tweaked some other util fns to make proper use of these new client/service accordingly.

Tweaks also so that the openAI/anthorpic API clients are generated just once, but the AIModel instances are generated on the fly with a given context (workspace/project) so that context can be used during the model operations (as logging the projectID if the moderation call fails)

If no `OPENAI_API_KEY` is given content moderation is skipped completely because atm azure openai doesnt support moderation calls (imagine private clouds where they want any calls be made to "normal openai" they only trust thei own "azure openai").


zhihil (approved):

lib/clients/contentModeration/openai/openai.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    failedModeration.forEach((failedModeration) => {
+      const failedModerationCategories = Object.entries(failedModeration.error.categories).reduce<string[]>(
+        (acc, [key, value]) => {
+          if (value) acc.push(key);
+          return acc;
+        },
+        []
+      );
+      log.warn(`[moderation error] input=${failedModeration.input} | categories=${failedModerationCategories}`);
```
Diophontine, line 49: @pmvrmc can you add the project id too?

pmvrmc: ofc! I figure out  the best way to do that

lib/clients/ai/ai-model.interface.ts
```diff
@@ -0,0 +1,23 @@
...
+  retries?: number;
+  retryDelay?: number;
+}
+
+export interface AIModelContext {
+  workspaceID?: string;
+  projectID?: string;
+}
+
+export const GPT4_ABLE_PLAN = new Set(['old_pro', 'old_team', 'pro', 'team', 'enterprise']);
```
zhihil, line 23: Should use the `PlanType` enum from `@voiceflow/internal`
```ts
import { PlanType } from "@voiceflow/internal";

export const GPT4_ABLE_PLAN = new Set([ PlanType.OLD_PRO, PlanType.OLD_TEAM, PlanType.PRO, PlanType.TEAM, PlanType.ENTERPRISE ]);
```

lib/clients/ai/openai/gpt.ts
```diff
@@ -1,58 +1,35 @@
...
   static RoleMapping = {
     [BaseUtils.ai.Role.ASSISTANT]: ChatCompletionRequestMessageRoleEnum.Assistant,
     [BaseUtils.ai.Role.SYSTEM]: ChatCompletionRequestMessageRoleEnum.System,
     [BaseUtils.ai.Role.USER]: ChatCompletionRequestMessageRoleEnum.User,
   };
 
-  constructor(config: Config) {
-    super(config.AI_GENERATION_TIMEOUT);
-
-    if (isAzureBasedGPTConfig(config)) {
```
zhihil, line 25: This refactor of the `if`-`else` statement for choosing the Azure client vs. the OpenAI client might be incorrect and may cause a significant regression. 

The Azure client is used by default, I think for performance reasons. However, GPT-4 is not available on Azure, so we need to use the OpenAI client. With the way that the code is refactor, does this still occur? 

It looks like at `lib/clients/ai/index.ts`, we initialize a single OpenAI client which use either the Azure client or the OpenAI client, but not both. What we want is that there are at least two different clients, an OpenAI client for GPT-4 and an Azure client for everything else. With this change, we either:

1. Regress all clients to the OpenAI one, which incurs latency
2. Force the GPT-4 code onto the Azure client, which does not support GPT-4

Do we have confirmation that GPT-4 is available on Azure to make this change? I'm just checking this right now to make sure GPT-4 isn't broken by this change.

cc @DecathectZero

zhihil (changes requested):

lib/services/runtime/handlers/aiResponse.ts
```diff
@@ -16,8 +16,11 @@ const AIResponseHandler: HandlerFactory<VoiceNode.AIResponse.Node> = () => ({
   handle: async (node, runtime, variables) => {
     const nextID = node.nextId ?? null;
     const workspaceID = runtime.project?.teamID;
-    const generativeModel = runtime.services.ai.get(node.model);
-    const kbModel = runtime.services.ai.get(runtime.project?.knowledgeBase?.settings?.summarization.model);
+    const generativeModel = runtime.services.ai.get(node.model, { projectID: runtime.project?._id, workspaceID });
+    const kbModel = runtime.services.ai.get(runtime.project?.knowledgeBase?.settings?.summarization.model, {
+      projectID: runtime.project?._id,
```
zhihil, line 21: Small style change:
```ts
const projectID = runtime.project?._id;
const generativemodel = runtime.services.ai.get(..., { projectID, .... });
const kbModel = runtime.services.ai.get(..., { projectID, ... });
```

lib/services/runtime/handlers/aiSet.ts
```diff
@@ -12,8 +12,11 @@ const AISetHandler: HandlerFactory<BaseNode.AISet.Node> = () => ({
   handle: async (node, runtime, variables) => {
     const nextID = node.nextId ?? null;
     const workspaceID = runtime.project?.teamID;
-    const generativeModel = runtime.services.ai.get(node.model);
-    const kbModel = runtime.services.ai.get(runtime.project?.knowledgeBase?.settings?.summarization.model);
+    const generativeModel = runtime.services.ai.get(node.model, { projectID: runtime.project?._id, workspaceID });
```
zhihil, line 15: Same style change here

lib/clients/ai/openai/gpt.ts
zhihil: I've tested the behaviour and I think we should fix the refactor as I mentioned above.

For some reason, the Azure client is able to resolve GPT-4 requests, even though we've never deployed a GPT-4 model on the cloud platform. Based on my tests, it looks your refactor is causing Azure client to fallback on GPT-3.5 when we select GPT-4.

I tested the AI Response step with GPT-3 vs. GPT-3.5 vs. GPT-4 with and without your changes. I turned the temperature of the AI Response step to 0 to remove randomness. The result is that in the current `master`, GPT-4 has a noticeably higher latency and different output than GPT-3.5. With your changes meshed, GPT-4 and -3.5 have about the same latency and exactly the same output. So I suspect that this refactor is breaking GPT-4.

pmvrmc: good catch. address this gpt4 problem by making sure it always uses the "normal" openAI client instead of the azure one.

I also noticed there was some weird code on the gpt3.5 code, where the client used was not the one passed to the function but the one on `this.client`.

PR #599, opened 2023-10-04
[Merged by Bors] - feat: (feature flagged) Call FAQ endpoint first (NLU-902)

e-vandenberg:
Adds a call to the new poc faq retrieval endpoint before hitting the KB. Now any KB hit will do the following:
1. call /retrieve/poc/faq to see if the question is an faq. Return the answer if so.
2. Otherwise call /retrieve (as usual) to search all chunks, not just faq.

`80627` is our Python KB Testing workspace. Behaviour for all other workspaces in all other environments remains the same.

- [x] Tested in preview, no match, and KB response. Tested for both flagged and non-flagged workspaces.
- [x] Must not be merged before https://github.com/voiceflow/KnowledgeLake/pull/133 
- [x] Must not be merged before https://github.com/voiceflow/creator-app/pull/7226


lib/services/runtime/handlers/utils/knowledgeBase/index.ts
```diff
@@ -34,9 +43,13 @@ const FLAGGED_WORKSPACES_MAP = new Map<string, string[]>([
   [CloudEnv.JPMC, []],
 ]);
 
+const FAQ_FLAGGED_WORKSPACES_MAP = new Map<string, string[]>([[CloudEnv.Public, ['80627']]]);
```
zhihil, line 46: Wait for this PR: https://github.com/voiceflow/general-runtime/pull/597/files to be merge in and use the Unleash client it adds instead of hardcoding the feature flag here.

lib/services/runtime/handlers/utils/knowledgeBase/index.ts
```diff
@@ -50,6 +63,42 @@ export const getAnswerEndpoint = (cloudEnv: string, workspaceID: string): string
...
+export const fetchFaq = async (
+  projectID: string,
+  workspaceID: string | undefined,
+  question: string,
+  settings?: BaseModels.Project.KnowledgeBaseSettings
+): Promise<KnowledgeBaseFaq | null> => {
+  const cloudEnv = Config.CLOUD_ENV || '';
+  const flaggedWorkspaces = FAQ_FLAGGED_WORKSPACES_MAP.get(cloudEnv);
+
+  if (FAQ_RETRIEVAL_ENDPOINT && (flaggedWorkspaces?.length === 0 || flaggedWorkspaces?.includes(String(workspaceID)))) {
```
zhihil, line 75: Replace with a call to Unleash feature flagging.

lib/services/runtime/handlers/utils/knowledgeBase/index.ts
```diff
@@ -50,6 +63,42 @@ export const getAnswerEndpoint = (cloudEnv: string, workspaceID: string): string
...
+    });
+    return data?.faq;
+  }
+
+  return null;
+};
+
+const addFaqTrace = (runtime: Runtime, faq: KnowledgeBaseFaq, question: AIResponse) => {
+  runtime.trace.addTrace({
+    type: 'knowledgeBase',
```
zhihil, line 90: This needs to be defined as a new trace type.

The `TraceType` will need to be added here:
 https://github.com/voiceflow/libs/blob/c9c2430856bc5ce1c50d186280eab0218bdfeaf7/packages/base-types/src/node/utils/trace.ts#L3

The object schema as a file in `libs/packages/base-types/src/node/knowledgeBase.ts`, following the example of this:

https://github.com/voiceflow/libs/blob/master/packages/base-types/src/node/speak.ts

lib/services/runtime/handlers/utils/knowledgeBase/index.ts
```diff
@@ -50,6 +63,42 @@ export const getAnswerEndpoint = (cloudEnv: string, workspaceID: string): string
...
+    payload: {
+      faqQuestion: faq?.question,
+      faqAnswer: faq?.answer,
+      query: {
+        messages: question.messages,
+        output: question.output,
+      },
+    },
+  } as any);
```
zhihil, line 99: When the new trace type is defined, remove every `as any` cast in this file.

zhihil (changes requested):

lib/services/runtime/handlers/utils/knowledgeBase/index.ts
e-vandenberg: 🥳 finally unleash in general-runtime

lib/services/runtime/handlers/utils/knowledgeBase/index.ts
e-vandenberg: thanks for the references!

lib/services/runtime/handlers/utils/knowledgeBase/index.ts
```diff
@@ -33,6 +44,8 @@ const FLAGGED_WORKSPACES_MAP = new Map<string, string[]>([
 const { KL_RETRIEVER_SERVICE_HOST: host, KL_RETRIEVER_SERVICE_PORT: port } = Config;
 const scheme = process.env.NODE_ENV === 'e2e' ? 'https' : 'http';
 export const RETRIEVE_ENDPOINT = host && port ? new URL(`${scheme}://${host}:${port}/retrieve`).href : null;
+export const FAQ_RETRIEVAL_ENDPOINT =
+  host && port ? new URL(`${scheme}://${host}:${port}/poc/retrieve/faq`).href : null;
```
zhihil, line 48: Should remove the duplication of 
```ts
new URL(`${scheme}://${host}:${port}
```
between `RETRIEVE_ENDPOINT` and `FAQ_RETRIEVAL_ENDPOINT`

zhihil (approved):
Minor styling comment, but besides that approved:

PR #603, opened 2023-10-17
[Merged by Bors] - fix: openai client layout (CT-000)

DecathectZero:
Overall the business need is to be able to turn off models based on ENV VAR configuration.
For example if no Anthropic API Key is given, runtime should still work, but any calls to "anthropic" will fail.
Today the whole thing will crash if any LLM env vars are not set.

So ultimately what is happening here is we are moving `AIModelContext` as part of `CompletionOptions` which is only triggered on calls to `getCompletion` or `getChatCompletion`.

So instead of injecting the `{ workspaceID, projectID }` as part of the GET to the model, it is now being passed when we make the actual calls. I think this makes sense because the context is part of the call and relevant to moderation, but has nothing to do with the model itself.

On the models side, we are no longer instantiating a new client every time someone calls `services.ai.get(model)` - but rather all models are instantiated at initalization and there is a fixed pool.
The clients are now living inside the model instead of being decoupled.

### DIFF
--- a/lib/clients/ai/ai-model.interface.ts
+++ b/lib/clients/ai/ai-model.interface.ts
@@ -17,7 +17,7 @@ export interface AIModelContext {
   projectID?: string;
 }
 export interface CompletionOptions {
-  context?: AIModelContext;
+  context: AIModelContext;
   retries?: number;
   retryDelay?: number;
 }
--- a/lib/clients/ai/ai-model.interface.ts
+++ b/lib/clients/ai/ai-model.interface.ts
@@ -12,14 +12,14 @@ export interface CompletionOutput {
   answerTokens: number;
 }
 
-export interface CompletionOptions {
-  retries?: number;
-  retryDelay?: number;
-}
-
 export interface AIModelContext {
   workspaceID?: string;
   projectID?: string;
+}
+export interface CompletionOptions {
+  context?: AIModelContext;
+  retries?: number;
+  retryDelay?: number;
 }
 
 export const GPT4_ABLE_PLAN = new Set([
--- a/lib/services/aiSynthesis.ts
+++ b/lib/services/aiSynthesis.ts
@@ -91,10 +91,16 @@ class AISynthesis extends AbstractManager {
         },
       ];
 
-      response = await fetchChat({ ...options, messages }, generativeModel, variables, {
-        retries: this.DEFAULT_ANSWER_SYNTHESIS_RETRIES,
-        retryDelay: this.DEFAULT_ANSWER_SYNTHESIS_RETRY_DELAY_MS,
-      });
+      response = await fetchChat(
+        { ...options, messages },
+        generativeModel,
+        {
+          retries: this.DEFAULT_ANSWER_SYNTHESIS_RETRIES,
+          retryDelay: this.DEFAULT_ANSWER_SYNTHESIS_RETRY_DELAY_MS,
+          context,
+        },
+        variables
+      );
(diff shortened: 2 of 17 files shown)


lib/services/aiSynthesis.ts
```diff
@@ -101,13 +101,14 @@ class AISynthesis extends AbstractManager {
         <context>
```
pmvrmc, line 101: we need to add the context on line 94.

to look something like

```
response = await fetchChat({ ...options, messages }, generativeModel, variables, {
  context,
  retries: this.DEFAULT_ANSWER_SYNTHESIS_RETRIES,
  retryDelay: this.DEFAULT_ANSWER_SYNTHESIS_RETRY_DELAY_MS,
});
```

lib/clients/ai/ai-model.interface.ts
```diff
@@ -12,15 +12,15 @@ export interface CompletionOutput {
...
-  retries?: number;
-  retryDelay?: number;
-}
-
 export interface AIModelContext {
   workspaceID?: string;
   projectID?: string;
 }
+export interface CompletionOptions {
+  context?: AIModelContext;
```
pmvrmc, line 20: why make context optional? Ideally we should always pass the workspaceID/projectID

pmvrmc (changes requested):
we also need to change the `general-runtime/lib/services/runtime/handlers/utils/generativeNoMatch.ts` line 41 to pass the context down.

and on the same file, change line 37 to not pass the context to the `get` call.

pmvrmc:
> Overall the business need is to be able to turn off models based on ENV VAR configuration. For example if no Anthropic API Key is given, runtime should still work, but any calls to "anthropic" will fail. Today the whole thing will crash if any LLM env vars are not set.
> 
> So ultimately what is happening here is we are moving `AIModelContext` as part of `CompletionOptions` which is only triggered on calls to `getCompletion` or `getChatCompletion`.
> 
> So instead of injecting the `{ workspaceID, projectID }` as part of the GET to the model, it is now being passed when we make the actual calls. I think this makes sense because the context is part of the call and relevant to moderation, but has nothing to do with the model itself.
> 
> On the models side, we are no longer instantiating a new client every time someone calls `services.ai.get(model)` - but rather all models are instantiated at initalization and there is a fixed pool. The clients are now living inside the model instead of being decoupled.

The idea was to decouple the client, but only one instance of every client was ever created, when the model was intantiated it reused the client. In the old coupled (and also this PR) implementation there is one client created per model, so we have 3 anthropic clients and 3 openAI clients.

the model instantiation on the fly idea was just to create the model abstraction where we can inject context for that given model instance. It was just a plain object class, cheap and expandable. Moving the context to the model methods instead of the model itself is a good idea. My option there was that the model instance (since it is ephemeral) would hold the context so that uses of its methods would not need to, but that works to!

To fix the not passing ENV vars we could have tweaked the clients to return null if they are not present, and during the get check if the model instance client was null, if so log a warning.

But your approach is fine! just need to add those missing contexts and we can merge it 👍

lib/clients/ai/ai-model.interface.ts
DecathectZero: `CompletionOptions` itself is actually optional. I can make a bigger revision to make everything required

DecathectZero: making it required also means making `CompletionOptions` required, this also means we have to redo the order of a ton of other parameters, notable in fetchChat and fetchPrompt.

We can do it but it does increase the magnitude of the PR more

pmvrmc (approved):

PR #604, opened 2023-10-18
[Merged by Bors] - feat: using labels for KB query insted of tags (NLU-916)

abuyak-vf:

Changing the method of filtering by KB tags, instead of `ids` the client will send tags `labels`

When requesting a user to provide a filter by tags, the user will now provide the value of the tags in the `include` and `exclude` fields via `labels` rather than via `ids`.

### Related PRs

* https://github.com/voiceflow/creator-api/pull/1332

### DIFF
--- a/lib/services/test/index.ts
+++ b/lib/services/test/index.ts
@@ -1,27 +1,8 @@
-import { BaseModels } from '@voiceflow/base-types';
-
 import log from '@/logger';
 
 import { AbstractManager } from '../utils';
 
 export class TestService extends AbstractManager {
-  async tagNamesToObjectIds(
-    tags: string[],
-    existingTags?: Record<string, BaseModels.Project.KBTag>
-  ): Promise<Set<string>> {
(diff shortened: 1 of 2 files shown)


lib/services/test/index.ts
```diff
@@ -1,8 +1,27 @@
+import { BaseModels } from '@voiceflow/base-types';
+
 import log from '@/logger';
 
 import { AbstractManager } from '../utils';
 
 export class TestService extends AbstractManager {
+  async tagNamesToObjectIds(
+    tags: string[],
+    existingTags?: Record<string, BaseModels.Project.KBTag>
```
DecathectZero, line 10: make this take in a `tagLabelMap: Record<string, string>` where it is a dictionary of the label to the tagID. 
In `lib/controllers/test/index.ts` there can be a helper function that generates this dictionary.

DecathectZero (approved):

DecathectZero (approved):

PR #605, opened 2023-10-18
feat: use versioned KB if available (nlu-906)

e-vandenberg:
- Added a check to make sure that DM API key, projectID and versionID are consistent when passed
- default to version level KB settings if available


lib/middlewares/auth.ts
```diff
@@ -95,6 +95,35 @@ class Auth extends AbstractMiddleware {
 
     return this.verifyIdentity(req, res, next);
   }
+
+  verifyParamConsistency = (
```
theProf, line 99: few notes on this middleware: 

```
const verifyParamConsistency = (
  // if you want it flexible, is accepting a single function okay for you? it 
  // can have the contract of setting those params, but has less overhead 
  // than 3 simple funcs
  mapFromRequest: (
    req: Request,
  ) => { projectID: string; auth: string; versionID: string },
) => {
  return async (req: Request, _res: Response, next: Next) => {
    try {
      const { projectID, auth, versionID } = mapFromRequest(req);

      const api = await this.services.dataAPI.get();
      const authenticatedProject = auth ? await api.getProject(auth) : null;
      const version = versionID ? await api.getVersion(versionID) : null;

      // i find naming in the positive is more readable in usage. 
      // "isInConsistent" leads to possible double negatives when doing if (!isInConsistent)
      const isConsistent = [
        projectID,
        authenticatedProject?._id,
        version?.projectID,
      ].filter((item) => !!item) // return boolean for filter
        // removed length check because you will always have at least projectID(?)
        .every((item) => item == projectID); // since you know what you're comparing to, don't rely on order of the array. if ALL are the correct projectID also reads better

      if (!isConsistent) {
        // more specific error, caught and transformed in catch
        throw new VError("Inconsistent projectIDs");
      }
    } catch (err) {
      // is there a difference between `next(Error)` and just throwing?
      return next(new VError("Unauthorized", VError.HTTP_STATUS.UNAUTHORIZED));
    }

    // if we get here, all worked out
    next();
  };
};
```

```
middlewares.auth.verifyParamConsistency(
  (req: Request) => ({
    projectID: req.body.projectID,
    auth: req.headers.authorization,
    versionID: req.body.versionID,
  }),
);
```

theProf (changes requested):

lib/middlewares/auth.ts
e-vandenberg: I agree with most of these points, but I left the order based comparison because in theory this should also allow consistency checks for someone who supplies only DM API key and versionID, and doesn't supply projectID

theProf: i think I need a bit more clarity here, on what are the valid scenarios and what are not. 
```
u: undefined
  projectID | DM Key  | versionID | isConsistent
============|=========|===========|=============
      u     |    u    |     u     | true
      u     |    u    |     1     | true
*     u     |    1    |     1     | true
      u     |    1    |     u     | true
*     1     |    1    |     u     | true
*     1     |    1    |     1     | true
*     1     |    u    |     1     | true
      1     |    u    |     u     | true

* more than 1 ID exists, add cases for different
      u     |    1    |     2     | false
      u     |    2    |     1     | false
      1     |    2    |     u     | false
      2     |    1    |     u     | false
      1     |    1    |     2     | false
      1     |    2    |     1     | false
      1     |    2    |     2     | false
      2     |    2    |     1     | false
      2     |    1    |     2     | false
      2     |    1    |     1     | false
      1     |    u    |     2     | false
      2     |    u    |     1     | false
```

If the assumption is (considering the DM Key as an "ID"): "any combination of at least 1 ID(s) is valid as long as they are the same", then what you're actually looking at is a `Set` and can shorten down to:

```
      const projectIDs = new Set([
        projectID,
        authenticatedProject?._id,
        version?.projectID,
      ].filter((item) => !!item));

      // 1 and exactly 1 unique ID
      const isConsistent = projectIDs.size == 1;
```

I'm unsure of the typings of those 3 ids right now, but they'll need to be cast in the mapping to the Set if not the same

e-vandenberg: yep that's exactly it, I've implemented this solution, much cleaner 😀

I appreciate the help!

theProf (approved):

PR #606, opened 2023-10-18
fix: add openai path

DecathectZero:
Add a new (optional) `OPENAI_API_ENDPOINT` env var. If not set the client just uses the default OpenAI API so this is very low risk.

We can use the FF to disable moderation for a private cloud. Just moved it up higher. It's turned on for 100% of users right now.

### DIFF
--- a/lib/clients/ai/contentModeration/openai/openai.ts
+++ b/lib/clients/ai/contentModeration/openai/openai.ts
@@ -16,11 +16,14 @@ export class OpenAIModerationClient extends ContentModerationClient {
   constructor(config: Config, unleashClient: UnleashClient) {
     super(config, unleashClient);
     if (config.OPENAI_API_KEY) {
-      this.openAIClient = new OpenAIApi(new Configuration({ apiKey: config.OPENAI_API_KEY }));
+      this.openAIClient = new OpenAIApi(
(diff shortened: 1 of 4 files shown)


lib/clients/ai/contentModeration/openai/openai.ts
```diff
@@ -16,11 +16,14 @@ export class OpenAIModerationClient extends ContentModerationClient {
   constructor(config: Config, unleashClient: UnleashClient) {
     super(config, unleashClient);
     if (config.OPENAI_API_KEY) {
-      this.openAIClient = new OpenAIApi(new Configuration({ apiKey: config.OPENAI_API_KEY }));
+      this.openAIClient = new OpenAIApi(
+        new Configuration({ apiKey: config.OPENAI_API_KEY, basePath: config.OPENAI_API_ENDPOINT ?? undefined })
```
frankgu968, line 20: You mean `||` instead of `??`?

frankgu968 (changes requested):

lib/clients/ai/contentModeration/openai/openai.ts
DecathectZero: they do the same thing in this case, but yeah for consistency

pmvrmc (approved):

frankgu968 (approved):

PR #610, opened 2023-10-24
[Merged by Bors] - fix: Fix claude prompt format (VF-000) [bugfix]

e-vandenberg:
Claude requires new prompt format

Plus a bug fix where the wrong var was passed to the template causing context not to be available to the claude prompt.

Following these rules: 
https://docs.anthropic.com/claude/reference/prompt-validation#examples

### DIFF
--- a/lib/services/aiSynthesis.ts
+++ b/lib/services/aiSynthesis.ts
@@ -124,7 +124,7 @@ class AISynthesis extends AbstractManager {
       // This prompt scored 10% higher than the previous prompt on the squad2 benchmark
       const prompt = dedent`
       Reference Information:
-      ${context}
+      ${synthesisContext}
         
       If the question is not relevant to the provided information print("NOT_FOUND") and return.
       If the question is cannot be directly answered by a quote from the provided information print("NOT_FOUND") and return.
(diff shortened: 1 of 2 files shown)


lib/services/aiSynthesis.ts
```diff
@@ -187,31 +187,29 @@ class AISynthesis extends AbstractManager {
...
-      Primary instructions: ${prompt}`;
+      ${prompt}`;
     } else {
       content = dedent`
       Reference Information:
       ${knowledge}
-         
-      Secondary instructions:
-      Very concisely answer exactly how the reference information would answer this query.
+
```
e-vandenberg, line 205: prompt changes to avoid hallucination

lib/services/aiSynthesis.ts
```diff
@@ -124,7 +124,7 @@ class AISynthesis extends AbstractManager {
       // This prompt scored 10% higher than the previous prompt on the squad2 benchmark
       const prompt = dedent`
       Reference Information:
-      ${context}
+      ${synthesisContext}
```
e-vandenberg, line 127: was previously using wrong var

Diophontine (approved):

PR #619, opened 2023-10-31
[Merged by Bors] - feat: handle prompt leak (VF-000)

e-vandenberg:
Some users experience terms from the prompt being output by the LLM.

The goals of this PR are to:
1. log when this happens to better understand things like why, how often, etc.
2. retry the LLM call if this happens in hopes of giving the user a better experience while we try to improve the underlying system.

### DIFF
--- a/lib/services/aiSynthesis.ts
+++ b/lib/services/aiSynthesis.ts
@@ -40,7 +40,7 @@ class AISynthesis extends AbstractManager {
     /Instructions/,
     /user:/,
     /assistant:/,
-    /<.*>/,
+    /<[^<>]*>/,
   ];
 
   private readonly MAX_LLM_TRIES = 2;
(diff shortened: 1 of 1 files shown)


lib/services/aiSynthesis.ts
```diff
@@ -34,6 +34,17 @@ class AISynthesis extends AbstractManager {
 
   private readonly DEFAULT_QUESTION_SYNTHESIS_RETRIES = 2;
 
+  private readonly REGEX_PROMPT_TERMS = [
+    /conversation_history/i,
+    /knowledge/i,
+    /instructions/i,
```
zhihil, line 40: `knowledge` and `instructions` might flag a lot of false positives. Can we try to make this more specific somehow?

zhihil (approved):

PR #623, opened 2023-11-08
fix: allow start step to process launch request (CT-000)

DecathectZero:
I apologize for this piece of somewhat convoluted logic. We need to set the request to the builtin `last_event`, so we don't want to set it to `null`.

The center of the issue is here:
https://github.com/voiceflow/general-runtime/blob/78a10d27ad3d0d2e34f7e939779d35ec242c2bba/runtime/lib/Runtime/index.ts#L224-L229

We used to manually transform any launch requests into `null` requests. But this preliminary handler would basically swallow all events and stop on the same node, so it looks like nothing happens.

What this does is allow the start step to handle events and continue, if there are no global commands. 

The side effect of this behavior, is the on the DMAPI the user could literally send ANY event (that has no global command match) and it would trigger the start.  The previous behavior would've been absolutely nothing happens, and it keeps listening.

### DIFF
--- a/lib/services/runtime/handlers/state/preliminary/index.ts
+++ b/lib/services/runtime/handlers/state/preliminary/index.ts
@@ -1,6 +1,6 @@
-import { VoiceflowNode } from '@voiceflow/voiceflow-types';
+import { BaseNode } from '@voiceflow/base-types';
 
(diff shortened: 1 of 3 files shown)


DecathectZero:
closed in favor of https://github.com/voiceflow/general-runtime/pull/624

PR #625, opened 2023-11-09
[Merged by Bors] - fix: new gpt35 token multipliers (CT-000)

DecathectZero:


lib/clients/ai/openai/gpt.ts
```diff
@@ -24,6 +24,6 @@ export abstract class GPTAIModel extends AIModel {
   }
 
   protected calculateTokenMultiplier(tokens: number): number {
-    return Math.floor(tokens * this.TOKEN_MULTIPLIER);
+    return Math.ceil(tokens * this.TOKEN_MULTIPLIER);
```
DecathectZero, line 27: we should `ceil`

Diophontine (approved):

PR #640, opened 2023-12-06
[Merged by Bors] - feat: pass intent & entity scopes (VF-000)

e-vandenberg:
Determine intent/entity scope and pass it to nlu-gateway via:
- `filteredIntents: string[]`
- `filteredEntities: string[]`
- `excludeFilteredIntents: boolean`
- `excludeFilteredEntities: boolean`

The scope is defined by the intersection of:
- "intent scoping" at the `Listen` step level (`Buttons`, `Choice`, `Capture`) of the current node
- reachable intents determined by all commands in all frames on the runtime stack. ("Available from other topics" toggle on intent block)

Changes are backwards compatible and have no impact on functionality at this point. 

For context, these parameters will be passed to NLU user inference in order to better calculate accuracies based on a refined intent/entity scope rather than the full set of intents/entities.

### DIFF
--- a/lib/services/nlu/utils.ts
+++ b/lib/services/nlu/utils.ts
@@ -5,9 +5,11 @@ import { GoogleConstants } from '@voiceflow/google-types';
 import { VoiceflowConstants } from '@voiceflow/voiceflow-types';
 import { match } from 'ts-pattern';
 
-import Runtime from '@/runtime/lib/Runtime';
+import { Stack } from '@/runtime';
+import { Context } from '@/types';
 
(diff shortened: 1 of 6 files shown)


lib/services/nlu/utils.ts
```diff
@@ -52,3 +57,51 @@ export const mapChannelData = (data: any, platform?: VoiceflowConstants.Platform
     },
   };
 };
+
+export const setIntersect = (set1: Set<any>, set2: Set<any>) => new Set([...set1].filter((i) => set2.has(i)));
```
pmvrmc, line 61: this doenst need to be exported

pmvrmc (approved):

PR #643, opened 2023-12-12
[Merged by Bors] - feat: replace ttypescript with tsc-alias (PL-000)

trs:
`ttypescript` was used to enable the `typescript-transform-paths` transformer, but `ttypescript` isn't maintained. This uses `typescript` directly with `tsc-alias` for the path alias transformations.


theProf (approved):

trs:
bors merge

PR #657, opened 2023-12-13
fix: remove vm2 dependency (PL-000)

trs:
`vm2` is deprecated and no longer supported. It was used in our original `if` and `set` steps, but those no longer exist. Removes `vm2` and all surrounding calls.

image

Fixes:
- https://github.com/voiceflow/general-runtime/security/dependabot/80
- https://github.com/voiceflow/general-runtime/security/dependabot/81


trs:
https://github.com/voiceflow/general-runtime/pull/704

PR #661, opened 2023-12-14
[Merged by Bors] - fix: error handling and variable assignment (CV3-000)

zhihil:
**Fixes or implements CV3-000**

### Related PRs

### Checklist

### DIFF
--- a/runtime/lib/HTTPClient/function-lambda/index.ts
+++ b/runtime/lib/HTTPClient/function-lambda/index.ts
@@ -1,3 +1,4 @@
+import { HTTP_STATUS } from '@voiceflow/verror';
 import AWS from 'aws-sdk';
 
 import { FunctionLambdaRequest } from './interface';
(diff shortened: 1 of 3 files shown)


runtime/lib/HTTPClient/function-lambda/index.ts
```diff
@@ -46,6 +47,11 @@ export class FunctionLambdaClient {
         } else {
           const parsedPayload = JSON.parse(data.Payload as string);
           const responseBody = parsedPayload.body;
+
+          if (parsedPayload.statusCode !== HTTP_STATUS.OK) {
+            reject(responseBody);
+          }
+
```
zhihil, line 54: Fixes an issue where the `function-lambda` is sending back an error response, but it was not being treated as such by the `general-runtime`.

theProf (commented):
single change request. not knowledgable enough on this repo to accept the other changes

runtime/lib/HTTPClient/function-lambda/index.ts
zhihil: I've added an `else` branch to wrap the `resolve`. It doesn't seem to matter, maybe because the first call from one of `resolve`, `reject` counts? Nonetheless, I've added the `else` to ensure only one is called at a time.

It would be nicer if `aws-sdk` provided Promises instead of using a callback interface...

JoshTheDeveloper (approved):

PR #669, opened 2024-01-03
feat: remove memory and question synthesis in kb fallback (BUG-688)

trs:
Remove the question synthesis and conversation memory from the knowledge base no match. This makes the no match closer to querying the KB directly, and removes weird responses due to using the history.


trs:
Replaced by: https://github.com/voiceflow/general-runtime/pull/670

PR #670, opened 2024-01-03
feat: remove memory and question synthesis in kb fallback (BUG-688)

trs:
Remove the question synthesis and conversation memory from the knowledge base no match. This makes the no match closer to querying the KB directly, and removes weird responses due to using the history.

### DIFF
--- a/lib/services/runtime/handlers/utils/knowledgeBase/index.ts
+++ b/lib/services/runtime/handlers/utils/knowledgeBase/index.ts
@@ -174,46 +174,39 @@ export const knowledgeBaseNoMatch = async (runtime: Runtime): Promise<AIResponse
   if (!input) return null;
 
   try {
-    // expiremental module, frame the question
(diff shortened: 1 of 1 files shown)


lib/services/runtime/handlers/utils/knowledgeBase/index.ts
```diff
@@ -174,46 +174,39 @@ export const knowledgeBaseNoMatch = async (runtime: Runtime): Promise<AIResponse
...
+      question,
       runtime.project?.knowledgeBase?.faqSets,
       kbSettings
     );
     if (faq?.answer) {
-      addFaqTrace(runtime, faq.question || '', faq.answer, question.output);
+      addFaqTrace(runtime, faq.question || '', faq.answer, question);
       return {
-        model: question.model,
+        model: 'faq',
```
trs, line 196: Unsure what to put here.

PR #678, opened 2024-01-09
[Merged by Bors] - feat: add speak trace, audio trace, URL buttons to carousel and card (CV3-794)

zhihil:
**Fixes or implements CV3-794**

Adding adapters to support simplified speak trace, audio trace, and adding URL buttons to the carousel & card. 

Add adapters to convert simplified speak traces, audio traces, and URL buttons to their correct format that is expected by the rest of our codebase.

### Setup information

N/A

### Deployment Notes

N/A

### Related PRs

N/A

### Checklist

### DIFF
--- a/runtime/lib/Handlers/function/lib/execute-function/lib/adapt-trace.ts
+++ b/runtime/lib/Handlers/function/lib/execute-function/lib/adapt-trace.ts
@@ -1,19 +1,29 @@
-import { BaseNode, BaseTrace } from '@voiceflow/base-types';
-import cuid from 'cuid';
+import { BaseNode, BaseRequest, BaseTrace } from '@voiceflow/base-types';
+import { Utils } from '@voiceflow/common';
 
 import {
+  SimpleAction,
+  SimpleAudioTrace,
+  SimpleButton,
+  SimpleCard,
   SimpleCardV2Trace,
   SimpleCarouselTrace,
+  SimpleSpeakTrace,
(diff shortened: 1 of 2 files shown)


runtime/lib/Handlers/function/lib/execute-function/lib/adapt-trace.ts
```diff
@@ -1,19 +1,29 @@
...
   SimpleTextTrace,
   SimpleTrace,
   SimpleTraceDTO,
+  SimpleTraceType,
   SimpleVisualTrace,
   Trace,
 } from '../../../runtime-command/trace-command.dto';
 
+const { cuid } = Utils.id;
+const { TraceType } = BaseTrace;
```
zhihil, line 21: This weird assignment is needed because the prototype tool fails locally at the function step complaining that `cuid_1` or `base_types_1` does not exist. 

I'm wondering if there's some kind of subtle build issue, e.g, TypeScript is compiling out the exports from `Utils` and `BaseTrace` which is only detected when it's executed.

trs: Typescript doesn't do any tree shaking or anything like that, so not sure it would be a TS issue. What were you attempting before? Something like:

```ts
import { Utils } from '@voiceflow/common';

...

Utils.id.cuid()
```

trs (approved):

PR #659, opened 2023-12-14
[Merged by Bors] - build: volta, node v20, yarn3 (PL-668)

theProf:
**Fixes or implements PL-668**

### Brief description. What is this change?

upgrade repo to use volta, yarn v3, node v20

### Deployment Notes

Updated circleci to use node v20 in tests and build contexts

### Related PRs

- https://github.com/voiceflow/general-runtime/pull/645

### Checklist
- [x] manual testing
- [x] run automated test suite now that it exists

PR #682, opened 2024-01-13
build: node v20 yarn v3 with ci updates (PL-668)

theProf:
**Fixes or implements PL-668**

Re-introduce node v20, yarn v3, volta changes

Don't use CircleCI yarn cache as it is build with node v16. Do all building in Dockerfile and update orb-common down the road

### Deployment Notes

CircleCI install_and_build will not be useful troubleshooting wise as it is will be using node v16 and will not have it's artifacts used for production image

### Reviewer Notes

Focus on the latest commits, which target Dockerfile(.e2e) and a slight change to package.json. these are the only 3 files changed from initial PR 659

### Related PRs

- https://github.com/voiceflow/general-runtime/pull/659
- https://github.com/voiceflow/general-runtime/commit/bc95625393a1cf3ddad6bd75a26c9ea35f5e7265
- https://github.com/voiceflow/orb-common/pull/187

### Checklist

- [ ] builds in pipelines
- [x] tracked in dev env
- [x] automated smoke tests


Dockerfile
```diff
@@ -15,15 +15,29 @@ ENV GIT_SHA=${build_GIT_SHA}
...
 
 WORKDIR /usr/src/app
-COPY build ./
-COPY package.json ./
-COPY yarn.lock ./
 
+FROM base as builder
+ARG NPM_TOKEN
+
+COPY . .
```
theProf, line 22: yuck. full copy. Didn't want to specify every dotfile for building, and wasn't sure how .dockerignore would be taken into consideration with the DinD/remote docker building from a docker container situation. so went the simplest route

Dockerfile
```diff
@@ -15,15 +15,29 @@ ENV GIT_SHA=${build_GIT_SHA}
...
+  yarn install --immutable && \
+  yarn build && \
+  yarn config unset -H npmRegistries
+
+FROM base as prod
+ARG NPM_TOKEN
+
+COPY --from=builder /usr/src/app/build .
+COPY --from=builder /usr/src/app/.yarn ./.yarn
+COPY --from=builder /usr/src/app/.yarnrc.yml .
```
theProf, line 35: need to copy the `.yarnrc.yml` and `.yarn/plugins` for the workspace plugin to do a `yarn workspaces focus`. These can be removed when going to yarn v4 where it is a builtin command that doesn't need a plugin

Dockerfile
```diff
@@ -15,15 +15,29 @@ ENV GIT_SHA=${build_GIT_SHA}
...
+  rm -rf ./node_modules/ .yarn/cache/ .yarn/install-state.gz ./build/ && \
+  yarn install --immutable && \
+  yarn build && \
+  yarn config unset -H npmRegistries
+
+FROM base as prod
+ARG NPM_TOKEN
+
+COPY --from=builder /usr/src/app/build .
+COPY --from=builder /usr/src/app/.yarn ./.yarn
```
theProf, line 34: transfers the `.yarn/cache` for re-use but decided against copying `node_modules` for more certainty about the build outcome, at the cost of a slightly longer build time (tops 2m addition, closer to 1m)

theProf:
bors try

Dockerfile
```diff
@@ -15,15 +15,29 @@ ENV GIT_SHA=${build_GIT_SHA}
...
-COPY package.json ./
-COPY yarn.lock ./
 
+FROM base as builder
+ARG NPM_TOKEN
+
+COPY . .
+
+RUN yarn config set -H 'npmRegistries["https://registry.yarnpkg.com"].npmAuthToken' "${NPM_TOKEN#"//registry.npmjs.org/:_authToken="}" && \
+  rm -rf ./node_modules/ .yarn/cache/ .yarn/install-state.gz ./build/ && \
```
xavidop, line 25: why dont we add this to the .gitignore file?

Dockerfile
```diff
@@ -15,15 +15,29 @@ ENV GIT_SHA=${build_GIT_SHA}
...
+  yarn install --immutable && \
+  yarn build && \
```
xavidop, line 27: we should re-use the build, if nor we are building twice

xavidop (changes requested):
Looks quite good! happy to see these changes!! just 2 small considerations

Dockerfile
theProf: -  `.yarn/cache` will exist from `vf_<save|restore>_cache` from `install_node_modules`
- `build/` will exist on disk from `monorepo_<save|restore>_cache` within `install_and_build`
- the others are probably superfluous

Dockerfile
theProf: to re-use the node deps (`node_modules + .yarn/cache`, we need a build from `alpine`. all convenience images provided by CircleCI have an [ubuntu base](https://circleci.com/developer/images/image/cimg/base). we will either need to roll our own `alpine` base/node image to use as an executor or extend the current flow to more easily specify the container image to run yarn commands in from `install_and_build`.

For the `build/` portion, where is it built twice? `update_track` and `build_push_image` (Dockerfile/Dockerfile.e2e builds)? 
I think I initially wanted a clean test build, but you're right, since we are only using `./build/` from the Dockerfile step, it can be re-used. I'll add this back in

xavidop: @theProf you are building the services in the install_and_build step, and then again in the Dockerfile. 
I do not fully understand why we need to build from alpine. I mean, why the linux distro is relevant for the .yarn/cache

Dockerfile
xavidop: yeah, but if we are going to copy them and then remove them, why don't we add them to the .gitignore file so we do not copy them?

Dockerfile
theProf: there are node packages that are dependencies that are not pre-built. they are built during the `yarn install` phase, which is run on an ubuntu executor
The Dockerfile is built from an `alpine` base, which does not share compatible linked libraries with `ubuntu`. 

The Dockerfile is also staged for small prod footprint, which does not have a `./build` dir to be copied out of after the build command, so no dir to be cached. We could add a pre- or post- build step that does a `docker build --target builder` and then copies the build dir out of there though

theProf:
bors try

PR #684, opened 2024-01-23
[Merged by Bors] - feat: button label sets last utterance (BUG-710)

trs:
If a button is clicked, use the label as the `last_utterance` variable.

Previously:
image

Now:
image


pmvrmc (approved):

trs:
bors merge

PR #694, opened 2024-02-07
feat: initializing variables from cms (CV3-942)

zhihil:
**Fixes or implements CV3-942**

Change the `general-runtime` so that we initialize the default values based on the CMS variable declarations. 

`general-service` will be changed so that compiled variable declaration information is stored in `version.prototype.surveyorContext.cmsVariables`. This includes the user-set default values for each variable.

At the `initializeVariables()` method of `general-runtime`, we add a new line called `initializeFromCMSVariables()` which will  create an overwrite patch to set each variable to its CMS default value.

### Setup information

N/A

### Deployment Notes

N/A

### Related PRs

- https://github.com/voiceflow/general-runtime/pull/694
- https://github.com/voiceflow/general-service/pull/558
- https://github.com/voiceflow/libs/pull/501
- https://github.com/voiceflow/creator-app/pull/7761

### Checklist

### DIFF
--- a/lib/services/state/index.ts
+++ b/lib/services/state/index.ts
@@ -87,7 +87,7 @@ class StateManager extends AbstractManager<{ utils: typeof utils }> implements I
 
   initializeFromCMSVariables(variables: Record<string, any>) {
     return Object.fromEntries(
-      Object.entries(variables).reduce(([name, declare]) => [name, this.parseDefaultValue(name, declare) ?? 0])
+      Object.entries(variables).map(([name, declare]) => [name, this.parseDefaultValue(name, declare) ?? 0])
     );
   }
 
--- a/lib/services/state/index.ts
+++ b/lib/services/state/index.ts
@@ -57,11 +57,11 @@ class StateManager extends AbstractManager<{ utils: typeof utils }> implements I
   }
 
   parseDefaultValue(variableName: string, declare: any) {
-    const { defaultValue } = declare;
+    const { defaultValue, isSystem, datatype } = declare;
 
-    let type = declare.datatype;
+    let type = datatype;
 
-    if (declare.isSystem) {
+    if (isSystem) {
       const builtInType = builtInVariableTypes.get(variableName);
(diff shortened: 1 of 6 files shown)


lib/services/state/index.ts
```diff
@@ -54,15 +56,56 @@ class StateManager extends AbstractManager<{ utils: typeof utils }> implements I
...
+        [name]: this.parseDefaultValue(name, declare) ?? 0,
+      }),
+      {}
+    );
+  }
+
   // initialize all entities and variables to 0, it is important that they are defined
   initializeVariables(version: BaseModels.Version.Model<any>, state: State) {
     const entities = version.prototype?.model.slots.map(({ name }) => name) || [];
+    const variables: Record<string, Variable> = version.prototype?.surveyorContext.cmsVariables ?? {};
```
zhihil, line 101: If the `general-service` changes aren't merged, then `.cmsVariables` is not generated by the compiler. Then, we fallback to an empty object which causes all of this code to do nothing and be backwards compatible.

lib/services/state/index.ts
```diff
@@ -54,15 +56,56 @@ class StateManager extends AbstractManager<{ utils: typeof utils }> implements I
...
+
+  initializeFromCMSVariables(variables: Record<string, any>) {
+    return Object.entries(variables).reduce<Record<string, string | number | boolean | Date>>(
+      (acc, [name, declare]) => ({
+        ...acc,
+        [name]: this.parseDefaultValue(name, declare) ?? 0,
+      }),
+      {}
+    );
+  }
```
z4o4z, line 96: better to use `Object.fromEntrie` to generate records, looks cleaner:
```suggestion
    return Object.fromEntries(Object.entries(variables).reduce<Record<string, string | number | boolean | Date>>(
      ([name, declare]) => [name, this.parseDefaultValue(name, declare) ?? 0]
    ));
  }
```

z4o4z (approved):

PR #702, opened 2024-02-22
fix: payload type is wrong, can be undefined (PL-752) [bugfix]

trs:
Caused by: https://github.com/voiceflow/general-runtime/pull/698

The `isPathRequest` guard did not assert a payload label existed, so when using that in the ai memory variable the types indicate that payload will be there.

image

So accessing `payload.label` resulted in a "Cannot read payload from undefined" error.

This change adds a `payload.label` check to guard itself, as the typing for `payload` is wrong, and could be null or undefined.

image

See: https://github.com/voiceflow/general-runtime/pull/690

### DIFF
--- a/lib/services/runtime/index.ts
+++ b/lib/services/runtime/index.ts
@@ -63,7 +63,7 @@ class RuntimeManager extends AbstractManager<{ utils: typeof utils }> implements
       }
     }
 
-    if (isPathRequest(request) && typeof request.payload?.label === 'string') {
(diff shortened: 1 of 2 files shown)


trs:
Ticket to fix the types here: https://linear.app/voiceflow/issue/PL-753/update-voiceflowbase-types-to-handle-null-or-undefined-payload

theProf (approved):

trs:
bors merge

PR #703, opened 2024-02-23
feat: enable kb on usaa (IN-1186)

edison-vflow:
### Fixes or implements 
* https://voiceflow.atlassian.net/browse/IN-1186

### Checklist

- [x] Similar change tested on private clouds


bhuvana-ak (approved):
lgtm!

PR #704, opened 2024-02-23
[Merged by Bors] - feat: run vm2 and isolated-vm in parallel (PL-754)

trs:
Related: https://github.com/voiceflow/general-runtime/pull/657

Run the remote [`code-block`](https://github.com/voiceflow/code-block) code executor alongside the `isolated-vm` executor in parallel. This will let us track any breaking changes that may occur when removing the remove `code-block` executor.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -27,7 +27,27 @@
       "sourceMaps": true,
       "pauseForSourceMap": true,
       "smartStep": true,
-      "console": "integratedTerminal"
+      "console": "integratedTerminal",
+      "killBehavior": "polite"
+    },
+
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "test:file",
+      "skipFiles": ["<node_internals>/**"],
+      "runtimeExecutable": "yarn",
+      "runtimeArgs": [
+        "test:single",
+        "${relativeFile}",
+        "--watch",
+        "--jobs 1"
+      ],
+      "sourceMaps": true,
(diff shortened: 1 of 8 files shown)


DecathectZero (changes requested):
Is there any chance that ivmExecute crashing or timing out might disrupt the current expected behavior with VM2? The goal is to fully shadow it but not alter any existing expected behavior.

### Main change request:
For new runtime changes **to existing flows** we might want to write a few tests, especially around edge cases of code crashing/looping/timing out etc.

Other notes:
It will also be useful to log the projectIDs of any issues. This way we can collect them and communicate if we do dedicate to go forward with any changes.
In the future we should figure out a better system for gathering this data instead of datadog logs, its fine here but it should go into a proper analytics engine in the future so we can make business/product decisions based on what we find.

DecathectZero (approved):

DecathectZero:
I've rebased everything and made a small change to how we do logging. This is just a learning I have from our logger package from debugging the realtime service.

Datadog reads exclusively the JSON format, not the inline format.
The first parameter is actually the general object, and the second is the message. If you made the first parameter the message, that is also fine.
But look at the JSON before and after my change:
BEFORE:
Screenshot 2024-03-23 at 1 43 02 PM
AFTER:
Screenshot 2024-03-23 at 1 42 52 PM

What is in the general object appears in Datadog in the body of the object:
Screenshot 2024-03-23 at 1 46 47 PM
Which makes it far, far, far more easy to read and debug.

PR #708, opened 2024-03-08
[Merged by Bors] - fix: simplified buttons are action requests (PL-839)

trs:
Carousel/card buttons created from a function can only open URLs, and should not send a request to the runtime. If a request is sent, we ensure it is of type `action`, so we do not execute the runtime.

See: https://github.com/voiceflow/general-runtime/pull/678

### DIFF
--- a/runtime/lib/Handlers/function/lib/execute-function/lib/adapt-trace.ts
+++ b/runtime/lib/Handlers/function/lib/execute-function/lib/adapt-trace.ts
@@ -90,7 +90,7 @@ const adaptButton = (button: SimpleButton): BaseRequest.GeneralRequestButton => 
   return {
     name: button.name,
     request: {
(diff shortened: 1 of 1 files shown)


runtime/lib/Handlers/function/lib/execute-function/lib/adapt-trace.ts
```diff
@@ -90,7 +90,7 @@ const adaptButton = (button: SimpleButton): BaseRequest.GeneralRequestButton =>
   return {
     name: button.name,
     request: {
-      type: `function-request-${button.name}`,
+      type: 'action',
```
pmvrmc, line 93: can we use the one defined on base-types?

pmvrmc (approved):

trs:
bors merge

PR #711, opened 2024-03-08
fix: write heapdump on sigusr2 (BUG-1167)

DecathectZero:
Tested locally, can try on staging environment before deployed to prod
Screenshot 2024-03-08 at 12 03 32 PM

### DIFF
--- a/start.ts
+++ b/start.ts
@@ -2,6 +2,9 @@ import 'core-js';
 import 'regenerator-runtime/runtime';
 import './tracer';
 
+import fs from 'fs';
+import v8 from 'v8';
+
 import { ServiceManager } from './backend';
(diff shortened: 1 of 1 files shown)


pmvrmc (approved):

theProf (approved):
test thoroughly in an env closer to production before pushing to production

DecathectZero:
FYI this PR is not needed at all. We can just send a `kill -SIGUSR1 <node PID>` and get it into inspect mode, and port forward with `kubectl port-forward`

https://betterprogramming.pub/hunting-for-a-memory-leak-in-a-node-js-application-on-kubernetes-why-heaps-dont-lie-e864c1d8a522

going to close it

PR #715, opened 2024-03-12
[Merged by Bors] - fix: update logger (BUG-1172)

DecathectZero:
I played around a lot with the logger, and I think this is the minimal risk one to solve the memory issue.
This doesn't change anything with `backend-utils` and ensures the same logging behavior with `vars`:
https://github.com/voiceflow/logger/blob/3f4220c97ffa69351a3058d2000fa55547d90a91/lib/logger.ts#L70-L89

added pino as a dev dependency PURELY for types, because `@voiceflow/backend-utils` uses a different version of pino and pino http.

The longer term update is over here:
https://github.com/voiceflow/general-runtime/pull/692#pullrequestreview-1930249904


package.json
```diff
@@ -119,6 +119,7 @@
     "lint-staged": "^10.5.3",
     "mocha": "^6.2.3",
     "nyc": "^15.1.0",
+    "pino": "8.17.2",
```
DecathectZero, line 122: this version of pino is added PURELY for types, the latest pino is already in typescript and `@types/pino` is no longer necessary.

xavidop (approved):

pmvrmc (approved):

PR #717, opened 2024-03-14
feat: set value to vf-chunks build-in variable (COR-1187)

abuyak-vf:

Update value of new build-in variable _**VF_CHUNKS**_ after KB search query


DecathectZero:
My main concern is that the chunks aren't too small, and we're going to permenantly going to save it to the variable (there's no way to remove or reset it)

Would it make sense you could only access on the same turn? At the end of the turn before we save back into the DB we scrub it again. This way it only lives in the memory of the turn.

lib/services/state/index.ts
```diff
@@ -1,5 +1,5 @@
 import { BaseModels, BaseRequest, BaseTrace } from '@voiceflow/base-types';
-import { Variable } from '@voiceflow/dtos';
+import { CompiledCMSVariable } from '@voiceflow/base-types/build/cjs/cms/variables';
```
z4o4z, line 2: why did we change this?

z4o4z (approved):

lib/services/state/index.ts
abuyak-vf: We updated base-types package and now _**cmsVariables**_ is _**Record<string, CompiledCMSVariable>**_ type in _**SurveyContext**_ type. I thinks before it was a _**AnyRecord**_ type , that is why we change and no need import _**Variable**_ anymore:

_const variables: Record<string, Variable> = version.prototype?.surveyorContext.cmsVariables ?? {};_
->
_const variables: Record<string, CompiledCMSVariable> = version.prototype?.surveyorContext.cmsVariables ?? {};_

image

PR #733, opened 2024-03-25
[Merged by Bors] - feat: validate trace payloads to avoid transcript issue (COR-000)

zhihil:
**Fixes or implements COR-000**

Transcript UI is failing because users are able to send debug traces from the Function step without `payload.message` defined. 

We will add validation on the Function step to throw an exception if the user provides a standard trace (e.g. debug) and does not fit the type definition.

Added a trace validation utility to ensure that if users are providing traces of a standard type, then their contents conforms to the type definition. If the trace has property `field`, then

1. If the `field` is a built-in property of the interface, then `field`'s value must match the type on the interface
2. If the `field` is not a built-in property of the interface, then `field`'s value is not type-checked and the value is simply "passed through"

This approach means that users have the flexibility to define fields that we "don't care about", but must satisfy type safety for when we do.

### Setup information

N/A

### Deployment Notes

N/A

### Related PRs

### Checklist

### DIFF
--- a/runtime/lib/Handlers/function/lib/execute-function/execute-function.ts
+++ b/runtime/lib/Handlers/function/lib/execute-function/execute-function.ts
@@ -34,9 +34,6 @@ export async function executeFunction(funcData: ExecuteFunctionArgs) {
   return {
     next,
     outputVars,
-    trace: trace?.map((tr) => {
-      validateTrace(tr);
-      return adaptTrace(tr);
-    }),
+    trace: trace?.map((tr) => adaptTrace(validateTrace(tr))),
   };
 }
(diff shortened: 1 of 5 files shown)


runtime/lib/Handlers/function/lib/execute-function/execute-function.ts
```diff
@@ -33,6 +34,9 @@ export async function executeFunction(funcData: ExecuteFunctionArgs) {
   return {
     next,
     outputVars,
-    trace: trace?.map((tr) => adaptTrace(tr)),
+    trace: trace?.map((tr) => {
+      validateTrace(tr);
```
trs, line 38: If you return from `validateTrace` you can pass the result to adapt trace.

```ts
return adaptTrace(validateTrace(tr));
```

Renaming to `parseTrace` in that case might make more sense, too?

trs (approved):

PR #739, opened 2024-04-01
[Merged by Bors] - feat: executing new compiled function nodes (COR-1757)

zhihil:
**Fixes or implements COR-1757**

Update the `general-runtime` to account for two kinds of compiled function nodes:

1. Legacy compiled nodes - Function definition is embedded inside of the node itself.
2. New compiled nodes - Function definition is stored in a mapping inside `version.prototype`. The compiled function node contains a `functionId` property which can be used to look up the definition. 

The function handler now calls a `resolveFunctionDefinition()` method before executing its main logic. This function detects if the incoming node is a legacy or new compiled node.

If it's a legacy node, we extract the definition out from the compile node.

If it's a new node, we lookup the definition in `version.prototype` and return it.

Either way, we return the function definition referenced by the function step, and execute the same logic as we have now. 

### Deployment Notes

The changes must be merged in this order:
1. `general-runtime`
2. `general-service`

This PR is designed to be backwards compatible, so it will not break users' existing projects.

### Related PRs

- https://github.com/voiceflow/libs/pull/512
- https://github.com/voiceflow/creator-app/pull/7919
- https://github.com/voiceflow/general-service/pull/566
- https://github.com/voiceflow/general-runtime/pull/739

### Checklist


runtime/lib/Handlers/function/index.ts
```diff
@@ -52,12 +53,35 @@ function applyNextCommand(command: NextCommand, paths: FunctionCompiledInvocatio
   return null;
 }
 
+function resolveFunctionDefinition(
+  definition: FunctionCompiledData['definition'],
+  version: BaseVersion.Version
+): FunctionCompiledDefinition {
+  if ('functionId' in definition) {
```
trs, line 60: This check is to ensure we support the old style, right?

trs (approved):

runtime/lib/Handlers/function/index.ts
zhihil: Yes, this `if` statement will handle the new style. Otherwise, we just return `definition` because it's an older style.

PR #754, opened 2024-04-04
fix: fix settings (PL-844)

DecathectZero:
the `version.settings` is strictly optional.
https://github.com/voiceflow/creator-app/blob/cfc893c267d5a9562973fceb01b3f8821668c504/libs/dtos/src/version/version.dto.ts#L33

All existing projects don't have it. 
And the migrations only target projects with existing projects.

The very first test I ran basically failed, so a bit of an oversight there.

I intend to just merge this into [tyler/test-classification-endpoint/PL-844](https://github.com/voiceflow/general-runtime/tree/tyler/test-classification-endpoint/PL-844)

### DIFF
--- a/lib/services/classification/classification.const.ts
+++ b/lib/services/classification/classification.const.ts
@@ -1,0 +1,24 @@
+import { DEFAULT_INTENT_CLASSIFICATION_PROMPT_WRAPPER_CODE } from '@voiceflow/default-prompt-wrappers';
+import {
+  AIModel,
+  IntentClassificationLLMSettings,
+  IntentClassificationNLUSettings,
+  IntentClassificationType,
+} from '@voiceflow/dtos';
+
+export const DEFAULT_NLU_INTENT_CLASSIFICATION: IntentClassificationNLUSettings = {
+  type: IntentClassificationType.NLU,
+  params: { confidence: 0.6 },
+};
+
+// can be removed after migration PL-846
+export const LEGACY_LLM_INTENT_CLASSIFICATION: IntentClassificationLLMSettings = {
+  type: IntentClassificationType.LLM,
+  params: {
+    model: AIModel.GPT_4_TURBO,
+    temperature: 0.1,
+  },
+  promptWrapper: {
+    content: DEFAULT_INTENT_CLASSIFICATION_PROMPT_WRAPPER_CODE,
+  },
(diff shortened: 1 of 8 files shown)


lib/services/classification/classification.const.ts
```diff
@@ -0,0 +1,23 @@
...
+import {
+  AIModel,
+  IntentClassificationLLMSettings,
+  IntentClassificationNLUSettings,
+  IntentClassificationType,
+} from '@voiceflow/dtos';
+
+export const DEFAULT_NLU_INTENT_CLASSIFICATION: IntentClassificationNLUSettings = {
+  type: IntentClassificationType.NLU,
+  params: { confidence: 60 },
```
DecathectZero, line 11: 60 was the original default value
https://github.com/voiceflow/general-runtime/blob/48fefb11b93f44794a00fa8ad07cd5ec16e8c375/lib/services/nlu/utils.ts#L32

lib/services/classification/predictor.class.ts
```diff
@@ -28,7 +28,7 @@ const ML_GATEWAY_TIMEOUT = 5000;
 const nonePrediction: Omit<Prediction, 'utterance'> = {
   predictedIntent: VoiceflowConstants.IntentName.NONE,
   predictedSlots: [],
-  confidence: 100,
```
DecathectZero, line 31: confidence is 0 to 1 I believe?

lib/services/classification/classification.const.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const DEFAULT_NLU_INTENT_CLASSIFICATION: IntentClassificationNLUSettings = {
+  type: IntentClassificationType.NLU,
+  params: { confidence: 0.6 },
+};
+
+export const DEFAULT_LLM_INTENT_CLASSIFICATION: IntentClassificationLLMSettings = {
+  type: IntentClassificationType.LLM,
+  params: {
+    model: AIModel.GPT_4_TURBO,
```
theProf, line 17: should this be gpt-4-turbo or did we downgrade default to gpt-3.5-turbo because of this: https://voiceflowhq.slack.com/archives/C067C1WFH97/p1711401316578709

theProf (approved):

PR #756, opened 2024-04-04
[Merged by Bors] - chore(build): bump @voiceflow/orb-common@0.62.2 for cache-to fix

theProf:

Bump orb-common to latest to get cache-to fix so master (and all branches) will push it's cache to registry for all branches to utilize

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ parameters:
     default: "SHA256:7FhWIIg1mYzFjcATPWZB3H9VVpUQihp2rQh1HFhPdks"
 
 orbs:
-  vfcommon: voiceflow/common@0.62.1
+  vfcommon: voiceflow/common@0.62.2
   gomplate: xavientois/gomplate@0.2.0
 
 workflows:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -38,7 +38,7 @@ workflows:
                 command: jq --raw-output '.volta.node // ""' package.json | sed -E "s/^([0-9]+)(\.[0-9]+)?.*$/\1\2/g"
             - vfcommon/set-json-string:
                 field: common_orb_version
-                value: "0.62.1"
+                value: "0.62.2"
```
xavidop, line 41: we should read this from the env var as we do in other repos so we do not need to open PRs every time we have. new version

xavidop (changes requested):

.circleci/config.yml
theProf: good call. how do i update that env var if i want to track latest instead of the test context COMMON_ORB_VERSION?

xavidop (approved):

PR #757, opened 2024-04-04
feat: skip NLC results if LLM set (PL-844)

theProf:

Don't use NLC if LLM enabled

### DIFF
--- a/lib/services/classification/predictor.class.ts
+++ b/lib/services/classification/predictor.class.ts
@@ -319,8 +319,8 @@ export class Predictor extends EventEmitter {
 
   public async predict(utterance: string): Promise<Prediction | null> {
     // 1. first try restricted regex (no open slots) - exact string match
-    const nlcPrediction = await this.nlc(utterance, false);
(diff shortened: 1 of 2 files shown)


DecathectZero (approved):

lib/services/classification/predictor.class.ts
```diff
@@ -320,7 +320,7 @@ export class Predictor extends EventEmitter {
   public async predict(utterance: string): Promise<Prediction | null> {
     // 1. first try restricted regex (no open slots) - exact string match
     const nlcPrediction = await this.nlc(utterance, false);
-    if (nlcPrediction) {
+    if (nlcPrediction && !isIntentClassificationLLMSettings(this.settings)) {
```
DecathectZero, line 323: shouldn't we move this condition up a level, it's better to never even run nlc if the setting doesn't allow it.

DecathectZero (changes requested):

trs (approved):

PR #761, opened 2024-04-05
[Merged by Bors] - feat: add listen functionality (COR-000)

zhihil:
**Fixes or implements COR-1754, COR-1752**

Adds the ability for function step code to listen for user input and follow a specific path depending on that input. Simply put, we are implementing a simplified version of Custom Actions in the Function step.

Please import the following Function into the Voiceflow CMS to test this feature:

[Listen_Functionality_20240424.json](https://github.com/voiceflow/general-runtime/files/14907894/Listen_Functionality_20240424.json)

Updated the Function step to support listen functionality. This feature is simple, but a bit technical to explain. This section describes what it involves. If the feature remains confusing, please ping @zhihil for a video chat to go over the feature.

To implement listen functionality, this PR introduces new options for the **runtime commands** returned by Function step code. As a reminder, the runtime commands is a JSON object providing declarative commands that trigger certain `general-runtime` behaviours, e.g, sending a speak trace.

To illustrate these new features, consider the following function:

```js
export default async function main(args) {
    return {
        trace: [
          {
            type: 'choice',
            payload: {
              buttons: [
                {
                  name: "Button B",
                  request: { type: 'event_A', payload: { label: 'example' } }
                },
                {
                  name: "Button C",
                  request: { type: 'event_B', payload: { label: 'another-example' } }
                }
              ]
            }
          }
        ],
        next: {
          listen: true,
          defaultTo: 'path-A',
          to: [
            {
              on: { 'event.type': 'event_A' },
              dest: 'path-B' // shorthand form `dest`
            },
            {
              on: { 'event.type': 'event_B' },
              dest: { type: 'path', path: 'path-C' } // long form `dest`
            }
          ]
        }
    };
}
```

In this example, we can see two new features:
1. Choice trace - Renders buttons UI on the Prototype Tool and `react-chat`
2. Next Many Command - Induces the Function step to listen for user input and follows a specific path based on the user input.

The flow of the listen functionality works as follows:

Screen Shot 2024-04-08 at 12 10 36 PM

1. User interacts with `creator-app` (or another Voiceflow UI client)
2. `creator-app` initiates an interaction with the `general-runtime`
3. `general-runtime` runs the Function step handler and executes the example above
4. The Function code returns the runtime commands to `general-runtime`
5. `general-runtime` returns the `choice` trace to the client.
6. `general-runtime` saves the data in `next` command in the program variables state (which gets written to MongoDB after `general-runtime` completes an interaction)
7. `general-runtime` stops execution of the assistant at the function step and the HTTP response is sent to `creator-app`. This completes one interaction
8. `creator-app` displays the `choice` trace as buttons on the UI
9. User clicks one of the buttons 
10. `creator-app` sends the `request` payload associated with the button back to `general-runtime`
11. `general-runtime` receives the `request` and loads the `next` command data saved in step (6). It checks if the `request` matches any of the `to` entries and follows that path. Otherwise, Function step follows the path listed in `defaultTo`
12. `general-runtime` leaves the Function step and continues execution of the assistant.

#### `defaultTo`

One question you may have is: "if the user is supposed to click the buttons, which typically has a "correct" payload matching some `to` branch, then when would the `defaultTo` branch actually execute?" 

The answer is that the user does **not** actually need to click the buttons. They can send a text utterance, which, if it does not match any `to` branch, will match with the `defaultTo` branch.

#### Underscore query matching system 

The user utterance matching is implemented using the [underscore-query](https://www.npmjs.com/package/underscore-query) library. How the matching system works is as follows:

Today, clients can send different types of request to `general-runtime`, such as a text utterance, an intent, or a general event. The `general-runtime` exposes the request object to step handlers via the `runtime.getRequest()` method. In turn, we expose this object to the Function code as a `FunctionRequestContext` with the following shape:

```js
const requestContext = {
   event: runtime.getRequest()
}
```

Functions code would then write `underscore-query` queries in the `on` property of a `ConditionalTransfer`, to indicate that it wants to match on a specific request payload.

For example, given a general event request payload

```js
{ type: 'event_A', payload: { label: 'example' } }
```

We may choose to match the request payload using any of the following queries:
```js
{
   on: { 'event.type':  'event_A' }
   dest: 'path-A'
}

{
   on: { 'event.payload.label':  'example' }
   dest: 'path-A'
}

{
   on: { 'event.type': 'event_A', 'event.payload.label':  'example' }
   dest: 'path-A'
}
```

Specifically, the matching is done by running the following code:

```js
const requestContext = {
   event: runtime.getRequest()
}

const conditionalTransfers = [
  {
    on: { 'event.type': 'event_A' },
    dest: { type: 'path', path: 'path-B' }
  },
  {
    on: { 'event.type': 'event_B' },
    dest: { type: 'path', path: 'path-C' }
  }
];

const firstMatchingTransfer = conditionalTransfers.to.find((item) => _query([requestContext], item.on).length > 0);
```

If we have a match, then we follow the path specified in `dest`.

### Setup information

N/A

### Deployment Notes

N/A

### Related PRs

N/A

### Checklist


runtime/lib/Handlers/function/index.ts
```diff
@@ -46,7 +55,19 @@ function applyTraceCommand(command: TraceCommand, runtime: Runtime): void {
...
+function applyNextCommand(
+  command: NextCommand,
+  runtime: Runtime,
+  { nodeId, paths }: { nodeId: string; paths: FunctionCompiledInvocation['paths'] }
+): string | null {
+  if ('listen' in command) {
+    if (!command.listen) return null;
+
+    const { defaultTo, to } = command;
+    runtime.variables.set(VoiceflowConstants.BuiltInVariable.FUNCTION_CONDITIONAL_TRANSFERS, { defaultTo, to });
```
zhihil, line 67: This line saves the conditional transfers specified by a Next Many Command to program variable state.

runtime/lib/Handlers/function/index.ts
```diff
@@ -46,7 +55,19 @@ function applyTraceCommand(command: TraceCommand, runtime: Runtime): void {
   });
 }
 
-function applyNextCommand(command: NextCommand, paths: FunctionCompiledInvocation['paths']): string | null {
+function applyNextCommand(
+  command: NextCommand,
+  runtime: Runtime,
+  { nodeId, paths }: { nodeId: string; paths: FunctionCompiledInvocation['paths'] }
+): string | null {
+  if ('listen' in command) {
```
zhihil, line 63: Here we detect if we are dealing with a Next Many Command or the existing Next (One) Command.

runtime/lib/Handlers/function/index.ts
```diff
@@ -73,6 +94,37 @@ function resolveFunctionDefinition(
...
+
+  throw new Error(`Function produced a transfer object with an unexpected type '${transfer.type}'`);
+}
+
+function handleListenResponse(
+  conditionalTransfers: NextBranches,
+  requestContext: FunctionRequestContext,
+  paths: FunctionCompiledInvocation['paths']
+): string {
+  const firstMatchingTransfer = conditionalTransfers.to.find((item) => _query([requestContext], item.on).length > 0);
```
zhihil, line 116: This line is responsible for processing the user utterance and determining a matching transfer from the saved Next Many Command.

runtime/lib/Handlers/function/index.ts
```diff
@@ -83,6 +135,31 @@ export const FunctionHandler: HandlerFactory<FunctionCompiledNode, typeof utilsO
...
+
+        runtime.variables.set(VoiceflowConstants.BuiltInVariable.FUNCTION_CONDITIONAL_TRANSFERS, null);
+
+        return nextId;
+      }
+
+      /**
+       * Case 2 - If there are no `parsedTransfers`, then we are hitting this Function step for the
+       *          first time
+       */
```
zhihil, line 162: Stage [1/2]

When the `general-runtime` enters a Function step, we execute Case 2, which runs the functions code, gets the runtime commands containing a Next Many Command, and then we pause assistant execution until user input.

runtime/lib/Handlers/function/index.ts
```diff
@@ -83,6 +135,31 @@ export const FunctionHandler: HandlerFactory<FunctionCompiledNode, typeof utilsO
...
     try {
+      const parsedTransfers = NextBranchesDTO.safeParse(
+        runtime.variables.get(VoiceflowConstants.BuiltInVariable.FUNCTION_CONDITIONAL_TRANSFERS)
+      );
+
+      /**
+       * Case 1 - If there is a `parsedTransfers`, then we are resuming Function step execution after
+       *          obtaining user input
+       */
+      if (parsedTransfers.success) {
```
zhihil, line 146: Stage [2/2] 

When the user provides the Function step user input, we execute Case 1 here. This processes the user input, determines a matching transfer from the saved Next Many Command, and then follows the corresponding path.

runtime/lib/Handlers/function/index.ts
```diff
@@ -83,6 +135,31 @@ export const FunctionHandler: HandlerFactory<FunctionCompiledNode, typeof utilsO
...
+        const conditionalTransfers = parsedTransfers.data;
+        const requestContext: FunctionRequestContext = {
+          event: runtime.getRequest(),
+        };
+
+        const nextId = handleListenResponse(conditionalTransfers, requestContext, invocation.paths);
+
+        runtime.variables.set(VoiceflowConstants.BuiltInVariable.FUNCTION_CONDITIONAL_TRANSFERS, null);
```
zhihil, line 154: This line simply resets the Function step state by deleting the saved Next Many Command. The next time a Function step is encountered, we execute Case 2 instead of Case 1, which is the expected behaviour.

lib/services/test/index.ts
```diff
@@ -43,7 +43,7 @@ export class TestService extends AbstractManager {
 
       const executionTime = endTime - startTime;
 
-      const debugTrace: Trace = createFunctionExceptionDebugTrace(err);
+      const debugTrace: UnknownTrace = createFunctionExceptionDebugTrace(err);
```
zhihil, line 46: You can more or less ignore the renaming of `Trace` to `UnknownTrace` changes. This is to disambiguate the simplified trace type in `general-runtime` with an identically named `Trace` type in one of the dependencies.

lib/services/runtime/handlers/state/preliminary/index.ts
```diff
@@ -14,6 +14,7 @@ import InteractionHandler from '../../interaction';
 
 const _v1Handler = _V1Handler();
 export const eventHandlers = [
+  FunctionHandler(),
```
zhihil, line 17: This is registers the Function step handler so that it can receive request payloads like intent request.

utils/underscore-query.ts
```diff
@@ -0,0 +1,9 @@
+import _ from 'lodash';
+import lodashModifier from 'underscore-query';
+
+lodashModifier(_);
+
+// !TODO! - There must be a better way to type this.
+const query = <T>(collection: T[], criteria: Record<string, any>): T[] => (_ as any).query(collection, criteria);
+
+export default query;
```
zhihil, line 9: This module is necessary because the `underscore-query` library has a very unusual import method. You are supposed to import `lodash` and modify a `_` instance with `underscore-query`. 

To quarantine the complexity of this import, I've created this file to handle importing `underscore-query`. All `general-runtime` code should import this file to acquire `_.query`, rather than doing the lodash modification logic.

zhihil (commented):
Additional comments for context:

runtime/lib/Handlers/function/index.ts
z4o4z, line 146: should we handle this part only during 'request' stage?
```suggestion
      if (runtime.getAction() === Action.REQUEST && parsedTransfers.success) {
```

runtime/lib/Handlers/function/index.ts
z4o4z: so i guess we shouldn't define that variable in the `BuiltInVariable` enum since this enum is only for public variables. instead we should define local variable name, something like: `const FUNCTION_CONDITIONAL_TRANSFERS = '__VF_FUNCTION_CONDITIONAL_TRANSFERS__'`

runtime/lib/Handlers/function/index.ts
zhihil: Not sure, what is the `Action` enum used for?

runtime/lib/Handlers/function/index.ts
zhihil: I've defined an `InternalVariables` enum which in `Runtime/Constants` which should store the types of internal variables from now on. 

```ts
import { Enum } from "@/runtime/typings/enum";

const internalVarName = (name: string): string => `__vf__internal_${name}__`

/**
 * NOTE: Internal variables should be prefixed with `__vf_internal_` and suffixed with `__` to
 * avoid name collisions with user-defined variables.
 */
export const InternalVariables = {
    FUNCTION_CONDITIONAL_TRANSFERS: internalVarName("function_conditional_transfers")
};

export type InternalVariables = Enum<typeof InternalVariables>;

```

runtime/lib/Handlers/function/index.ts
z4o4z: yeah you right

trs (approved):

PR #768, opened 2024-04-09
[Merged by Bors] - fix: add voice to global no reply (PL-924)

trs:
Utilize the `generateOutput` utility to add the configured voice to the global no reply message.

PR #769, opened 2024-04-09
fix: add voice to global no match (COR-1712)

DecathectZero:
repeat the no-reply solution https://github.com/voiceflow/general-runtime/pull/768 for no-match as well.
This ensures consistent behavior.


trs (approved):

trs:
bors merge

trs:
bors cancel

PR #776, opened 2024-04-11
feat: remove concurrent vm execution (PL-894)

pmvrmc:
Remove concurrent vm execution, we have settled on doing remoteVM execution

### DIFF
--- a/runtime/lib/Handlers/code/index.ts
+++ b/runtime/lib/Handlers/code/index.ts
@@ -4,7 +4,6 @@ import safeJSONStringify from 'json-stringify-safe';
 import _ from 'lodash';
 import { isDeepStrictEqual } from 'util';
 
-import log from '@/logger';
 import { HandlerFactory } from '@/runtime/lib/Handler';
 
 import * as utils from './utils';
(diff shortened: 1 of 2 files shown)


runtime/lib/Handlers/code/index.ts
```diff
@@ -44,39 +43,7 @@ const CodeHandler: HandlerFactory<BaseNode.Code.Node, CodeOptions> = ({
...
-            result: ivmResult,
-          }
-        );
-
-        // Remote execution result is the source of truth
-        if (endpointResult.status === 'rejected') {
-          throw endpointResult.reason;
-        }
-        newVariableState = endpointResult.value;
+        newVariableState = await utils.remoteVMExecute(endpoint, reqData);
```
trs, line 46: We could further simplify this, we have 3 cases right now, only need two.

```ts
if (endpoint) {
  newVariableState = await utils.remoteVMExecute(endpoint, reqData);
} else {
  newVariableState = await utils.ivmExecute(reqData, callbacks);
}
```

We can get rid of the `useStrictVM` option all together.

pmvrmc: yeah, good call! updated

trs (approved):

PR #777, opened 2024-04-11
feat: llm error messages in preview endpoint (PL-931)

theProf:

Return LLM error messages in preview endpoint for display in toasts

### DIFF
--- a/lib/controllers/test/index.ts
+++ b/lib/controllers/test/index.ts
@@ -212,9 +212,6 @@ class TestController extends AbstractController {
       ];
     })();
 
-    // Only returning LLM related error message for now (PL-931)
-    const errors = predictions.llm?.error && [predictions.llm.error];
-
     return {
       utterance: predictions.utterance ?? data.utterance,
       nlu: {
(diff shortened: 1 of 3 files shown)


lib/controllers/test/interface.ts
```diff
@@ -75,6 +75,13 @@ export const TestClassificationResponseDTO = z
     nlu: ClassificationDTO,
     llm: ClassificationDTO,
     utterance: z.string(),
+    errors: z.optional(
+      z.array(
```
pmvrmc, line 79: you can do `z.array().optional()` and avoid the extra wrapper

lib/controllers/test/index.ts
```diff
@@ -212,6 +212,9 @@ class TestController extends AbstractController {
       ];
     })();
 
+    // Only returning LLM related error message for now (PL-931)
+    const errors = predictions.llm?.error && [predictions.llm.error];
```
pmvrmc, line 216: nit: mentally the ternary operator is easier for me, somethinglike 
`const errors = predictions.llm?.error ? [predictions.llm.error] : undefined;`

but your solution works fine!

pmvrmc (approved):

lib/controllers/test/index.ts
pmvrmc: you could even inline this since its not used anywhere else

theProf: i think i'll leave this one, but that's good to know!

PR #790, opened 2024-04-23
[Merged by Bors] - feat: updating type-guards to use Zod parsing for more reliable runtime validation (PL-753)

zhihil:
**Fixes or implements PL-753**

Updated the request type guards in `general-runtime` to use Zod parsing instead. This provides more reliable runtime validation. 

Previously, the `general-runtime` only checke the `type` property of the request object, which was unreliable as the actual data sent to the API may not match the type declaration. By using Zod, we provide complete validation that the incoming request object matches the type. 

The main changes are found here: `lib/services/runtime/types.ts`. This PR replaces many of the simple type-guards with an `DTO.isXRequest()` function. Under the hood, this function uses Zod parsing, i.e, `XRequestDTO.parse(value).success`

Additionally, Zod parsing has been added to the following controllers which expect an object of type `RuntimeRequest` in the body:
1. `lib/controllers/interact.ts`
2. `lib/controllers/public/index.ts`
4. `lib/controllers/stateManagement/index.ts`

This ensures that the incoming request objects are being validated.

All other changes simply apply the new type-guards.

### Setup information

N/A

### Deployment Notes

N/A

### Related PRs

1. https://github.com/voiceflow/libs/pull/522
5. https://github.com/voiceflow/general-runtime/pull/790
6. https://github.com/voiceflow/creator-app/pull/8046

### Checklist


lib/services/runtime/index.ts
```diff
@@ -64,7 +64,7 @@ class RuntimeManager extends AbstractManager<{ utils: typeof utils }> implements
     }
 
     if (isPathRequest(request)) {
-      runtime.variables.set(VoiceflowConstants.BuiltInVariable.LAST_UTTERANCE, request.payload.label);
+      runtime.variables.set(VoiceflowConstants.BuiltInVariable.LAST_UTTERANCE, request.payload?.label);
```
zhihil, line 67: `.label` is already potentially undefined, so the optional chaining shouldn't cause a huge deviation from existing behaviour.

lib/services/aiAssist.ts
```diff
@@ -25,7 +25,7 @@ class AIAssist extends AbstractManager implements ContextHandler {
     }
 
     if (isPathRequest(request)) {
-      return request.payload.label ?? null;
+      return request.payload?.label ?? null;
```
zhihil, line 28: `.label` here was already potentially `undefined`, so adding this optional chaining should not create big issues.

lib/services/runtime/types.ts
```diff
@@ -13,32 +13,32 @@ export interface Prompt {
...
+export const isTextRequest = (request?: RuntimeRequest): request is BaseRequest.TextRequest =>
   !!request && BaseRequest.isTextRequest(request) && typeof request.payload === 'string';
 
 /**
  * Intent request is being reused for both Alexa events and intent events. To distinguish them we check
  * if `request.payload.data` exists, if so this is an Alexa event request
  * otherwise it is a normal intent request
  */
-export const isIntentRequest = (request?: RuntimeRequest | null): request is BaseRequest.IntentRequest =>
+export const isIntentRequest = (request?: RuntimeRequest): request is BaseRequest.IntentRequest =>
```
zhihil, line 24: The `| null` here is redundant because `RuntimeRequest` already comes with `| null`.

zhihil (commented):
Context:

lib/services/runtime/types.ts
```diff
@@ -13,34 +14,34 @@ export interface Prompt {
   voice?: string;
 }
 
-export const isTextRequest = (request?: RuntimeRequest | null): request is BaseRequest.TextRequest =>
-  !!request && BaseRequest.isTextRequest(request) && typeof request.payload === 'string';
+export const isTextRequest = (request?: RuntimeRequest): request is BaseRequest.TextRequest =>
+  !!request && isTextRequest(request) && typeof request.payload === 'string';
```
pmvrmc, line 18: arent we missing `DTO.isTextRequest` here? otherwise this will make an infinite loop

zhihil: You're right.

lib/services/runtime/types.ts
```diff
@@ -13,34 +14,26 @@ export interface Prompt {
   voice?: string;
 }
 
-export const isTextRequest = (request?: RuntimeRequest | null): request is BaseRequest.TextRequest =>
-  !!request && BaseRequest.isTextRequest(request) && typeof request.payload === 'string';
+export const isTextRequest = (request?: RuntimeRequest): request is BaseRequest.TextRequest =>
+  !!request && DTO.isTextRequest(request);
```
zhihil, line 18: The `typeof request.payload === 'string'` check is redundant.

lib/services/slots/index.ts
```diff
@@ -1,19 +1,28 @@
...
+      BaseRequest.isIntentRequest(value) &&
+      !!value.payload?.intent?.name &&
+      Array.isArray(value.payload.entities) &&
+      !value.payload?.data
+    );
+  }
+
   handle = async (context: Context) => {
-    if (!isIntentRequest(context.request)) {
+    if (!this.isLUISIntentRequest(context.request)) {
```
zhihil, line 25: This is legacy (possibly unused) code which expects a value that doesn't define the required property `value` on slots. 

I don't think we should make `value` optional to accommodate this old type, so I've instead created a legacy type guard to allow the code to continue working.

lib/controllers/interact.ts
```diff
@@ -31,6 +32,13 @@ class InteractController extends AbstractController {
       { locale?: string; logs: RuntimeLogs.LogLevel }
     >
   ): Promise<ResponseContext> {
+    if (req.body.request) {
+      req.body.request = AnyRequestDTO.parse(req.body.request);
+    }
+    if (req.body.action) {
+      req.body.action = AnyRequestDTO.parse(req.body.action);
+    }
```
zhihil, line 40: Various controller methods have been updated to add Zod parsing to the request objects. This ensures that the incoming request data is valid and adapted when necessary, e.g, intent request objects may lack the required `entities` and `query` properties and the DTO provides default values for them.

lib/services/runtime/types.ts
```diff
@@ -1,51 +1,61 @@
...
+export type PathRequest = z.infer<typeof PathRequestDTO>;
+
 export interface Prompt {
   content: BaseText.SlateTextValue | string;
   voice?: string;
 }
 
-export const isTextRequest = (request?: RuntimeRequest | null): request is BaseRequest.TextRequest =>
-  !!request && BaseRequest.isTextRequest(request) && typeof request.payload === 'string';
-
```
zhihil, line 18: Many of these type-guards have been deleted and replaced with type-guards from the DTO package, which have more stringent validation requirements.

lib/services/aiAssist.ts
```diff
@@ -16,16 +17,16 @@ class AIAssist extends AbstractManager implements ContextHandler {
   static StringStorageKey = 'vf_memory';
 
   static getInput(request: RuntimeRequest) {
-    if (isIntentRequest(request)) {
+    if (DTO.isLegacyIntentRequest(request)) {
```
trs, line 20: Curious about the reason to have these under `DTO` rather than imported directly.

trs (approved):
This could cause some unintended issues with peoples requests, we'll definitely want to keep an eye on requests.

lib/controllers/interact/index.ts
```diff
@@ -26,7 +27,7 @@ class InteractController extends AbstractController {
     const { versionID } = params;
     const { sessionID } = body.session;
     const { userID } = body.session;
-    const { action } = body;
+    const action = await AnyRequestDTO.parseAsync(req.body.action);
```
trs, line 30: Could update the above `InteractRequestBody` to have the `AnyRequestDTO`, and not have to do parsing here.

trs (changes requested):

lib/services/aiAssist.ts
zhihil: It was to disambiguate it from the type guards that are in `general-runtime`. It's a style that's also used on `creator-app` to provide clearer separation between what Voiceflow code belongs to what package.

Is it fine to adopt this convention for DTOs here?

trs (approved):

PR #791, opened 2024-04-23
refactor: use rxjs, review changes (PL-909)

trs:


lib/services/runtime/handlers/aiResponse/index.ts
```diff
@@ -107,73 +109,42 @@ const AIResponseHandler: HandlerFactory<VoiceNode.AIResponse.Node, void, General
...
+                acc.queryTokens += completion.queryTokens;
+                acc.tokens += completion.tokens;
+                acc.model = completion.model;
+                acc.multiplier = completion.multiplier;
+                return acc;
+              }, response)
+            )
+          ),
+          completion$.forEach((trace) => runtime.trace.addTrace(trace)),
+        ]);
```
theProf, line 147: i believe i had an if condition on `isEmpty` before doing the `response` or `forEach`, the reason being that if there were no values, we shouldn't mutate `response` or `addTraces`, however, the use of `endWith` injects a single value always, so you can end up with only the EndTrace. 

I'm looking for a more elegant way to do it, but for now doing something like this would help: 
```
        const emptyStream = await lastValueFrom(promptStream$.pipe(isEmpty()));

        const completion$ = iif(() => emptyStream, NEVER, promptStream$).pipe(
          skipLast(1),
          filter((completion) => completion.output !== ''),
          map((completion, i) =>
            i > 0 ? completionToContinueTrace(completion) : completionToStartTrace(runtime, node, completion)
          ),
          endWith(endTrace())
        );

        [response] = await Promise.all([
          lastValueFrom(
            promptStream$.pipe(
              reduce((acc: any, completion: any) => {
                acc.output += completion.output;
                acc.answerTokens += completion.answerTokens;
                acc.queryTokens += completion.queryTokens;
                acc.tokens += completion.tokens;
                acc.model = completion.model;
                acc.multiplier = completion.multiplier;
                return acc;
              }, response)
            )
          ),
          completion$.forEach((trace) => runtime.trace.addTrace(trace)),
        ]);
```

`if and only if` the event stream is empty, send a `NEVER`, otherwise, use the stream. the `NEVER` is important because it will not emit values AND will not complete. since `endWith` relies on the `complete`, we effectively bypass the pipeline, and no values are added to trace

theProf (approved):

lib/services/runtime/handlers/aiResponse/index.ts
trs: Makes sense!

PR #797, opened 2024-04-26
feat: use rxjs, refactor systems (PL-909)

trs:


lib/services/runtime/handlers/aiResponse/index.ts
```diff
@@ -107,73 +123,50 @@ const AIResponseHandler: HandlerFactory<VoiceNode.AIResponse.Node, void, General
...
+
+        const completion$ = concat(
+          promptStream$.pipe(
+            filter((completion) => completion.output != null),
+            map((completion, i) =>
+              i > 0 ? completionToContinueTrace(completion) : completionToStartTrace(runtime, node, completion)
+            )
+          ),
+          promptStream$.pipe(
+            take(1),
```
theProf, line 145: Don't need `take`

> As soon as the source Observable emits a value, isEmpty will emit a false and complete.

lib/services/runtime/handlers/aiResponse/index.ts
```diff
@@ -107,73 +123,50 @@ const AIResponseHandler: HandlerFactory<VoiceNode.AIResponse.Node, void, General
...
+            isEmpty(),
+            switchMap((isEmpty) => iif(() => isEmpty, NEVER, of(endTrace())))
```
theProf, line 147: No longer need `iif` since no `endWith`, can be a regular if: `=> isEmpty ? EMPTY : of(endTrace())`
`NEVER` and `iif` was only needed to chabge the source to not emit or complete, but sending an `EMPTY` to just complete is fine now that we don't inject values

PR #799, opened 2024-05-02
feat: support streaming with knowledge base (PL-987)

trs:


lib/services/aiSynthesis/index.ts
```diff
@@ -61,22 +60,55 @@ class AISynthesis extends AbstractManager {
...
+      map((completion) => {
+        if (!completion.output) return completion;
+
+        completion.output = this.filterNotFound(completion.output);
+        completion.output = removePromptLeak(completion.output);
+        completion.output = completion.output || null;
+
+        return completion;
+      }),
+      filter((completion) => completion != null)
```
pmvrmc, line 84: do we really want `!=` instead of `!==` here? we can use `!==` and detail each edge case (null/undefined) instead of relying on implicit js casting

lib/services/aiSynthesis/index.ts
```diff
@@ -220,22 +252,53 @@ class AISynthesis extends AbstractManager {
...
+    return from(
+      this.services.aiSynthesis.answerSynthesisStream({
+        question,
+        instruction,
+        data,
+        options: settings?.summarization,
+        context: { projectID: project._id, workspaceID: project.teamID },
+      })
+    ).pipe(
+      filter((answer) => answer != null),
```
pmvrmc, line 266: same question on the use of `!=`

lib/services/aiSynthesis/index.ts
```diff
@@ -61,22 +60,54 @@ class AISynthesis extends AbstractManager {
...
+    question: string;
+    instruction?: string;
+    data: KnowledgeBaseResponse;
+    variables?: Record<string, any>;
+    options?: Partial<BaseUtils.ai.AIModelParams>;
+    context: AIModelContext;
+  }): Promise<AIResponse | null> {
+    const response: AIResponse = await lastValueFrom(
+      from(this.answerSynthesisStream(params)).pipe(
+        reduce((acc, completion) => {
```
trs, line 97: This same reduce pattern is in a few places, but I'm not sure it's worth making generic. It's just different enough in each place.

lib/services/runtime/handlers/aiResponse/index.ts
```diff
@@ -1,209 +1,46 @@
...
-            })),
-            query: {
-              messages: answer.messages,
-              output: answer.output,
-            },
-          },
-        });
-
+      // Exit early if the model is not available
+      if (!!node.model && !canUseModel(node.model, runtime)) {
```
trs, line 21: Moved the model check out of the behaviour, as we weren't checking model with the KB previously.

trs (commented):
The ai response methods are very similar, but require slightly different implementations. A future refactor of how we call LLMs might be useful: easier to chain KB with the LLM.

lib/services/aiSynthesis/index.ts
```diff
@@ -220,22 +251,50 @@ class AISynthesis extends AbstractManager {
...
+
+    return from(
+      this.services.aiSynthesis.answerSynthesisStream({
+        question,
+        instruction,
+        data,
+        options: settings?.summarization,
+        context: { projectID: project._id, workspaceID: project.teamID },
+      })
+    ).pipe(map((answer, i) => (i === 0 ? { chunks, ...answer! } : answer!)));
```
theProf, line 264: not a huge fan of `!` making it to production code. is this something that can be typed, commented, or fixed with a version upgrade (to typescript or rxjs)?

lib/services/aiSynthesis/index.ts
```diff
@@ -220,22 +251,50 @@ class AISynthesis extends AbstractManager {
...
+    synthesis?: boolean;
+    options?: {
+      search?: Partial<BaseModels.Project.KnowledgeBaseSettings['search']>;
+      summarization?: Partial<BaseModels.Project.KnowledgeBaseSettings['summarization']>;
     };
+    tags?: BaseModels.Project.KnowledgeBaseTagsFilter;
+  }): Promise<KBResponse> {
+    return lastValueFrom(
+      from(this.knowledgeBaseQueryStream(params)).pipe(
+        switchMap((stream) => stream),
```
theProf, line 281: do you have specific reasoning for `switchMap` here? I've noticed you seem to gravitate towards `switch` over `merge` or `concat`. I find that `switch` is more useful when there is external signaling so that you can cancel/drop emissions (ex: when fetching to be able to ignore a previous fetch for the latest, switch on click events). we have mostly been dealing with single source streams and want all of the emissions, where `merge|concat` are better, and since we generally want to keep order, `concat(Map)` is the best pick.

lib/services/runtime/handlers/aiResponse/knowledge-base.ts
```diff
@@ -0,0 +1,121 @@
...
+        output: answer.output,
+      },
+    },
+  });
+
+  const outputString = answer?.output || 'Unable to find relevant answer.';
+  const output = generateOutput(outputString, runtime.project, node.voice ?? getVersionDefaultVoice(runtime.version));
+
+  // Set last response to entire AI response, not just a partial chunk
+  variables.set(VoiceflowConstants.BuiltInVariable.LAST_RESPONSE, outputString);
```
theProf, line 116: do we / should we enforce any limits here?

lib/services/runtime/handlers/aiResponse/model.ts
```diff
@@ -0,0 +1,111 @@
...
+    getOutputTrace({
+      output,
+      variables,
+      version: runtime.version,
+      ai: true,
+    })
+  );
+
+  // Set last response to entire AI response, not just a partial chunk
+  variables.set(VoiceflowConstants.BuiltInVariable.LAST_RESPONSE, response.output);
```
theProf, line 106: same as above re: limits

theProf (approved):
looks good. would just change all your `switchMap`s to `concatMap`s

lib/services/aiSynthesis/index.ts
```diff
@@ -61,22 +60,54 @@ class AISynthesis extends AbstractManager {
...
+        {
+          retries: this.DEFAULT_ANSWER_SYNTHESIS_RETRIES,
+          retryDelay: this.DEFAULT_ANSWER_SYNTHESIS_RETRY_DELAY_MS,
+          context,
+        },
+        variables
+      )
+    ).pipe(
+      map((completion) => {
+        if (!completion.output) return completion;
```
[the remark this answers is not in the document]
trs: Nope, that'll trim each section, which can and does have spaces

lib/services/runtime/handlers/aiResponse/knowledge-base.ts
trs: I don't think so. AI Responses have a maximum number of tokens we can return anyway.

lib/services/aiSynthesis/index.ts
trs: The `switchMap` is used because `knowledgeBaseQueryStream` returns a `Promise`. So we resolve the promise and switch to the resolved observable. I don't think merge/concat is the correct behaviour here.

trs: Thanks for clarifying this. The inner observable could never emit if the source emits sooner, since it switches. `concatMap` is totally the correct choice.

lib/services/runtime/handlers/aiResponse/knowledge-base.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  const completion$ = concat(
+    promptStream$.pipe(
+      map((completion, i) =>
+        i > 0 ? completionToContinueTrace(completion) : completionToStartTrace(runtime, node, completion)
+      )
+    ),
+    promptStream$.pipe(
+      isEmpty(),
+      switchMap((isEmpty) => (isEmpty ? NEVER : of(endTrace())))
```
theProf, line 48: `concatMap`

lib/services/runtime/handlers/aiResponse/model.ts
```diff
@@ -0,0 +1,111 @@
...
+  const completion$ = concat(
+    promptStream$.pipe(
+      filter((completion) => completion.output != null),
+      map((completion, i) =>
+        i > 0 ? completionToContinueTrace(completion) : completionToStartTrace(runtime, node, completion)
+      )
+    ),
+    promptStream$.pipe(
+      isEmpty(),
+      switchMap((isEmpty) => (isEmpty ? NEVER : of(endTrace())))
```
theProf, line 47: `concatMap`

theProf (approved):

trs:
bors merge

PR #805, opened 2024-05-07
feat: use git shallow clone (IN-1006)

edison-vflow:
### Fixes or implements 
* Linear [IN-1006 ](https://linear.app/voiceflow/issue/IN-1006/use-shallow-clone-instead-of-checkout)

* We can speed up builds by up to and shave off 30 seconds, eg for repos like creator-app 
* Where it takes 30 seconds to clone, it will now take 1 second 
* Do shallow clones for both branches and tags 

### Checklist

- [ ] Tested on creator-app
- [ ] Tested on database-cli 
- [ ] Tested on general-runtime

### DIFF
--- a/.circleci/continue_config.yml.tmpl
+++ b/.circleci/continue_config.yml.tmpl
@@ -50,7 +50,7 @@ defaults:
     tag: {{ .values.node_version }}-vf-1
 
 orbs:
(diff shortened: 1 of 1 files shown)


edison-vflow:
bors try

PR #810, opened 2024-05-08
feat: logging for malformed requests (PL-753)

zhihil:
**Fixes or implements PL-753**

Add logging to track any requests that violate the new DTO type guards we want to introduce.

### Related PRs

### Checklist


lib/controllers/utils.ts
```diff
@@ -1,7 +1,18 @@
...
 import { FullServiceMap } from '../services';
 
 export abstract class AbstractController {
   constructor(public services: FullServiceMap, public config: Config) {}
 }
+
+export function logMalformedRequest(request: Record<string, any> | null | undefined, type: 'action' | 'request') {
+  if (!!request && !AnyRequestDTO.safeParse(request).success) {
+    logger.info(
+      `malformed request object [${type}], [type]=${request?.type}, [json]=${JSON.stringify(request, null, 2)}`
```
pmvrmc, line 15: looks good, but would it be worth it to also log the zod parsing error? it should contain information on why the validation fails

pmvrmc (approved):

lib/controllers/utils.ts
zhihil: The Zod parsing error is somewhat massive and unwieldly. As long as we have the original object, we should be able to just reproduce the Zod parsing error.

I've seen some packages that allow you to write more concise Zod errors, so I'll try those.

lib/controllers/interact/index.ts
```diff
@@ -28,6 +28,7 @@ class InteractController extends AbstractController {
     const { userID } = body.session;
     const { action } = body;
     const { state } = body;
+    logMalformedRequest(action, 'action');
```
trs, line 31: Let's leave the streaming route. This is net new and should have a solid DTO

trs (changes requested):

lib/controllers/interact/index.ts
zhihil: The logging has been removed and replaced with actual DTO input validation of the request object.

lib/controllers/utils.ts
zhihil: Screen Shot 2024-05-09 at 11 04 36 AM

The error messages are a bit too verbose to be useful from a glance 😬 . I'm fairly certain if I see the original request object, I could spot the error more easily.

PR #814, opened 2024-05-13
[Merged by Bors] - feat: support streaming with knowledge base (PL-987)

trs:
Refactor ai synthesis and the ai response step to support streaming via the knowledge base.
We utilize a form of "buffering" the KB responses to look for specific phrases to indicate if we "found" a result.


lib/services/aiSynthesis/buffer-reduce.subject.ts
```diff
@@ -0,0 +1,37 @@
...
+      super.next(value);
+      return;
+    }
+
+    this.buffer = this.buffer ? this.accumulator(this.buffer, value) : value;
+
+    const result = this.predicate(this.buffer);
+    if (result === 'stop') {
+      delete this.buffer;
+      super.error(new Error('BufferedReducerSubject'));
```
trs, line 27: I tried completing the stream, but the downstream use case would just freeze. This way it throws and I catch it.

lib/services/runtime/handlers/aiResponse/model.ts
```diff
@@ -0,0 +1,104 @@
+import { CompletionPrivateHTTPControllerGenerateChatCompletionStream200 as ChatCompletionStream } from '@voiceflow/sdk-http-ml-gateway/generated';
```
trs, line 1: This was moved directly from `aiRespopnse/index.ts`

lib/services/runtime/handlers/aiResponse/knowledge-base.ts
```diff
@@ -0,0 +1,124 @@
+import { deepVariableSubstitution } from '@voiceflow/common';
```
trs, line 1: Moved from `aiResponse/index.ts` and refactored to support streaming

lib/services/aiSynthesis/buffer-reduce.subject.ts
```diff
@@ -0,0 +1,37 @@
...
+  next(value: T): void {
+    if (!this.buffering) {
+      super.next(value);
+      return;
+    }
+
+    this.buffer = this.buffer ? this.accumulator(this.buffer, value) : value;
+
+    const result = this.predicate(this.buffer);
+    if (result === 'stop') {
```
pmvrmc, line 25: maybe replace this 'stop' with a const/enum/symbol for clarity?

lib/services/aiSynthesis/buffer-reduce.subject.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  next(value: T): void {
+    if (!this.buffering) {
+      super.next(value);
+      return;
+    }
+
+    this.buffer = this.buffer ? this.accumulator(this.buffer, value) : value;
+
+    const result = this.predicate(this.buffer);
```
pmvrmc, line 24: a bit confused here, shouldnt the predicate run before the accumulator?

so only if the predicate returns true we actually run the accumulator.

lib/services/aiSynthesis/buffer-reduce.subject.ts
pmvrmc, line 25: its a bit weird that the predicate returns true/false/stop.

would it make sense instead to just return true/false and do a try/catch around it to get the 'stop' usecase?

then the code that calls the buffer would throw the error instead of returning 'stop'. also I guess the error message could be better than just the class name

lib/services/aiSynthesis/index.ts
```diff
@@ -220,22 +259,90 @@ class AISynthesis extends AbstractManager {
...
-      instruction,
-      data,
-      options: settings?.summarization,
-      context: { projectID: project._id, workspaceID: project.teamID },
-    });
+    const stream$ = new BufferedReducerSubject<KBResponse>(
+      (resp) => {
+        const output = (resp.output ?? '').toLocaleUpperCase();
+        // Stop stream all together if output is a bad phrase
+        if (this.filterNotFound(resp.output ?? '') === null) return 'stop';
```
pmvrmc, line 268: we shouldnt need the `?? ''` here, its already on the previous line

lib/services/aiSynthesis/index.ts
```diff
@@ -220,22 +259,90 @@ class AISynthesis extends AbstractManager {
...
-      question,
-      instruction,
-      data,
-      options: settings?.summarization,
-      context: { projectID: project._id, workspaceID: project.teamID },
-    });
+    const stream$ = new BufferedReducerSubject<KBResponse>(
+      (resp) => {
+        const output = (resp.output ?? '').toLocaleUpperCase();
+        // Stop stream all together if output is a bad phrase
```
pmvrmc, line 267: I would replace the "bad phrase" comment by "not found phrase"

lib/services/aiSynthesis/index.ts
```diff
@@ -220,22 +259,90 @@ class AISynthesis extends AbstractManager {
...
 
-    const answer = await this.services.aiSynthesis.answerSynthesis({
-      question,
-      instruction,
-      data,
-      options: settings?.summarization,
-      context: { projectID: project._id, workspaceID: project.teamID },
-    });
+    const stream$ = new BufferedReducerSubject<KBResponse>(
+      (resp) => {
```
pmvrmc, line 265: the resp here is really the buffer right? would it make sense to name this variable buffer?

lib/services/aiSynthesis/buffer-reduce.subject.ts
theProf: the order makes sense as is because we want to check the result of appending the new/next token and seeing if that's a bad phrase. so on the second emission (first: "NO", second: "T_") we want to send `NOT_` to the `predicate` instead of just `NO` which has already been checked

lib/services/aiSynthesis/index.ts
theProf: correct, the `resp` here is the bufferd `acc: KBResponse`. I don't think `buffer` would make sense, as that's good for the generic implementation of `BufferedReducerSubject`, but here we have more context as to what _exactly_ the `buffer` is. 
I agree that `resp` isn't the best name, as it's not a single `resp`, but the buffered/accumulated `resp`s so far. but without coffee and a tenuous grasp on language itself, i have no good suggestion

lib/services/aiSynthesis/index.ts
```diff
@@ -220,22 +259,90 @@ class AISynthesis extends AbstractManager {
...
+        buffer.chunks.push(...(value.chunks ?? []));
+        buffer.output += value.output ?? '';
+        buffer.answerTokens += value.answerTokens;
+        buffer.queryTokens += value.queryTokens;
+        buffer.tokens += value.tokens;
+        buffer.model = value.model;
+        buffer.multiplier = value.multiplier;
+        buffer.faqSet = value.faqSet;
+        return buffer;
+      }
```
theProf, line 286: lets keep this custom Subject usage as immutable as possible, avoid mutate methods like `.push` and opt for spread, `concat`, `reduce`, etc
```suggestion
      (buffer, value) => ({
        chunks: (buffer.chunks || []).concat(value.chunks ?? []),
        output: (buffer.output ?? "") + (value.output ?? ""),
        answerTokens: buffer.answerTokens + value.answerTokens,
        queryTokens: buffer.queryTokens + value.queryTokens,
        tokens: buffer.tokens + value.tokens,
        model: value.model,
        multiplier: value.multiplier,
        faqSet: value.faqSet,
      })
```

lib/services/aiSynthesis/buffer-reduce.subject.ts
theProf, line 1: this will definitely require unit testing. may be worth looking at the marble test frameworks if we start rolling more of our own / using rxjs more

lib/services/aiSynthesis/buffer-reduce.subject.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (result === 'stop') {
+      delete this.buffer;
```
theProf, line 26: should probably delete _after_ the error throw (the error throw sets some other vars that start locking up access, so we'll want to delete _after_ we're sure no one is going to try to access)

may want to add a check and free in the unsubscribe too. we rely on state and emitted values to do clean up right now

lib/services/aiSynthesis/index.ts
trs: It _may_ be a buffered response, but it could be just a plain response. Either way, it's the same type `KBResponse`, and is treated the same as a non buffered response.

lib/services/aiSynthesis/buffer-reduce.subject.ts
```diff
@@ -21,17 +21,20 @@ export class BufferedReducerSubject<T> extends Subject<T> {
...
-      return;
+    try {
+      const result = this.predicate(this.buffer);
+
+      if (!result) {
+        this.buffering = false;
+        super.next(this.buffer);
+      }
+    } catch (err) {
+      super.error(err);
```
pmvrmc, line 32: not sure if applicable, but should we return after `super.error(err);` inside the catch?

I know in this case it doesnt relaly matter, but what Im worried is someone in the future adding something outside of the try/catch. In said scenearion in case of an error it would run both the catch and rhe new code. Also that future possible error wont show up on PR review as the catch clause is not touched, but it will impact the flow.

lib/services/aiSynthesis/index.ts
theProf: i can accept that, i redact my request for change

lib/services/aiSynthesis/buffer-reduce.subject.ts
```diff
@@ -0,0 +1,37 @@
...
+    super();
+  }
+
+  next(value: T): void {
+    if (!this.buffering) {
+      super.next(value);
+      return;
+    }
+
+    this.buffer = this.buffer ? this.accumulator(this.buffer, value) : value;
```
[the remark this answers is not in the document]
trs: That can be done in the `accumulator` when implementing, yeah?

tests/lib/services/aiSyntesis/buffer-reduce.subject.unit.ts
```diff
@@ -0,0 +1,79 @@
...
+    const subject = new BufferedReducerSubject<number>(predicate, accumulator);
+
+    testScheduler.run(({ cold, expectObservable }) => {
+      const test$ = cold('a-b-c-|', { a: 1, b: 2, c: 3 });
+
+      expectObservable(
+        subject.pipe(
+          // convert error to "error" to match expectation of `#` in marble diagram
+          catchError((err) =>
+            err instanceof BufferedReducerStopException ? throwError(() => 'error') : throwError(() => err)
```
theProf, line 71: `throwError` is overkill for this case (usually for when a subject or high order observable unwrapping, etc), you can just regular `throw`. And since this is just testing, it doesn't even need to be an `Error` type. 

bonus fun fact: `throwError` takes a factory OR the thing to throw, so `throwError(new Error('beep'))` or even `throwError('error')` would have sufficed

```suggestion
            err instanceof BufferedReducerStopException ? throw 'error') : throw err
```

theProf (changes requested):

tests/lib/services/aiSyntesis/buffer-reduce.subject.unit.ts
trs: Apparently passing a value to `throwError` is deprecated, the factory is the recommended way.
But yeah, overkill is fair.

theProf: https://github.com/ReactiveX/rxjs/blob/7.8.1/src/internal/observable/throwError.ts#L104

well i'll be, makes sense. thanks!

theProf (approved):

lib/services/aiSynthesis/buffer-reduce.subject.ts
trs: Seems like a really odd thing to guard against IMO.

theProf: if it's being removed by linting, lets go with that

lib/services/aiSynthesis/buffer-reduce.subject.ts
```diff
@@ -0,0 +1,40 @@
...
+      if (!result) {
+        this.buffering = false;
+        super.next(this.buffer);
+      }
+    } catch (err) {
+      super.error(err);
+    }
+  }
+
+  unsubscribe(): void {
```
theProf, line 36: nice

trs:
bors merge

PR #817, opened 2024-05-15
[Merged by Bors] - fix: dont treat as trained if no intents (PL-1040) [bugfix]

trs:
Don't consider the version trained if it does not contain any intents.

### DIFF
--- a/lib/services/classification/predictor.class.ts
+++ b/lib/services/classification/predictor.class.ts
@@ -177,7 +177,6 @@ export class Predictor extends EventEmitter {
           message: 'Something went wrong with NLU prediction',
         },
       };
-      this.debug(DebugType.NLU, this.predictions.nlu.error?.message);
       return null;
     }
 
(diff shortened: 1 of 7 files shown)


lib/services/classification/predictor.class.ts
```diff
@@ -362,6 +360,13 @@ export class Predictor extends EventEmitter {
 
       this.predictions.result = 'llm';
 
+      if (!(llmPrediction.predictedIntent in this.intentNameMap)) {
```
theProf, line 363: is this case not caught [here](https://github.com/voiceflow/general-runtime/blob/fecd85cf2cecac4de55acd445615df659bebfd84/lib/services/classification/predictor.class.ts#L216) in the llm func itself?

theProf (approved):

lib/services/classification/predictor.class.ts
trs: I had this occur during testing, so I added this check 🤔

theProf: ah true, my link if pre-prompt. but maybe this logic can go with the [post-checks](https://github.com/voiceflow/general-runtime/blob/fecd85cf2cecac4de55acd445615df659bebfd84/lib/services/classification/predictor.class.ts#L268) in llm?

trs:
bors merge

PR #822, opened 2024-05-17
[Merged by Bors] - feat: implement text variant handler (PL-1043)

zhihil:
**Fixes or implements PL-1043**

Adds the skeleton code for the Response node handler. This PR only enables the ability to send a text variant, which should unblock the initial demo for the CORE team.

The remainder of the Response step will be implemented in subsequent PRs.

Implemented a new `ResponseHandler` in `general-runtime`. This handler has been placed under `libs/services/runtime` rather than `runtime/lib/Handlers`. I _think_ the latter is supposed to be purely logic and control node handlers, while the former dictates specific content like text, video, and image node handlers.

At a high-level, this is how to execute a response node
1. Read a `responseID` from the compiled response node.
2. Read the Response entity corresponding to `responseID` from `version.programResources`, which is a new data structure that contains compiled version-scoped data like function definitions, NLU data, or CMS responses.
3. In the response entity, select the discriminator corresponding to the current channel + language configuration. For this project, the handler always selects the `default:en-us` discriminator. 
4. In the discriminator entity, output each text variant as a text trace.
5. Leave the response node from the default port.

### Setup information

N/A

### Deployment Notes

N/A

### Related PRs

N/A

### Checklist

- [x] Breaking changes have been communicated, including:
- [x] New environment variables have [been deployed](https://www.notion.so/voiceflow/Add-Environment-Variables-be1b0136479f45f1adece7995a7adbfb)

### DIFF
--- a/lib/services/runtime/handlers/response/index.ts
+++ b/lib/services/runtime/handlers/response/index.ts
@@ -10,9 +10,7 @@ const handlerUtils = {
 };
 
 export const ResponseHandler: HandlerFactory<CompiledResponseNode, typeof handlerUtils> = (utils) => ({
-  canHandle: (node) => {
-    return node.type === NodeType.RESPONSE;
-  },
+  canHandle: (node) => node.type === NodeType.RESPONSE,
 
   handle: async (node, runtime, variables): Promise<string | null> => {
     const { programResources } = runtime.version as any;
--- a/lib/services/runtime/handlers/response/index.ts
+++ b/lib/services/runtime/handlers/response/index.ts
@@ -1,4 +1,4 @@
-import { CompiledResponseNode, NodeType, ResponseVariantType, VersionProgramResources } from '@voiceflow/dtos';
+import { CompiledResponseNode, NodeType, ResponseVariantType, Version } from '@voiceflow/dtos';
 
 import { HandlerFactory } from '@/runtime';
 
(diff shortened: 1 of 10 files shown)


lib/services/runtime/handlers/response/index.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  handle: async (node, runtime, variables): Promise<string | null> => {
+    if (!runtime.version) {
+      throw new Error(`[${RESPONSE_HANDLER_ERROR_TAG}]: Runtime was not loaded with a version`);
+    }
+    /**
+     * !TODO! - Need to replace this `as Version` cast by refactoring `general-runtime` to use the
+     *          `Version` from the DTOs.
+     */
+    const version = runtime.version as unknown as Version;
```
zhihil, line 27: This type-cast is intentional. I am aware it's not good practice to do this, however, the `Version` type from `@voiceflow/dtos` is incomplete and strongly-typing the DTO type is a rabbit-hole that will eat up the scope of the response step project.

zhihil:
bors try

tests/lib/services/runtime/handlers/response/index.unit.ts
```diff
@@ -0,0 +1,123 @@
...
+    });
+
+    it('outputs text trace for each text variant', async () => {
+      const result = await handler.handle(responseNode, runtime, variables, program, eventHandler);
+
+      await expect(result).to.eql(responseNode.ports.default);
+      await expect((runtime.trace.addTrace as sinon.SinonStub).callCount).to.eql(1);
+
+      const addTraceArgs = (runtime.trace.addTrace as sinon.SinonStub).args[0][0];
+      delete addTraceArgs.payload.slate.id;
```
zhihil, line 105: This `id` is randomly generated and causes issues for the unit tests.

lib/services/runtime/handlers/response/index.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    if (!version.programResources) {
+      throw new Error(`[${RESPONSE_HANDLER_ERROR_TAG}]: Version was not compiled`);
+    }
+    const responses = version.programResources.responses[node.data.responseID];
+    const variants = responses.variants[DEFAULT_DISCRIMINATOR];
```
pmvrmc, line 33: should the variant come from some config on the node and we fallback to the deault one if not present?

lib/services/runtime/handlers/response/index.ts
```diff
@@ -0,0 +1,58 @@
...
+
+const RESPONSE_HANDLER_ERROR_TAG = 'response-handler';
+
+const DEFAULT_DISCRIMINATOR = 'default:en-us';
+
+export const ResponseHandler: HandlerFactory<CompiledResponseNode, typeof handlerUtils> = (utils) => ({
+  canHandle: (node) => CompiledResponseNodeDTO.safeParse(node).success,
+
+  handle: async (node, runtime, variables): Promise<string | null> => {
+    if (!runtime.version) {
```
pmvrmc, line 20: should there ever be a case of this? when we create the runtime object and attach the version shouldnt it crash there if version is not present?

pmvrmc (approved):

lib/services/runtime/handlers/response/index.ts
trs: Good point, this might be just old bad typing, since we should definitely have a version at this point.

trs (approved):

lib/services/runtime/handlers/response/index.ts
zhihil: I believe the best choice for the future is to have a project-level setting that is affixed to the version by compilation, since you don't often want default values to vary node-by-node.

Right now I've just defined this constant since the discriminator is a feature that will have to built in the future. We don't have a clear idea of what the schema looks like.

lib/services/runtime/handlers/response/index.ts
zhihil: I don't think I've ever seen it not have a `.version` so it might be a typing issue.

PR #830, opened 2024-05-28
fix: leave slot filling to only dialog service using nlu (PL-1109)

theProf:

Remove slot filling from nlu service and leave solely to dialog service

there are 2 places we do slot filling:
first in the nlu service
the in the dialog manager if needed
They both do the same fillSlots call (now, in my test). but the main difference is that the dialog manager adds a prefix to the utterance.
I'm unsure why the NLU performs poorly on the number match (only takes the first 3?) given the same intent and utterance (without prefix). but the solution i'll be pushing is to remove slot filling from the nlu service and leave all slot filling to be done in the dialog manager until a proper investigation can be done to tweak/optimize/etc

### Checklist

- [x] bors try
- [ ] manual

### DIFF
--- a/lib/services/classification/predictor.class.ts
+++ b/lib/services/classification/predictor.class.ts
@@ -149,16 +149,14 @@ export class Predictor extends EventEmitter {
   }
 
   private async nluGatewayPrediction(utterance: string, options: Partial<NLUPredictOptions>) {
-    const { filteredIntents = [], excludeFilteredIntents = true } = options;
+    const { limit } = options;
 
     const { data: prediction } = await this.config.axios
       .post<NLUIntentPrediction | null>(`${this.nluGatewayURL}/v1/predict/${this.props.versionID}`, {
         utterance,
         tag: this.props.tag,
         workspaceID: this.props.workspaceID,
-        filteredIntents,
-        excludeFilteredIntents,
-        limit: 10,
+        ...(limit ? { limit } : {}),
       })
       .catch((err: Error) => {
         logger.error(err, 'Something went wrong with NLU prediction');
@@ -315,7 +313,7 @@ export class Predictor extends EventEmitter {
       return nlcPrediction;
     }
 
-    const nluPrediction = this.props.isTrained && (await this.nlu(utterance, this.options));
+    const nluPrediction = this.props.isTrained && (await this.nlu(utterance, { limit: 10 }));
 
     if (!nluPrediction) {
       if (isIntentClassificationLLMSettings(this.settings)) {
--- a/lib/services/classification/predictor.class.ts
+++ b/lib/services/classification/predictor.class.ts
@@ -1,4 +1,3 @@
-/* eslint-disable sonarjs/cognitive-complexity */
 import { Utils } from '@voiceflow/common';
 import { DEFAULT_INTENT_CLASSIFICATION_PROMPT_WRAPPER_CODE } from '@voiceflow/default-prompt-wrappers';
 import { IntentClassificationSettings } from '@voiceflow/dtos';
@@ -6,7 +5,6 @@ import { ISlotFullfilment } from '@voiceflow/natural-language-commander';
 import { VoiceflowConstants } from '@voiceflow/voiceflow-types';
 import type { AxiosStatic } from 'axios';
 import { EventEmitter } from 'node:events';
-import { match } from 'ts-pattern';
 
 import MLGateway from '@/lib/clients/ml-gateway';
 import logger from '@/logger';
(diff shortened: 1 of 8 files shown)


theProf:
bors try

lib/services/classification/predictor.class.ts
```diff
@@ -314,6 +312,7 @@ export class Predictor extends EventEmitter {
     const nlcPrediction = !isIntentClassificationLLMSettings(this.settings) ? await this.nlc(utterance, false) : null;
     if (nlcPrediction) {
       this.predictions.result = 'nlc';
+      this.debug('NLC' as DebugType);
```
trs, line 315: We don't want to expose NLC to the end user

theProf: whoops, missed a debug line. thanks

lib/services/dialog/index.ts
```diff
@@ -177,14 +176,30 @@ class DialogManagement extends AbstractManager<{ utils: typeof utils }> implemen
...
+            },
             {
               locale: version.prototype?.data.locales[0] as VoiceflowConstants.Locale,
               hasChannelIntents: project?.platformData?.hasChannelIntents,
               platform: version.prototype.platform as VoiceflowConstants.PlatformType,
             }
           );
-          dmPrefixedResult = getIntentRequest(await predictor.predict(`${prefix} ${query}`));
+
+          const filledSlots = await predictor.fillSlots(`${prefix} ${query}`, {
```
trs, line 190: Move this back to the `predict` call.

lib/services/dialog/index.ts
```diff
@@ -177,14 +176,30 @@ class DialogManagement extends AbstractManager<{ utils: typeof utils }> implemen
               dmRequest: dmStateStore.intentRequest.payload,
               isTrained,
             },
-            intentClassificationSettings,
+            {
+              type: 'nlu',
+              params: { confidence: 0.6 },
```
trs, line 181: Is `version.platformData.settings.intentConfidence` still valid to use here?

trs (changes requested):

lib/services/dialog/index.ts
theProf: i wouldn't trust it as it's not a feature we currently explicitly support (storing both nlu + hybrid settings to switch between), you currently are expected to lose changes. Default (as in your PR) seems most solid right now

lib/services/classification/predictor.class.ts
```diff
@@ -317,7 +296,7 @@ export class Predictor extends EventEmitter {
       return nlcPrediction;
     }
 
-    const nluPrediction = this.props.isTrained && (await this.nlu(utterance, this.options));
+    const nluPrediction = this.props.isTrained && (await this.nlu(utterance, { limit: 10 }));
```
theProf, line 299: this should switch back to `this.options` with a default somewhere

lib/services/dialog/index.ts
```diff
@@ -155,9 +158,15 @@ class DialogManagement extends AbstractManager<{ utils: typeof utils }> implemen
 
       try {
         const prefix = dmPrefix(dmStateStore.intentRequest.payload.intent.name);
-        const { intentClassificationSettings, intents, isTrained, slots } = castToDTO(version, project);
+        const { intents, isTrained, slots } = castToDTO(version, project);
+
+        const scopedIntent = intents?.find((intent) => intent.name === dmStateStore.intentRequest!.payload.intent.name);
+        const scopedSlots = scopedIntent?.slots?.map((slot) => slots?.find((entity) => entity.key === slot.id));
```
theProf, line 164:
```suggestion
        const scopedSlots = scopedIntent?.slots?.map((slot) => slots?.find((entity) => entity.key === slot.id)).filter(Boolean);
```
then below you don't have to `slot?.name`

lib/services/dialog/index.ts
```diff
@@ -177,22 +186,42 @@ class DialogManagement extends AbstractManager<{ utils: typeof utils }> implemen
...
           );
-          dmPrefixedResult = getIntentRequest(await predictor.predict(`${prefix} ${query}`));
+
+          const prediction = await predictor.predict(`${prefix} ${query}`);
+
+          (prediction?.predictedSlots ?? []).forEach((predictedSlot) => {
+            const slot = scopedSlots?.find((scopedSlot) => scopedSlot?.name === predictedSlot.name);
+            if (!slot) return;
+
+            const existingEntity = slotFillingRequest.payload.entities.find((entity) => entity.name === slot.name);
```
theProf, line 207: pending size we may want to compute this once into a Map

lib/services/dialog/index.ts
trs: Typescript isn't smart enough to know that slot is not undefined here

lib/services/dialog/index.ts
trs: For sure, this code is gross and I think deserves a refactor

theProf: for later days

trs:
bors try

PR #836, opened 2024-06-03
[Merged by Bors] - fix: reduce entitlement lookup failure log level (PL-000)

trs:
This log isn't very helpful right now, since most workspaces will fail. We can change it back to `error` once our billing migration is complete.


theProf (approved):

trs:
bors merge

PR #843, opened 2024-06-13
[Merged by Bors] - feat: move rate limit before project lookup (PL-000) [bugfix]

trs:
The [rate limit `versionConsume` method](https://github.com/voiceflow/general-runtime/blob/master/lib/middlewares/rateLimit.ts#L32) uses the authorization header for these routes, so we don't need to look up the project and version first.


theProf (approved):

trs:
bors merge

trs:
bors retry

PR #844, opened 2024-06-14
[Merged by Bors] - feat: add compiled expression condition (PL-1045)

zhihil:
**Fixes or implements PL-1045**

Implements the Expression Condition for Message Step, as well as a small framework to write additional types of conditions.

#### Condition evaluation
We introduce `BaseCondition`, a parent class which is used to implement execution logic for conditions. It contain two principle methods:

1. `constructor` - Which saves a reference to the raw compiled `condition` data, the `runtime`, and `variables`. 
2. `evaluate()` - Which returns a `boolean` or `Promise<boolean>` representing the result of evaluating that condition.

Derived types of the `BaseCondition` are initialized by the factory method `createCondition()` in `lib/services/runtime/handlers/message/lib/conditions/condition.ts`, which dynamically initializes the correct subclass based on the given compiled condition's `type`.

In this PR, we only implement the `ExpressionCondition`. Evaluating LHS and RHS expressions is implement exactly like the IfV2 step. It has the following methods:

1. `evaluate()` - Determines if the expression is an AND expression (`matchAll = true`) or an `OR` expression (`matchAll = false`), then we call the helper method `.every()` or `.some()` respectively.
2. `.every()` - Evaluates every assertion in the expression condition using a V8 isolate and returns `true` if every assertion evaluates to `true`
3. `.some()` - Returns `true` if at least one assertion evaluates to `true`
4. `.compileJITAssertion()` - Converts the given condition assertion into a JS expression. The implementations of these expressions are identical to their implementation in IfV2. 

Please see the unit tests which tests every operator for correctness.

Additionally, the `ConditionIsolate` class wraps around the logic to initialize an `isolated-vm` sandbox to execute expressions for expression condition and code for the script condition.

The other conditions will be implemented in later PRs.

#### Variant selection
The `selectVariant()` is the main logic that chooses variants . This algorithm makes use of the `BaseCondition` class we described above and it works like so:
1. Partition the variants into conditioned variants and unconditioned variants
2. Check each conditioned variant and return the first matching variant if it exists.
3. Otherwise, randomly sample an unconditioned variant. 

#### Message step handler
The Message step handler has been refactored slightly. It now includes the following new logic:
1. Initialize `preprocessedVariants` by converting the `condition` property of each `variant` into an executable `BaseCondition`.
2. Pass `preprocessedVariants` into `selectVariant` to select the variant. 

#### Rationale
The message handler code was structured in this way to provide encapsulation.

1. `selectVariant()` does not care how conditions are evaluated, so it calls a polymorphic `.evaluate()` method which could run an expression, script, or prompt condition
2. Conditions do not care how they are used or who uses it. They are given their underlying `condition`, `variables` data to initialize the code sandbox, and a `log` callback to inform the caller of interesting events. 
3. Conditions are implemented as classes. It can be implemented as functions and interfaces, but classes were chosen because the syntax can be useful to communicate that OOP patterns like polymorphism is occurring. 

This approach hopefully produces clearer and more reusable code.

### Setup information

N/A

### Deployment Notes

N/A

### Related PRs

N/A

### Checklist


zhihil:
The actual amount of business logic changes is less than 500 lines. The Graphite warning (^) is flagged because of the tests I added, which hopefully aren't too difficult to review and understand.

zhihil:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=voiceflow_general-runtime&pullRequest=844) **Quality Gate failed**
> Failed conditions  [1 Security Hotspot](https://sonarcloud.io/project/security_hotspots?id=voiceflow_general-runtime&pullRequest=844&resolved=false&sinceLeakPeriod=true)
> 
> [See analysis details on SonarCloud](https://sonarcloud.io/dashboard?id=voiceflow_general-runtime&pullRequest=844)

This analysis is a false flag. It marked `Math.random()` as not being security-safe (cryptographically random I assume), but we are only using `Math.random()` to sample the variants which is not a security use-case.

lib/services/runtime/handlers/message/lib/conditions/conditionIsolate.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    this.context = await this.isolate.createContext();
+
+    await Promise.all(
+      Object.keys(this.variables).map(async (key) => {
+        await this.context!.global.set(key, this.variables[key], {
+          copy: true,
+        });
+      })
+    );
```
theProf, line 28:
```suggestion
    await Promise.all(
      Object.entries(this.variables).map(([key, value]) =>
        this.context?.global.set(key, value, {
          copy: true,
        })
      )
    );
```

1. you're already returning a promise, no need to `async/await`
2. try and avoid using `!` for typings. this may be from typescript not inferring that it is definitely not undefined after the `await` call, but either casting above, or adding a check, or etc, etc, should be done before the use of `!`
3. use `Object.entries` to get both key and value to save a second lookup of the value within the loop

lib/services/runtime/handlers/message/lib/selectVariant.ts
```diff
@@ -0,0 +1,35 @@
...
+  const conditionedVariants: VariantConditionPair[] = [];
+  const unconditionedVariants: VariantConditionPair[] = [];
+
+  variants.forEach((pair) => {
+    if (pair.condition) {
+      conditionedVariants.push(pair);
+    } else {
+      unconditionedVariants.push(pair);
+    }
+  });
```
theProf, line 20:
```suggestion
  const [conditionedVariants, unconditionedVariants]: VariantConditionPair[][] = variants.reduce(
    ([conditioned, unconditioned], pair) => {
      if (pair.condition) {
        conditionedVariants.push(pair);
      } else {
        unconditionedVariants.push(pair);
      }
      return [conditioned, unconditioned];
    },
    [[], []]
  );
```

immutable with reduce. i find it an odd anti-pattern to use `const` with an array that gets mutated. so the other option is to keep what you have and use `let` over `const` for understanding at a glance that the arrays are mutable.

utils/logError/logError.ts
```diff
@@ -0,0 +1,46 @@
...
+          break;
+        case LogLevel.DEBUG:
+          log.debug(errorMessage);
+          break;
+        case LogLevel.TRACE:
+          log.trace(errorMessage);
+          break;
+        default:
+          throw new Error('unknown log level recevied');
+      }
```
theProf, line 42: i've noticed this pattern twice, and wondering if it may be better to use a `Map|Object` to cut the repetition in invocation: (i'm no good at naming right now): 
```
// listed in order of most to least fatality
const levelToLog = {
  [LogLevel.FATAL]: log.fatal,
  [LogLevel.ERROR]: log.error,
  [LogLevel.WARN]: log.error,
  [LogLevel.INFO]: log.info,
  [LogLevel.DEBUG]: log.debug,
  [LogLevel.TRACE]: log.trace,
};

export const createErrorRaiser =
  (context: string): ErrorRaiser =>
  (message, options = {}) => {
    const { shouldLog = true, logLevel = LogLevel.ERROR } = options;

    const errorMessage = `[${context}]: ${message}`;

    if (shouldLog) {
      if (!(logLevel in levelToLog)) {
        throw new Error('unknown log level received');
      }
      levelToLog[logLevel](errorMessage);
    }

    return new Error(errorMessage);
  };```

File: utils/logError/logError.ts [hunk 4]
```diff
@@ -0,0 +1,46 @@
...
+          log.info(errorMessage);
+          break;
+        case LogLevel.DEBUG:
+          log.debug(errorMessage);
+          break;
+        case LogLevel.TRACE:
+          log.trace(errorMessage);
+          break;
+        default:
+          throw new Error('unknown log level recevied');
```
theProf, line 41: recevied -> received

lib/services/runtime/handlers/message/lib/conditions/expression.condition.ts
```diff
@@ -0,0 +1,108 @@
...
+import { BaseCondition } from './base.condition';
+import { ConditionIsolate } from './conditionIsolate';
+
+export class ExpressionCondition extends BaseCondition<CompiledExpressionCondition> {
+  private compileJITAssertion(assertion: CompiledConditionAssertion): string {
+    switch (assertion.operation) {
+      case ConditionOperation.CONTAINS:
+        return `String(${assertion.lhs}).toLowerCase().includes(String(${assertion.rhs}).toLowerCase())`;
+      case ConditionOperation.NOT_CONTAINS:
+        return `!(String(${assertion.lhs}).toLowerCase().includes(String(${assertion.rhs}).toLowerCase()))`;
```
theProf, line 12: instead of re-implementing `CONTAINS`, may be better suited to call `CONTAINS` and negate it

lib/services/runtime/handlers/message/lib/conditions/expression.condition.ts
```diff
@@ -0,0 +1,108 @@
...
+      case ConditionOperation.LESS_OR_EQUAL:
+        return `${assertion.lhs} <= ${assertion.rhs}`;
+      case ConditionOperation.LESS_THAN:
+        return `${assertion.lhs} < ${assertion.rhs}`;
+      case ConditionOperation.IS:
+        return `${assertion.lhs} == ${assertion.rhs}`;
+      case ConditionOperation.IS_NOT:
+        return `${assertion.lhs} != ${assertion.rhs}`;
+      case ConditionOperation.IS_EMPTY:
+        return `${assertion.lhs} == ''`;
```
theProf, line 30: i always forget when to use it, do we want these to be the triple equals? `===` and `!==`?

lib/services/runtime/handlers/message/lib/conditions/expression.condition.ts
```diff
@@ -0,0 +1,108 @@
...
+    return `${assertion.lhs} ${assertion.operation} ${assertion.rhs}`;
+  }
+
+  private async every(isolate: ConditionIsolate): Promise<boolean> {
+    this.log('--- evaluating expression, matchAll = true ---');
+
+    const assertionResults = await Promise.all(
+      this.condition.data.assertions.map((assert) => this.evaluateAssertion(isolate, assert))
+    );
+    const result = assertionResults.every((val) => val);
```
theProf, line 53:
```suggestion
    const result = assertionResults.every(Boolean);
```

theProf (approved):

lib/services/runtime/handlers/message/lib/conditions/expression.condition.ts
zhihil: The use of `==` appears to be deliberate, according to [Tyler H](https://voiceflowhq.slack.com/archives/CCF7MAGC8/p1718738916055769), especially because the original code did something like `lhs == 0`, which suggests a deliberate abuse of `==`'s implicit type coercion. 

I'm not a big fan of this either, but I've written it this way to maintain parity with the IfV2 step, as requested by product. I will try to refactor later to see if I can remove the `==`.

theProf: thanks for the explanation! appreciated

lib/services/runtime/handlers/message/message.handler.ts
```diff
@@ -0,0 +1,96 @@
...
+
+function getMessageData(runtime: Runtime, messageID: string) {
+  if (!runtime.version) {
+    throw raiseMessageHandlerError('Runtime was not loaded with a version');
+  }
+
+  /**
+   * !TODO! - Need to replace this `as Version` cast by refactoring `general-runtime` to use the
+   *          `Version` from the DTOs.
+   */
```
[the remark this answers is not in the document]
zhihil: Here is a ticket: https://linear.app/voiceflow/issue/PL-1173/replace-version-type-from-libs-in-general-runtime-with-the-version

Here is an associated project, to which we should add additional tickets:  https://linear.app/voiceflow/project/refactor-general-runtime-to-replace-its-existing-types-with-30f2c9bdc0a8/overview

lib/services/runtime/handlers/message/lib/selectVariant.ts
```diff
@@ -0,0 +1,35 @@
...
+  /* eslint-disable no-await-in-loop */
+  // eslint-disable-next-line no-restricted-syntax
+  for (const pair of conditionedVariants) {
+    const isMatching = await pair.condition!.evaluate();
+
+    if (isMatching) {
+      return pair.variant;
+    }
+  }
+  /* eslint-enable no-await-in-loop */
```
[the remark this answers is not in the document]
zhihil: I think it should be fine. I found this (rather heated) [GitHub thread](https://github.com/airbnb/javascript/issues/1271) concerning the rationale for this eslint rule. Airbnb's rationale is that there is no lightweight polyfill for this syntax on browsers, but this is a frontend use-case that doesn't appear to apply to us.

One user makes this comment that the eslint rule might be irrelevant to backend developers.
Screenshot 2024-06-24 at 3 48 57 PM

One of the maintainers counters by questioning if regenerator runtime has an actual use on the backend, to which the user replies with this example (which is very similar to my code):
Screenshot 2024-06-24 at 3 49 51 PM

The maintainer then concedes that using `await` in `for..of` might be useful:
Screenshot 2024-06-24 at 3 50 36 PM

My takeaway from this thread is that this eslint rule is a frontend best practice, that is inappropriately spliced into a backend development context.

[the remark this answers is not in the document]
theProf: that's a good breakdown, if it's only the front-end polyfill, then we should definitely get rid of that eslint rule for our backend only services, agreed!

lib/services/runtime/handlers/message/message.handler.ts
```diff
@@ -0,0 +1,96 @@
...
+      if (!chosenDiscriminator) {
+        throw raiseMessageHandlerError(
+          `could not resolve response step, missing variants list for channel='${currentChannel}', language='${currentLanguage}'`
+        );
+      }
+
+      const preprocessedVariants = chosenDiscriminator.map((variant) => ({
+        variant,
+        condition: variant.condition
+          ? createCondition(variant.condition, variables.getState(), (message: string) => runtime.trace.debug(message))
```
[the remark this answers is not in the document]
zhihil: Good point, this will avoid a separate heap allocation for each distinct condition.

PR #849, opened 2024-06-19
[Merged by Bors] - feat: move rate limit for public interact route (PL-000)

trs:
Instead of needing to look up the project to use the version ID for rate limiting, we can use the given project ID for rate limiting instead. This allows the rate limiting to be the first check, before any database lookups.


DecathectZero (approved):

trs:
bors merge

PR #853, opened 2024-06-25
[Merged by Bors] - feat: add script condition (PL-1045) 

zhihil:
**Fixes or implements PL-1045**

Adds the Script Condition for the Message Step, which allows users to execute JavaScript code to evaluate a condition.

Borrowed code from `function-lambda` and implemented a simplified wrapper for the `isolated-vm` sandbox. This wrapper expects user-code with the pattern:
```js
export default function main({ variables }) {
  return 'something';
}
```

The wrapper then imports this user code and calls it with the current runtime variable values, which is exposed as `variables` in the function. 

The user code should return some value (which `isolated-vm` believes is safe  to copy outside of the sandbox, e.g, non-executable content) and the `ScriptCondition` will convert that value into a `boolean` according to JS truthy/falsy rules. 

The resulting `boolean` is used as the evaluation result of the `ScriptCondition`

### Setup information

N/A

### Deployment Notes

N/A

### Related PRs

N/A

### Checklist

### DIFF
--- a/lib/services/runtime/handlers/message/lib/conditions/conditionIsolate.ts
+++ b/lib/services/runtime/handlers/message/lib/conditions/conditionIsolate.ts
@@ -9,7 +9,7 @@ export class ConditionIsolate {
 
   private static readonly userModuleName = `userModule`;
 
-  private static readonly userVariablesName = `inputVars`;
+  private static readonly userVariablesName = `variables`;
 
   private static readonly resolveFuncName = `__resolve__`;
 
--- a/lib/services/runtime/handlers/message/lib/conditions/conditionIsolate.ts
+++ b/lib/services/runtime/handlers/message/lib/conditions/conditionIsolate.ts
@@ -22,6 +22,12 @@ export class ConditionIsolate {
     maxExecutionTimeMs: 1 * 1000,
   };
 
+  /**
+   * Given the `resolve` and `reject` function instantiated by a `Promise` constructor, injects
+   * these functions into the `isolated-vm` environment as a proxied function. The `isolated-vm`
+   * entrypoint module will propagate values outside the sandbox through the `resolve` function,
+   * it does the same for thrown exceptions with the `reject` function
+   */
   private async setupResolve(resolve: (value: any) => void, reject: (reason?: any) => void) {
     await this.context!.evalClosure(
       `
@@ -37,15 +43,30 @@ export class ConditionIsolate {
     );
   }
 
+  /**
+   * Injects all given user variables into the execution environment.
+   */
   private async setupUserVariables() {
     await this.context!.global.set(ConditionIsolate.userVariablesName, this.variables, { copy: true });
   }
 
+  /**
+   * Compiles user-submitted code into an ESM module, which is later imported and executed by the
+   * entrypoint module
+   * @param code User-submitted code
+   * @returns `ivm.Module` containing the user code
+   */
   private async compileUserModule(code: string) {
     return this.isolate!.compileModule(code);
   }
 
+  /**
+   * Compiles the entrypoint module, which imports the default member of the user code module, executes
+   * it, and propagates the return value outside the sandbox using `resolve()`.
+   * @returns
+   */
   private async compileMainModule() {
+    // Removes injected user variables from the `global` object to keep this namespace clean.
     const sanitizeVarsFromGlobal = `
             const ${ConditionIsolate.userVariablesName} = globalThis.${ConditionIsolate.userVariablesName};
             delete globalThis.${ConditionIsolate.userVariablesName};
@@ -93,12 +114,15 @@ export class ConditionIsolate {
 
     const executeCode = async (resolve: (val: unknown) => void, reject: (reason?: unknown) => void) => {
       try {
+        // inject the `Promise`'s resolve and reject functions as proxied functions in the sandbox
         await this.setupResolve(resolve, reject);
 
+        // forbid user-submitted code from importing any modules
         await userModule.instantiate(this.context!, () => {
           throw new Error(`User code cannot import modules.`);
         });
 
+        // allow entrypoint module to import the user-submitted code.
         await mainModule.instantiate(this.context!, (specifier: string) => {
           if (specifier === ConditionIsolate.userModuleName) {
             return userModule;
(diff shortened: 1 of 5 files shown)


lib/services/runtime/handlers/message/lib/conditions/script.condition.ts
```diff
@@ -1,13 +1,37 @@
...
-  private evaluateCode(_code: string): boolean {
-    throw new Error('script condition is not implemented');
+  private async evaluateCode(isolate: ConditionIsolate, code: string): Promise<boolean> {
+    const result = await isolate.executeUserModule(code);
+    return !!result;
   }
 
-  public evaluate(): boolean {
-    return this.evaluateCode(this.condition.data.code);
+  public async evaluate(): Promise<boolean> {
```
zhihil, line 12: It is best to start your review from here. This `.evaluate()` function is shared with all conditions and is called to determine the boolean value of any message step condition.

This is the polymorphic implementation for the script condition.

lib/services/runtime/handlers/message/lib/conditions/conditionIsolate.ts
```diff
@@ -5,13 +5,105 @@ export class ConditionIsolate {
...
+  private static readonly userFunctionName = `userFunction`;
+
+  private static readonly userModuleName = `userModule`;
+
+  private static readonly userVariablesName = `inputVars`;
+
+  private static readonly resolveFuncName = `__resolve__`;
+
+  private static readonly rejectFuncName = `__reject__`;
+
```
pmvrmc, line 17: maybe better in a const file?

lib/services/runtime/handlers/message/lib/conditions/script.condition.ts
```diff
@@ -1,13 +1,37 @@
...
 
 import { BaseCondition } from './base.condition';
+import { ConditionIsolate } from './conditionIsolate';
 
 export class ScriptCondition extends BaseCondition<CompiledScriptCondition> {
-  private evaluateCode(_code: string): boolean {
-    throw new Error('script condition is not implemented');
+  private async evaluateCode(isolate: ConditionIsolate, code: string): Promise<boolean> {
+    const result = await isolate.executeUserModule(code);
+    return !!result;
```
pmvrmc, line 9: wont this throw if execute fails/rejects?
so the value if returned is always true?

if so we can simplify
```
await isolate.executeUserModule(code);
return true
```

zhihil: The `result` is the return value that the user provides in their script condition function:

```js
export default function main() {
   return "hello, world";
}
```

So the code becomes this:

```ts
const result = "hello, world";
return !!result;
```

If the user returns a falsy value `result` will be false.

zhihil: It may be worth pairing to understand the code, because you are forced to do something unusual with `isolated-vm` when working with ESM modules.

lib/services/runtime/handlers/message/lib/conditions/conditionIsolate.ts
```diff
@@ -29,6 +121,55 @@ export class ConditionIsolate {
...
+        if (!this.context) {
+          throw new Error(`condition isolate was not initialized before an attempt to execute code`);
+        }
+
+        // inject the `Promise`'s resolve and reject functions as proxied functions in the sandbox
+        await this.setupResolve(resolve, reject);
+
+        // forbid user-submitted code from importing any modules
+        await userModule.instantiate(this.context, () => {
+          throw new Error(`User code cannot import modules.`);
```
pmvrmc, line 139: its a bit weird that we are wrapping this inside a promise, but on some occasions we throw an error, on others we reject the promise. Both endup rejecting the promise, maybe we should stick to one nomenclature?

lib/services/runtime/handlers/message/lib/conditions/conditionIsolate.ts
```diff
@@ -29,6 +121,55 @@ export class ConditionIsolate {
...
+        executeCode(resolve, reject);
+      } catch (err) {
+        if (err instanceof Error) {
+          reject(err);
+        } else {
+          reject(new Error(JSON.stringify(err)));
+        }
+      }
+    });
+  }
```
pmvrmc, line 171: could we forego the try/catches since we are going with promise approach?

this part could be just
```
return executeCode.catch((err) => {
    if (err instanceof Error) {
       throw err;
     }
     throw new Error(JSON.stringify(err));
   });
```

and the executeCode function could create the promise itself

lib/services/runtime/handlers/message/lib/conditions/script.condition.ts
pmvrmc: lets pair :) 
we can do it during our donut, would that work for you?

lib/services/runtime/handlers/message/lib/conditions/conditionIsolate.ts
zhihil: Based on our pair programming discussion, the simplest we can make this code is:
```ts
// eslint-disable-next-line no-async-promise-executor
return new Promise(async (resolve, reject) => {
  try {
    // ... executeCode()'s logic should be copy-pasted in here 
  } catch (err) {
    // Need this otherwise `Promise` doesn't reject if the `async` function rejects. This is the error that
    // `no-async-promise-executor` is trying to protect developers against. This may or may not work.
    reject(err);
  }
});
```

pmvrmc (approved):

lib/services/runtime/handlers/message/lib/conditions/conditionIsolate.ts
zhihil: As discussed in the pair meeting, it would be a cosmetic change. I think keeping it inside of the class has the very small benefit of keeping the IntelliSense namespace clean.

lib/services/runtime/handlers/message/lib/conditions/conditionIsolate.ts
zhihil: Reverted to the original code (but removed a `catch` clause) to avoid tripping ESLint and SonarCloud.

PR #863, opened 2024-07-04
feat: move runtime to turn handler context (PL-000)

trs:
Currently, we don't have context to your diagram/current node until we reach the runtime turn handler. This change moves the loading of the "runtime" component (not turn handler), to the turn "context" instead. This lets us check the runtime during any turn handler. This is then used to determine if we should skip the NLU/dialog management turn handlers if we're on a new listen step.
The goal here being to reduce the scope of changes needed, as well as reducing the risk of changing old behaviour.

### DIFF
--- a/lib/services/dialog/index.ts
+++ b/lib/services/dialog/index.ts
@@ -21,7 +21,7 @@ import { Context, ContextHandler, VersionTag } from '@/types';
 import { Predictor } from '../classification';
 import { castToDTO } from '../classification/classification.utils';
 import { getIntentRequest } from '../nlu';
-import { getNoneIntentRequest } from '../nlu/utils';
+import { getNoneIntentRequest, shouldBypassNLU } from '../nlu/utils';
 import { isIntentRequest, StorageType } from '../runtime/types';
 import { addOutputTrace, getOutputTrace } from '../runtime/utils';
 import { AbstractManager, injectServices } from '../utils';
@@ -120,6 +120,8 @@ class DialogManagement extends AbstractManager<{ utils: typeof utils }> implemen
     if (!isIntentRequest(context.request)) {
       return context;
     }
+
+    if (await shouldBypassNLU(context)) return context;
 
     const version = await context.data.api.getVersion(context.versionID);
 
--- a/lib/services/nlu/utils.ts
+++ b/lib/services/nlu/utils.ts
@@ -1,14 +1,10 @@
 import { AlexaConstants } from '@voiceflow/alexa-types';
-import { BaseModels, BaseNode, BaseRequest } from '@voiceflow/base-types';
-import { CommandType, EventType } from '@voiceflow/base-types/build/cjs/node/utils';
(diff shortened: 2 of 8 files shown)


lib/services/dialog/index.ts
```diff
@@ -121,6 +121,8 @@ class DialogManagement extends AbstractManager<{ utils: typeof utils }> implemen
       return context;
     }
 
+    if (await shouldBypassNLU(context)) return context;
```
trs, line 124: Bad naming and such, but illustrates the idea.

lib/services/nlu/utils.ts
```diff
@@ -94,107 +90,18 @@ export const mapChannelData = (data: any, platform?: VoiceflowConstants.Platform
...
-    intentInteractions
-      .flatMap((interaction) => interaction.event.mappings)
-      .flatMap((mapping) => mapping?.slot || [])
-      .forEach((entity) => nodeInteractionEntityNames.add(entity));
-  }
-
-  return { nodeInteractionIntentNames, nodeInteractionEntityNames };
-};
-
-export const getAvailableIntentsAndEntities = async (
```
trs, line 140: This was an attempt at scoping intents, it didn't work and was unused. Removing it here to avoid confusion. But it does outline the problem before, a new runtime instance needed to be created to check the project.

lib/services/nlu/utils.ts
```diff
@@ -94,107 +90,18 @@ export const mapChannelData = (data: any, platform?: VoiceflowConstants.Platform
...
-      availableEntities: setIntersect(commandEntityNames, nodeInteractionEntityNames),
-    };
-  }
-
-  return {
-    availableIntents: commandIntentNames,
-    availableEntities: commandEntityNames,
-  };
+  // TODO: update with actual node type checks, probably via DTOs
+  return node?.type === 'interaction';
```
trs, line 106: Like the TODO states, this would be where we'd check for a new listen step.

lib/services/runtime/index.ts
```diff
@@ -97,37 +79,15 @@ class RuntimeManager extends AbstractManager<{ utils: typeof utils }> implements
...
+      runtime,
       request,
       userID,
       versionID,
       state: runtime.getFinalState(),
       trace: runtime.trace.get(),
     };
   }
-
-  private getRuntimeForContext(context: Context, eventHandler: HandleContextEventHandler): Runtime {
```
trs, line 108: Runtime creation is moved out of the runtime handler, into the turn state context.

lib/services/state/index.ts
```diff
@@ -163,6 +166,47 @@ class StateManager extends AbstractManager<{ utils: typeof utils }> implements I
...
     };
+
+    const client = this.createClient(api);
+    const runtime = this.getRuntimeForContext(client, updatedContext);
+
+    return {
+      ...updatedContext,
+      client,
+      runtime,
+    };
```
trs, line 177: A bit hacky, cause we need the context to create the runtime, but the runtime is part of the context.

PR #864, opened 2024-07-05
[Merged by Bors] - fix: missing variable substitution (PL-000)

zhihil:

Add variable substitution of Voiceflow variables in expression condition.

### Related PRs

### Checklist

### DIFF
--- a/lib/services/runtime/handlers/message/lib/conditions/base.condition.ts
+++ b/lib/services/runtime/handlers/message/lib/conditions/base.condition.ts
@@ -1,3 +1,4 @@
+import { sanitizeVariables } from '@voiceflow/common';
 import { AnyCompiledCondition } from '@voiceflow/dtos';
 import noop from 'lodash/noop';
 
@@ -8,7 +9,9 @@ export abstract class BaseCondition<Condition extends AnyCompiledCondition = Any
     protected readonly condition: Condition,
     protected readonly variables: Record<string, unknown>,
     protected readonly log: BaseConditionLogger = noop
-  ) {}
+  ) {
+    this.variables = sanitizeVariables(variables);
+  }
 
   abstract evaluate(): Promise<boolean> | boolean;
 }
--- a/lib/services/runtime/handlers/message/lib/conditions/expression.condition.ts
+++ b/lib/services/runtime/handlers/message/lib/conditions/expression.condition.ts
@@ -1,3 +1,4 @@
+import { replaceVariables } from '@voiceflow/common';
 import { CompiledConditionAssertion, CompiledExpressionCondition, ConditionOperation } from '@voiceflow/dtos';
 
 import { BaseCondition } from './base.condition';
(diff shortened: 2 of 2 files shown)


lib/services/runtime/handlers/message/lib/conditions/base.condition.ts
```diff
@@ -8,7 +9,9 @@ export abstract class BaseCondition<Condition extends AnyCompiledCondition = Any
     protected readonly condition: Condition,
     protected readonly variables: Record<string, unknown>,
     protected readonly log: BaseConditionLogger = noop
-  ) {}
+  ) {
+    this.variables = sanitizeVariables(variables);
```
zhihil, line 13: All conditions receive variables, so we'll apply this `sanitizeVariables` utility to the newly received variables as soon as the condition is created.

lib/services/runtime/handlers/message/lib/conditions/expression.condition.ts
```diff
@@ -1,9 +1,15 @@
+import { replaceVariables } from '@voiceflow/common';
 import { CompiledConditionAssertion, CompiledExpressionCondition, ConditionOperation } from '@voiceflow/dtos';
 
 import { BaseCondition } from './base.condition';
 import { ConditionIsolate } from './conditionIsolate';
 
 export class ExpressionCondition extends BaseCondition<CompiledExpressionCondition> {
+  private parseString(value: unknown) {
+    if (typeof value !== 'string') return value;
+    return `"${value.replace(/"/g, '\\"')}"`;
```
zhihil, line 10: String values needed to be quoted so that they're interpreted properly by `isolated-vm`

lib/services/runtime/handlers/message/lib/conditions/expression.condition.ts
```diff
@@ -52,7 +58,12 @@ export class ExpressionCondition extends BaseCondition<CompiledExpressionConditi
   }
 
   private async evaluateAssertion(isolate: ConditionIsolate, assertion: CompiledConditionAssertion): Promise<boolean> {
-    const result: unknown = await isolate.executeCode(this.compileJITAssertion(assertion));
+    const resolvedAssertion = {
+      ...assertion,
+      lhs: replaceVariables(assertion.lhs, this.variables, this.parseString),
+      rhs: replaceVariables(assertion.rhs, this.variables, this.parseString),
```
zhihil, line 64: Go through each variable and replace it with the variable value. The `parseString` callback is used to quote string values so that they can be evaluated properly as JS code.

lib/services/runtime/handlers/message/lib/conditions/expression.condition.ts
theProf: does this need to happen for any other characters? ex: `\n`? or `"\n"`?

trs: Looks like `JSON.stringify` does this, unsure if it's the best option or if it'll come with other weird side effects
image

trs (approved):

lib/services/runtime/handlers/message/lib/conditions/expression.condition.ts
zhihil: Let me take a look at `JSON.stringify()` before I merge this in. We might have to apply the name transformation to all special characters if the string variable was assigned  a value containing `'\n'`, e.g, the string value comes from API response.

zhihil: Using `JSON.stringify()` may be the best choice. As we'll likely need to escape `'\n'` as Greg pointed out, using `JSON.stringify()` has more benefits than known cons.

PR #866, opened 2024-07-05
[Merged by Bors] - feat: add trace time property (ENG-172)

DecathectZero:
add a `time` property with `Date.now()` for all traces.

Right now the time property is optional in `BaseTraceFrame`,  however for the purpose of the PR I intentionally modified my local type to have `time` to be required and fixed all resulting type and build errors. This way we can remove the optional field and have it fully be required.

Reference my comments for more details. The majority of traces just use `runtime.trace.addTrace` so I made it automatically assign the time there, and have the time field optional so other places can be less verbose.

### DIFF
--- a/lib/services/tts/index.ts
+++ b/lib/services/tts/index.ts
@@ -15,12 +15,12 @@ export const utils = {};
 
 @injectServices({ utils })
 class TTS extends AbstractManager<{ utils: typeof utils }> implements ContextHandler {
-  fetchTTS = async (node: BaseNode.Speak.TraceFrame, locale?: string): Promise<BaseTrace.SpeakTrace[]> => {
+  fetchTTS = async (trace: BaseNode.Speak.TraceFrame, locale?: string): Promise<BaseTrace.SpeakTrace[]> => {
     try {
       const { data } = await this.services.axios.post<BaseTrace.SpeakTrace['payload'][]>(
         `${this.config.GENERAL_SERVICE_ENDPOINT}/tts/convert`,
         {
-          ssml: node.payload.message,
+          ssml: trace.payload.message,
         },
         {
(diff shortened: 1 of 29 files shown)


lib/services/tts/index.ts
```diff
@@ -15,28 +15,30 @@ export const utils = {};
 
 @injectServices({ utils })
 class TTS extends AbstractManager<{ utils: typeof utils }> implements ContextHandler {
-  fetchTTS = async (node: BaseNode.Speak.TraceFrame, locale?: string): Promise<BaseTrace.SpeakTrace[]> => {
+  fetchTTS = async (trace: BaseNode.Speak.TraceFrame, locale?: string): Promise<BaseTrace.SpeakTrace[]> => {
```
DecathectZero, line 18: it just renamed `node` to `trace` here (not idea why it was called node before), and do a additional `...trace` spread, so that the time property is carried over.

runtime/lib/Handlers/function/lib/execute-function/lib/validate-trace.ts
```diff
@@ -5,6 +5,10 @@ import { SimpleTraceDTO } from '../../../runtime-command/trace-command.dto';
 import { isSimpleTraceType } from './is-simple-trace';
 
 export function parseTrace(trace: UnknownTrace): UnknownTrace {
+  // force all function declared traces to have a timestamp
+  // eslint-disable-next-line no-param-reassign
+  if (!trace.time) trace.time = Date.now();
+
```
DecathectZero, line 11: for incoming function traces, we can't expect the users to also declare a time, we can do it automatically for them

trs (approved):

PR #874, opened 2024-07-15
[Merged by Bors] - feat: treat unused intent predictions as None (PL-1294)

trs:
With the introduction of the Intent CMS, we changed so every intent is trained. This means we can trigger an intent, but it won't do anything. It also means if they have a `None` intent trigger, it won't be triggered.

We still keep track of used intents in the survey context, so if we have it, we use it to treat unused intents as `None`.

### DIFF
--- a/lib/services/dialog/index.ts
+++ b/lib/services/dialog/index.ts
@@ -187,6 +187,7 @@ class DialogManagement extends AbstractManager<{ utils: typeof utils }> implemen
 
           const prediction = await predictor.predict(`${prefix} ${query}`);
 
+          // LATER: look into using the `filteredIntents` field in the Predictor class instead
           dmPrefixedResult =
             prediction && isUsedIntent(version.prototype?.surveyorContext.usedIntentsSet, prediction.predictedIntent)
               ? getIntentRequest(prediction)
--- a/lib/services/dialog/index.ts
+++ b/lib/services/dialog/index.ts
@@ -21,7 +21,7 @@ import { Context, ContextHandler, VersionTag } from '@/types';
 import { Predictor } from '../classification';
 import { castToDTO } from '../classification/classification.utils';
 import { getIntentRequest } from '../nlu';
-import { getNoneIntentRequest, isUsedIntent } from '../nlu/utils';
+import { findIntent, getNoneIntentRequest, isUsedIntent } from '../nlu/utils';
 import { isIntentRequest, StorageType } from '../runtime/types';
 import { addOutputTrace, getOutputTrace } from '../runtime/utils';
 import { AbstractManager, injectServices } from '../utils';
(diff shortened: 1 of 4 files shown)


lib/services/nlu/utils.ts
```diff
@@ -94,107 +88,9 @@ export const mapChannelData = (data: any, platform?: VoiceflowConstants.Platform
...
-    intentInteractions
-      .flatMap((interaction) => interaction.event.mappings)
-      .flatMap((mapping) => mapping?.slot || [])
-      .forEach((entity) => nodeInteractionEntityNames.add(entity));
-  }
-
-  return { nodeInteractionIntentNames, nodeInteractionEntityNames };
-};
-
-export const getAvailableIntentsAndEntities = async (
```
trs, line 140: This and the above functions have been long usused. Removing it as it can be confusing to see.

lib/services/dialog/index.ts
```diff
@@ -184,7 +184,13 @@ class DialogManagement extends AbstractManager<{ utils: typeof utils }> implemen
               platform: version.prototype.platform as VoiceflowConstants.PlatformType,
             }
           );
-          dmPrefixedResult = getIntentRequest(await predictor.predict(`${prefix} ${query}`));
+
+          const prediction = await predictor.predict(`${prefix} ${query}`);
+
+          dmPrefixedResult =
+            prediction && isUsedIntent(version.prototype?.surveyorContext.usedIntentsSet, prediction.predictedIntent)
```
theProf, line 191: should the `predictor` own this or the caller should have the context?

theProf (approved):

trs:
bors merge

trs:
bors merge

trs:
bors retry

lib/services/nlu/utils.ts
```diff
@@ -94,107 +88,12 @@ export const mapChannelData = (data: any, platform?: VoiceflowConstants.Platform
...
-  return {
-    availableIntents: commandIntentNames,
-    availableEntities: commandEntityNames,
-  };
-};
+export const isUsedIntent = (usedIntents: string[] | undefined, intent: { key: string; name: string } | undefined) =>
+  /* If we don't have a used intents array, consider it "used" for compatibility */
+  !Array.isArray(usedIntents) ||
+  (!!intent &&
+    /* The used intent set contains both keys and names :confused: */
```
trs, line 95: Examples from general-service surveyor:
[Using intent name](https://github.com/voiceflow/general-service/blob/d7655355092c1cfad4a8a30a1a8b86da1fb49790/lib/clients/platform/handlers/interaction/survey.ts#L12)
[Using intent key/ID](https://github.com/voiceflow/general-service/blob/d7655355092c1cfad4a8a30a1a8b86da1fb49790/lib/clients/platform/handlers/interaction/survey.ts#L18)

trs:
bors try

trs:
bors try

PR #887, opened 2024-07-26
[Merged by Bors] - fix: add additional chaining to locales lookup (PL-1347)

trs:
Seeing some weird errors around accessing this `locales` property, even though it exists on the project data. Adding these optional chaining as a stop-gap while we investigate.


theProf (approved):

trs:
bors merge

PR #888, opened 2024-07-27
fix: allow production instances no unleash config (ENG-208)

DecathectZero:
If no env vars are set:
https://github.com/voiceflow/general-runtime/blob/5aa2cef79f5c519014b376f29d17f048655e9d6b/lib/clients/unleash/index.ts#L54-L58

instance never gets initialized, so we should not throw an error.


DecathectZero:
bors try

PR #890, opened 2024-07-30
fix: force steps to use strict vm (ENG-209)

DecathectZero:
we introduced a bit of a regression with this:
https://github.com/voiceflow/general-runtime/pull/776

This was the standard state before that PR:
https://github.com/voiceflow/general-runtime/blob/48cba2d4ac54b3d5da745111587d592cb96b66a6/runtime/lib/Handlers/ifV2.ts#L38

Right now the IfV2 and setV2 steps are also going into the code execute lambda and using VM2, we want it to use isolated-vm on general-runtime itself.

### DIFF
--- a/runtime/lib/Handlers/code/index.ts
+++ b/runtime/lib/Handlers/code/index.ts
@@ -11,13 +11,18 @@ import * as utils from './utils';
 export interface CodeOptions {
   endpoint?: string | null;
   callbacks?: Record<string, (...args: any) => any>;
+  useStrictVM?: boolean;
 }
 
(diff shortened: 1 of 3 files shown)


runtime/lib/Handlers/code/index.ts
```diff
@@ -11,13 +11,18 @@ import * as utils from './utils';
 export interface CodeOptions {
   endpoint?: string | null;
   callbacks?: Record<string, (...args: any) => any>;
+  useStrictVM?: boolean;
```
DecathectZero, line 14: this was already the established pattern

PR #896, opened 2024-08-01
[Merged by Bors] - refactor: resolve to false if exception throws an error (PL-1360)

zhihil:
**Fixes or implements PL-1360**

Changing the message conditions so that if an exception occurs when evaluating the condition, we resolve the condition to `false` instead of throwing an exception. 

Changed the Condition classes to accept a new callback called `logError`. We use this callback to receive error message strings from the condition and log these errors to DataDog. 

The old `log` callback passed to Condition classes was renamed to `emitTraceMessage` (no relation to DM API traces, it means "tracing" through the evaluation logic of the condition classes). The rationale here is that "logging" should specifically refer to _our_ logging to DataDog for observability, while the "logging" on the Prototype Tool is a customer-facing feature intended to help them debug their conditions. 

We need two different callbacks because error messages on DataDog vs. error messages on Prototype Tool have different audiences, us vs. the users. Therefore, the content of the error messages we log to DataDog vs. the Prototype Tool need to be different. 

Changed the Condition classes' `catch`-all clause to `return false` rather than throw an `Error`. 

### Setup information

N/A

### Deployment Notes

N/A

### Related PRs

N/A

### Checklist

- ~~[ ] Appropriate tests have been written~~

### DIFF
--- a/lib/services/runtime/handlers/message/lib/conditions/expression.condition.ts
+++ b/lib/services/runtime/handlers/message/lib/conditions/expression.condition.ts
@@ -67,38 +67,38 @@ export class ExpressionCondition extends BaseCondition<CompiledExpressionConditi
   }
 
   private async every(isolate: ConditionIsolate): Promise<boolean> {
-    this.logToPrototype('--- evaluating expression, matchAll = true ---');
+    this.log('--- evaluating expression, matchAll = true ---');
 
     const assertionResults = await Promise.all(
       this.condition.data.assertions.map((assert) => this.evaluateAssertion(isolate, assert))
     );
     const result = assertionResults.every(Boolean);
 
-    this.logToPrototype(`--- evaluated expression, result = ${result} ---`);
+    this.log(`--- evaluated expression, result = ${result} ---`);
 
     if (!result) {
       const firstFalse = assertionResults.findIndex((val) => !val);
       const assertion = this.condition.data.assertions[firstFalse];
-      this.logToPrototype(`- assertion '${this.formatAssertion(assertion)}' was false`);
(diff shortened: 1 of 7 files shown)


lib/services/runtime/handlers/message/lib/conditions/expression.condition.ts
```diff
@@ -114,15 +114,22 @@ export class ExpressionCondition extends BaseCondition<CompiledExpressionConditi
...
-        `an unknown error occurred executing an expression condition, details = ${JSON.stringify(
-          err,
-          null,
-          2
-        ).substring(0, 300)}`
-      );
+
+      this.log(`expression condition encountered an unexpected error and automatically resolved to 'false'`);
+
+      return false;
```
zhihil, line 132: Main Change (1/2)

We no longer throw a formatted `Error` here and instead return `false`. The error messages will still be logged to DataDog.

lib/services/runtime/handlers/message/lib/conditions/script.condition.ts
```diff
@@ -22,14 +22,19 @@ export class ScriptCondition extends BaseCondition<CompiledScriptCondition> {
...
-      throw new Error(
-        `an unknown error occurred executing a script condition, details = ${JSON.stringify(err, null, 2).substring(
-          0,
-          300
-        )}`
-      );
+
+      this.log(`script condition encountered an unexpected error and automatically resolved to 'false'`);
+
+      return false;
```
zhihil, line 37: Main Change (2/2)

We no longer throw a formatted `Error` here and instead return `false`. The error messages will still be logged to DataDog.

trs (approved):

PR #898, opened 2024-08-02
[Merged by Bors] - feat: update error messages (COR-2871)

zhihil:
**Fixes or implements COR-2871**

Rework the error messaging for the Message step. 

Additionally, this PR reworks the Script Condition execution to fit with product's new requirements that the script condition shouldn't require the user to include a `return` statement, that is, all of these are valid and equivalent script condition code:
```js
const hello = 1;
1 + 2 == 3 + hello;
```
```js
const hello = 1;
return 1 + 2 == 3 + hello;
```

See notes

### Setup information

N/A

### Deployment Notes

N/A

### Related PRs

### Checklist

### DIFF
--- a/lib/services/runtime/handlers/message/lib/conditions/conditionIsolate.ts
+++ b/lib/services/runtime/handlers/message/lib/conditions/conditionIsolate.ts
@@ -31,10 +31,24 @@ export class ConditionIsolate {
     this.isolate?.dispose();
   }
 
-  public async executeFunction(code: string): Promise<unknown> {
+  public async executeFunctionOrScript(code: string): Promise<unknown> {
     const functionBody = `
       const { ${ConditionIsolate.userVariablesName} } = $0;
-      ${code}
+      try {
+        // evaluate 'code' as though it's a top-level script and get last evaluated value
+        return eval(${JSON.stringify(code)});
+      } catch (err) {
+        // evaluate 'code' as though it's the function body and execute it within a function
+        if (err.message.includes('Illegal return statement')) {
+          return eval(\`
+            function main() {
+              ${code}
+            }
+            main()
+          \`)
+        }
+        throw err;
+      }
     `;
 
     return this.context?.evalClosure(functionBody, [{ [ConditionIsolate.userVariablesName]: this.variables }], {
--- a/lib/services/runtime/handlers/message/lib/conditions/expression.condition.ts
+++ b/lib/services/runtime/handlers/message/lib/conditions/expression.condition.ts
@@ -105,23 +105,17 @@ export class ExpressionCondition extends BaseCondition<CompiledExpressionConditi
       //           executes before the evaluated condition is fully computed.
       return await (this.condition.data.matchAll ? this.every(isolate) : this.some(isolate));
     } catch (err) {
-      if (err instanceof Error) {
-        this.logError(
-          `an unknown error occurred executing an expression condition, msg = ${err.message.substring(0, 200)}`
-        );
-      } else {
-        this.logError(
-          `an unknown error occurred executing an expression condition, details = ${JSON.stringify(
-            err,
-            null,
-            2
-          ).substring(0, 200)}`
-        );
-      }
+      const errMessage =
+        err instanceof Error
+          ? `Expression condition encountered an error and automatically resolved to 'false'. Details: ${err.message}`
+          : `Expression condition encountered an unexpected error and automatically resolved to 'false'. Details: ${JSON.stringify(
+              err,
+              null,
+              2
+            ).substring(0, 300)}`;
 
-      this.emitTraceMessage(
-        `Expression condition encountered an unexpected error and automatically resolved to 'false'`
-      );
+      this.logError(errMessage);
+      this.emitTraceMessage(errMessage);
 
       return false;
     } finally {
--- a/lib/services/runtime/handlers/message/lib/conditions/expression.condition.ts
+++ b/lib/services/runtime/handlers/message/lib/conditions/expression.condition.ts
@@ -67,38 +67,30 @@ export class ExpressionCondition extends BaseCondition<CompiledExpressionConditi
   }
 
   private async every(isolate: ConditionIsolate): Promise<boolean> {
-    this.emitTraceMessage('--- evaluating expression, matchAll = true ---');
-
     const assertionResults = await Promise.all(
       this.condition.data.assertions.map((assert) => this.evaluateAssertion(isolate, assert))
     );
     const result = assertionResults.every(Boolean);
 
-    this.emitTraceMessage(`--- evaluated expression, result = ${result} ---`);
-
     if (!result) {
       const firstFalse = assertionResults.findIndex((val) => !val);
       const assertion = this.condition.data.assertions[firstFalse];
-      this.emitTraceMessage(`- assertion '${this.formatAssertion(assertion)}' was false`);
+      this.emitTraceMessage(`Expression was false because '${this.formatAssertion(assertion)}' was false`);
     }
 
     return result;
   }
 
   private async some(isolate: ConditionIsolate): Promise<boolean> {
-    this.emitTraceMessage('--- evaluating expression, matchAll = false ---');
-
     const assertionResults = await Promise.all(
       this.condition.data.assertions.map((assert) => this.evaluateAssertion(isolate, assert))
     );
     const result = assertionResults.some(Boolean);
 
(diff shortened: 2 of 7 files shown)


lib/services/runtime/handlers/message/lib/conditions/conditionIsolate.ts
```diff
@@ -31,10 +31,24 @@ export class ConditionIsolate {
...
+        if (err.message.includes('Illegal return statement')) {
+          return eval(\`
+            function main() {
+              ${code}
+            }
+            main()
+          \`)
+        }
+        throw err;
+      }
```
zhihil, line 51: This code was reworked because product wants the `return` statement to be optional in the Script condition.

The `try` branch attempts to evaluate the condition code as though it were a top-level script. 

If instead, the code has a `return` statement, the `try` branch fails because `eval()` contains about an "Illegal return statement".

We catch the "Illegal return statement" error and attempt to run the code as a function body instead.

lib/services/runtime/handlers/message/lib/conditions/expression.condition.ts
```diff
@@ -67,38 +67,30 @@ export class ExpressionCondition extends BaseCondition<CompiledExpressionConditi
   }
 
   private async every(isolate: ConditionIsolate): Promise<boolean> {
-    this.emitTraceMessage('--- evaluating expression, matchAll = true ---');
```
zhihil, line 70: Removing some of the trace messages to avoid creating too many debug messages at once.

lib/services/runtime/handlers/message/lib/conditions/expression.condition.ts
```diff
@@ -67,38 +67,30 @@ export class ExpressionCondition extends BaseCondition<CompiledExpressionConditi
...
       this.condition.data.assertions.map((assert) => this.evaluateAssertion(isolate, assert))
     );
     const result = assertionResults.every(Boolean);
 
-    this.emitTraceMessage(`--- evaluated expression, result = ${result} ---`);
-
     if (!result) {
       const firstFalse = assertionResults.findIndex((val) => !val);
       const assertion = this.condition.data.assertions[firstFalse];
-      this.emitTraceMessage(`- assertion '${this.formatAssertion(assertion)}' was false`);
```
zhihil, line 82: Remove the `--` prefix in the debug messages and making it into a pure English sentence.

lib/services/runtime/handlers/message/lib/conditions/expression.condition.ts
```diff
@@ -113,23 +105,17 @@ export class ExpressionCondition extends BaseCondition<CompiledExpressionConditi
...
+        err instanceof Error
+          ? `Expression condition encountered an error and automatically resolved to 'false'. Details: ${err.message}`
+          : `Expression condition encountered an unexpected error and automatically resolved to 'false'. Details: ${JSON.stringify(
+              err,
+              null,
+              2
+            ).substring(0, 300)}`;
+
+      this.logError(errMessage);
+      this.emitTraceMessage(errMessage);
```
zhihil, line 118: Reworking the `catch`-all in expression and script condition to report the details of the error, rather than just giving a `expression condition encountered an unexpected error and automatically resolved to 'false'` debug trace without any elaboration.

This allows errors like an "undefined variable" to bubble up from the expression/script condition and into the prototype tool.

lib/services/runtime/handlers/message/lib/selectVariant.ts
```diff
@@ -26,12 +33,21 @@ export async function selectVariant(variants: VariantConditionPair[]): Promise<C
   for (const pair of conditionedVariants) {
     const isMatching = await pair.condition!.evaluate();
 
+    const notToken = isMatching ? '' : ' not';
+    log(
+      `Variant '${slateToPlaintext(pair.variant.data.text)}' was${notToken} used because its '${
+        pair.variant.condition?.type
+      }' condition evaluated to '${isMatching}'`
+    );
+
```
zhihil, line 42: Added debug traces to track the flow of the `selectVariant` logic. The idea here is to help the user see how the select variant logic works, i.e:

1. Check each conditioned variant from the top of the list to the bottom and choose the first matching one.
2. If no conditioned variant matches, then output a random unconditioned variant. 

Example:
Screenshot 2024-08-02 at 3 26 04 PM

matheuspoleza: that's pretty nice! really helpful

matheuspoleza (approved):

lib/services/runtime/handlers/message/lib/conditions/expression.condition.ts
```diff
@@ -113,23 +105,17 @@ export class ExpressionCondition extends BaseCondition<CompiledExpressionConditi
...
-          ).substring(0, 200)}`
-        );
-      }
-
-      this.emitTraceMessage(
-        `expression condition encountered an unexpected error and automatically resolved to 'false'`
-      );
+      const errMessage =
+        err instanceof Error
+          ? `Expression condition encountered an error and automatically resolved to 'false'. Details: ${err.message}`
```
pmvrmc, line 110: shouldnt we limit here the lenght to 300 characters too?

lib/services/runtime/handlers/message/lib/conditions/script.condition.ts
```diff
@@ -5,34 +5,30 @@ import { ConditionIsolate } from './conditionIsolate';
...
-
     const isolate = new ConditionIsolate(this.variables);
     try {
       await isolate.initialize();
-      const result = await this.evaluateCode(isolate, this.condition.data.code);
-
-      this.emitTraceMessage(`--- script evaluated to ${result} ---`);
-
-      return result;
+      return await this.evaluateCode(isolate, this.condition.data.code);
```
pmvrmc, line 19: do we really need `return await`? isnt return `enough`?

lib/services/runtime/handlers/message/lib/conditions/script.condition.ts
```diff
@@ -5,34 +5,30 @@ import { ConditionIsolate } from './conditionIsolate';
...
-            0,
-            300
-          )}`
-        );
-      }
-
-      this.emitTraceMessage(`script condition encountered an unexpected error and automatically resolved to 'false'`);
+      const errMessage =
+        err instanceof Error
+          ? `Script condition encountered an error and automatically resolved to 'false'. Details: ${err.message}`
```
pmvrmc, line 23: trimming here to 300 might be better too

lib/services/runtime/handlers/message/message.handler.ts
```diff
@@ -68,18 +67,16 @@ export const MessageHandler: HandlerFactory<CompiledMessageNode> = () => ({
         );
       }
 
-      const executionId = Utils.id.cuid();
-      const composeMessage = (message: string) => `[${executionId}]: ${message}`;
-      const logToPrototype = (message: string) => runtime.trace.debug(composeMessage(message));
-      const logToObservability = (message: string) => log.error(composeMessage(message));
+      const logToPrototype = (message: string) => runtime.trace.debug(message);
+      const logToObservability = (message: string) => log.error(message);
```
pmvrmc, line 71: nit pick: because this doesnt depend on anything on this fn, we could move this to the top level instead of the fn declaration to avoid having it being instantiated everytime the funciton runs.

really really minor

tests/lib/services/runtime/handlers/message/lib/conditions/conditionIsolate.unit.ts
```diff
@@ -36,7 +36,7 @@ describe('ConditionIsolate', () => {
       const isolate = new ConditionIsolate(variables);
       await isolate.initialize();
 
-      const result = await isolate.executeFunction('return propA + propB + propC');
+      const result = await isolate.executeFunctionOrScript('return propA + propB + propC');
```
pmvrmc, line 39: shouldnt we add a new test when. not using return? ie, `propA + propB + propC`

pmvrmc (approved):

lib/services/runtime/handlers/message/lib/conditions/expression.condition.ts
```diff
@@ -113,23 +105,17 @@ export class ExpressionCondition extends BaseCondition<CompiledExpressionConditi
...
+          : `Expression condition encountered an unexpected error and automatically resolved to 'false'. Details: ${JSON.stringify(
+              err,
```
trs, line 112: This would have the `stack` from our code, right? Can we remove the `stack` in that case.

PR #909, opened 2024-08-12
[Merged by Bors] - feat: add set v3 handler (COR-2920)

zhihil:
**Fixes or implements COR-2920**

Implemented runtime handler for the new Set V3 step. See the associated review environment to E2E test the functionality of this PR.

Added a new SetV3Handler in `runtime/lib/Handlers/setV3/setV3.handler.ts` which will execute the new Set V3 compiled node.

Set V3 is largely identical to the existing Set V2, however, the step now allows two approaches to defining the set value:
1. **Value set** - Implemented in `evaluateValue()` in `runtime/lib/Handlers/setV3/lib/evaluateValue.ts`
2. **Script set** - Implemented in `evaluateScript()` in `runtime/lib/Handlers/setV3/lib/evaluateScript.ts`

The Set V3 step will execute `evaluateValue()` and `evaluateScript()` to get the value which we will assign to a variable. These values are collected into an `afterValues` object, essentially an update patch, which is then applied to the Voiceflow variables.

Finally, we compare `afterValues` to the `beforeValues`, the values of the Voiceflow variables before the set updates, to output a mutation diff on the Prototype Tool for the user's debugging. 

### Setup information

N/A

### Deployment Notes

N/A

### Related PRs

N/A

### Checklist


runtime/lib/Handlers/setV3/setV3.handler.ts
```diff
@@ -0,0 +1,109 @@
+import { RuntimeLogs } from '@voiceflow/base-types';
```
zhihil, line 1: Start reviews from this file, as this contains the main changes of the PR.

runtime/lib/Handlers/setV3/setV3.handler.ts
```diff
@@ -0,0 +1,109 @@
...
+async function evaluate(
+  setItems: CompiledSetV3NodeDataItem[],
+  variables: Store,
+  logToPrototype: (message: string) => void
+) {
+  const scriptSets = setItems.filter((item) => item.type === SetV3NodeDataItemType.SCRIPT);
+  const valueSets = setItems.filter((item) => item.type === SetV3NodeDataItemType.VALUE);
+
+  const scriptSetResults = await evaluateScript(scriptSets, variables, logToPrototype);
+  const valueSetResults = await evaluateValue(valueSets, variables, logToPrototype);
```
zhihil, line 36: NOTE: Here, we pass in _all_ of the script sets into `evaluateScript()` and _all_ of the expression sets into `evaluateValue()`.

The reason this is done is because of the script set. We do _not_ want to call `evaluateScript()` on each individual script set and allocate a new sandbox instance for each execution. The `isolated-vm` sandbox requires a minimum of 8 MB to instantiate, so for example, if we ran 4 script conditions in parallel, we would consume 32 MB of memory.

To avoid wasteful memory usage, we allocate a single `isolated-vm` instance (8 MB) and execute _all_ of the script sets on this single instance. 

The `evaluateValue()` does not need to be called on all of the value sets, we could have implemented it to be called on each individual value set. The `evaluateValue()` is called on all value sets just so that its interface is consistent with `evaluateScript()`.

runtime/lib/Handlers/setV3/lib/evaluateScript.ts
```diff
@@ -0,0 +1,59 @@
...
+  logToPrototype: (message: string) => void
+): Promise<Record<string, unknown>> {
+  const isolate = new IVMIsolate(variables.getState());
+
+  try {
+    const assignmentMap: Record<string, unknown> = {};
+
+    await isolate.initialize();
+
+    await Promise.all(setItems.map((item) => evaluateSingleScript(isolate, item, assignmentMap, logToPrototype)));
```
zhihil, line 53: The `evaluateScript()` function executes each script code in parallel and writes the result of the code into `assignmentMap`.

Once all writes are complete, the update patch `assignmentMap` is returned to the set v3 handler.

runtime/lib/Handlers/setV3/lib/evaluateValue.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export async function evaluateValue(
+  setItems: CompiledSetV3NodeDataItem[],
+  variables: Store,
+  logToPrototype: (message: string) => void
+): Promise<Record<string, unknown>> {
+  const assignmentMap: Record<string, unknown> = {};
+
+  setItems.forEach((item) => {
```
zhihil, line 28: The `evaluateValue()` function evaluates each "set value" assignment sequentially and creates an update patch called `assignmentMap`

runtime/lib/Handlers/setV3/lib/evaluateValue.ts
```diff
@@ -0,0 +1,54 @@
...
+
+import { getSetValue, getSetVariable } from './utils';
+
+/**
+ * Evaluate the given `value` as follows:
+ * 1. Substitute any Voiceflow variables in `value`
+ * 2. Parse and return `value` as a number (if `value` can be interpreted as a number)
+ * 3. Otherwise, parse and return `value` as a string
+ */
+function evaluate(value: string, variables: Record<string, unknown>) {
```
zhihil, line 17: This function performs a simple, naive parsing scheme where assigned values in a set are parsed as `number` (if this makes sense) and otherwise falls back to the `string` type.

runtime/lib/Handlers/utils/ivmIsolate.ts
```diff
@@ -1,6 +1,6 @@
 import ivm from 'isolated-vm';
```
zhihil, line 1: Renamed `ConditionIsolate` to `IVMIsolate` and moved to a new folder, because it can be used between thestep (its original context) and Set step.

utils/error/toErrorReasonString.ts
```diff
@@ -0,0 +1,6 @@
+import { truncateOutputString } from '../string/truncateOutputString';
+
+export const toErrorReasonString = (err: unknown, maxLength = 100) =>
```
zhihil, line 3: This utility function generates a "reason string" that serializes the given error, truncates the error if it's too long, and outputs a sentence fragment `"due to error:"` which can be used to construct an error message in various places.

utils/string/joinTokens.ts
```diff
@@ -0,0 +1,4 @@
+export const joinTokens = (
+  tokens: string[],
+  { filterCond = Boolean, sep = ' ' }: { filterCond?: (value: any) => boolean; sep?: string } = {}
+) => tokens.filter(filterCond).join(sep);
```
zhihil, line 4: The `joinTokens` utility takes an array of string tokens, filters out any falsy values, and joins the filtered array with the separator.

The use-case is in constructing dynamic error messages, while ensuring that message strings only have a single space between words.

lib/services/runtime/handlers/message/lib/conditions/script.condition.ts
```diff
@@ -1,10 +1,10 @@
 import { CompiledScriptCondition } from '@voiceflow/dtos';
 
+import { IVMIsolate } from '../../../../../../../runtime/lib/Handlers/utils/ivmIsolate';
```
theProf, line 3: use '@' to meet repo root (as set in tsconfig)
```suggestion
import { IVMIsolate } from '@/runtime/lib/Handlers/utils/ivmIsolate';
```

lib/services/runtime/handlers/message/lib/conditions/value-variable.condition.ts
```diff
@@ -1,8 +1,8 @@
 import { replaceVariables } from '@voiceflow/common';
 import { CompiledConditionAssertion, CompiledValueVariableCondition, ConditionOperation } from '@voiceflow/dtos';
 
+import { IVMIsolate } from '../../../../../../../runtime/lib/Handlers/utils/ivmIsolate';
```
theProf, line 4:
```suggestion
import { IVMIsolate } from '@/runtime/lib/Handlers/utils/ivmIsolate';
```

runtime/lib/Handlers/setV3/lib/evaluateScript.ts
```diff
@@ -0,0 +1,60 @@
+import { CompiledSetV3NodeDataItem, SetV3NodeDataItemType } from '@voiceflow/dtos';
+import _ from 'lodash';
```
theProf, line 2:
```suggestion
```
This doesn't seem used

runtime/lib/Handlers/setV3/lib/evaluateScript.ts
```diff
@@ -0,0 +1,60 @@
...
+  const isolate = new IVMIsolate(variables.getState());
+
+  try {
+    const updatePatch: Record<string, unknown> = {};
+
+    await isolate.initialize();
+
+    await evaluate(isolate, item, updatePatch, logToPrototype);
+
+    return updatePatch;
```
theProf, line 56:
```suggestion
    await isolate.initialize();

    return await evaluate(isolate, item, updatePatch, logToPrototype);
```
passing by reference and mutating adds unnecessary complexity. `evaluate` can be in charge of creating and returning the `updatePatch`

runtime/lib/Handlers/setV3/lib/evaluateScript.ts
```diff
@@ -0,0 +1,60 @@
...
+    if (!item.value) {
+      throw new Error(`code is undefined`);
+    }
+
+    if (!item.variable) {
+      throw new Error(`target variable is undefined`);
+    }
+
+    const value = await isolate.executeFunctionOrScript(item.value);
+    updatePatch[item.variable] = value;
```
theProf, line 32:
```suggestion
    return { [item.variable]: await isolate.executeFunctionOrScript(item.value) };
```
and shouldn't take `updatePatchItem` as a param. see below comment for reasoning

runtime/lib/Handlers/setV3/lib/evaluateValue.ts
```diff
@@ -0,0 +1,53 @@
...
+  } catch (err) {
+    const setValue = getSetValue(item);
+    const prefix = joinTokens([`Could not set`, getSetVariable(item), setValue ? 'to' : '', setValue]);
+    const suffix = toErrorReasonString(err);
+    const errMessage = joinTokens([prefix, suffix]);
+
+    logToPrototype(errMessage);
+  }
+
+  return assignmentMap;
```
theProf, line 52: keep the function shape consistent between `evaluateValue` and `evaluateScript` (this has a return outside of the try/catch)

runtime/lib/Handlers/setV3/setV3.handler.ts
```diff
@@ -0,0 +1,111 @@
...
+  // eslint-disable-next-line no-restricted-syntax
+  for (const item of setItems) {
+    const updatePatch = await evaluateSet(item, variables, logToPrototype);
+
+    Object.entries(updatePatch).forEach(([variableName, newValue]) => {
+      variables.set(variableName, newValue);
+      runtime.variables.set(variableName, newValue);
+      afterValues[variableName] = newValue;
+    });
+  }
```
theProf, line 39: does this rely on ordering? does one iteration of the loop/`evaluateSet` rely on the previous mutation (of `variables` for example)?

runtime/lib/Handlers/setV3/setV3.handler.ts
```diff
@@ -0,0 +1,111 @@
...
+export const SetV3Handler: HandlerFactory<CompiledSetV3Node> = () => ({
+  canHandle: (node) => CompiledSetV3NodeDTO.safeParse(node).success,
+  handle: async (node, runtime, variables) => {
+    try {
+      if (!node.data?.items) {
+        return node.nextId ?? null;
+      }
+
+      // skip set assignments with a missing value or missing variable
+      const validSets = node.data.items.filter((item) => item.value && item.variable);
```
theProf, line 82: i see this assumption check in many places. not for this PR but it would be better to have a type and a guard so we only have to check once

utils/string/truncateOutputString.ts
```diff
@@ -0,0 +1,5 @@
+export function truncateOutputString(text: string, maxLength = 50) {
+  const truncatedText = text.substring(0, maxLength);
+  const isTruncated = truncatedText.length === text.length;
+  return truncatedText + (isTruncated ? '' : '...');
+}
```
theProf, line 5: a consideration is if `maxLength` is meant to be the full length of the string, you'll need consider the length of your suffix:
```suggestion
export function truncateOutputString(text: string, maxLength = 50) {
  const suffix = '...';
  const truncatedText = text.substring(0, maxLength - suffix.length);
  return  text.length > maxLength ? `${truncatedText}${suffix}` : text;
}
```

theProf (approved):

runtime/lib/Handlers/setV3/setV3.handler.ts
```diff
@@ -0,0 +1,111 @@
...
+    Object.entries(updatePatch).forEach(([variableName, newValue]) => {
+      variables.set(variableName, newValue);
+      runtime.variables.set(variableName, newValue);
+      afterValues[variableName] = newValue;
+    });
+  }
+  /* eslint-enable no-await-in-loop */
+
+  return afterValues;
+}
```
theProf, line 43:
```suggestion
async function evaluate(
  setItems: CompiledSetV3NodeDataItem[],
  runtime: Runtime,
  variables: Store,
  logToPrototype: (message: string) => void
) {
  const afterValues: Record<string, unknown> = {};
  const setItem$ = setItems.map((item) => evaluateSet(item, variables, logToPrototype));

  // TODO: more functional but highly mutation/side effect based
  (await Promise.all(setItem$))
    .flatMap((updatePatch) => Object.entries(updatePatch))
    .forEach(([variableName, newValue]) => {
      variables.set(variableName, newValue);
      runtime.variables.set(variableName, newValue);
      afterValues[variableName] = newValue;
    });

  return afterValues;
}
```

lib/services/runtime/handlers/message/lib/conditions/script.condition.ts
zhihil: Automatically rewrite imports after moving file messed up the transformation

runtime/lib/Handlers/setV3/lib/evaluateScript.ts
zhihil: You're right, not sure why linting didn't catch that.

runtime/lib/Handlers/setV3/lib/evaluateScript.ts
zhihil: Leftover pattern from a previous iteration where this was more useful. Removed.

runtime/lib/Handlers/setV3/setV3.handler.ts
zhihil: Yes, ordering is part of the product requirements. Originally this was an `await Promise.all()` but that doesn't work.

runtime/lib/Handlers/setV3/setV3.handler.ts
zhihil: We cannot make this change because the product requirements are that subsequent sets are dependent on previous sets' results. 

The `Promise.all` will execute all sets concurrently which has nondeterministic order.

theProf: thanks! this suggestion went along with my other comment asking if the order did matter, and if it didn't we could do the above. not a sticking point

runtime/lib/Handlers/setV3/setV3.handler.ts
zhihil: Screenshot 2024-08-15 at 9 50 58 AM

So we are not filtering out the sets on the compiler level.

zhihil: I've update this code so that the runtime contains a simplified type where `variable` and `value` are defined. After the filtering and clean-up is complete, the runtime code should mostly use the. simplified type to simplify the overall code.

runtime/lib/Handlers/setV3/lib/evaluateScript.ts
```diff
@@ -40,9 +28,9 @@ async function evaluate(
 }
 
 export const evaluateScript: SetEvaluator = async (
-  item: CompiledSetV3NodeDataItem,
-  variables: Store,
-  logToPrototype: (message: string) => void
+  item,
+  variables,
+  logToPrototype
```
theProf, line 33: why remove the typings?

PR #930, opened 2024-08-23
fix: not properly checking the auth key (BUG-1332)

zhihil:
**Fixes or implements BUG-1332**

`resolveVersionAlias()` isn't properly checking the DM API key in some cases, allowing users to bypass API authorization, e.g, by providing the wrong DM API key or by providing a value that isn't even a valid DM API key.

Restructured the `resolveVersionAlias()` middleware to do the following:
1. If `Authorization` doesn't contain a DM API key (undefined, random gibberish, typo) then the interaction request is completely rejected (this fixes the case where `'dsdafsd'` can be used in `Authorization`)
2. Performs a lookup with the DM API key. If the DM API key does not correspond to any Voiceflow project, then the interaction is denied. 
3. Extracts the `req.headers.versionID` value. If the request `versionID` does not match with the `devVersion` or `liveVersion` of the looked up project, then the interaction is denied (fixes the case where any valid DM API key can look up any project)
4. If the request `versionID` is an alias like `'development'` or `'production'`, then it is resolved to the `devVersion` or `liveVersion` of the retrieved project, respectively. 
5. By this point, `Authorization` is defined, is a DM API key, corresponds to some valid project, and the correct `versionID` is the `liveVersion` or `devVersion` of the project. Under these conditions, we define the interaction as authorized. 

### Setup information

N/A

### Related PRs

### Checklist


trs:
See: https://github.com/voiceflow/general-runtime/pull/931

PR #931, opened 2024-08-23
feat: use auth guard on interact routes (BUG-1332)

trs:
resolveVersionAlias() isn't properly checking the DM API key in some cases, allowing users to bypass API authorization, e.g, by providing the wrong DM API key or by providing a value that isn't even a valid DM API key.

The current middleware is responsible for resolving the version from the given key and/or alias, no necessarily auth. Instead, we can actually authorize the request using the auth sdk.


backend/api/routers/state.ts
```diff
@@ -15,6 +15,12 @@ export default (middlewares: MiddlewareMap, controllers: ControllerMap) => {
     middlewares.rateLimit.versionConsume,
     middlewares.project.resolveVersionAlias,
     middlewares.project.attachProjectID,
+    middlewares.auth.authorize(['project:READ'], (req) => {
+      return {
+        id: req.headers.versionID,
+        kind: 'version',
+      };
+    }),
```
zhihil, line 23: We may need this check in `nlu.ts` and `transcript.ts` as well? Those two routers also use `.resolveVersionAlias`.

It makes sense that transcripts, potentially sensitive information, should be protected under API key. Not sure about the NLU router.

zhihil (approved):

trs:
bors merge

trs:
bors cancel

trs:
bors merge

trs:
bors retry

PR #947, opened 2024-08-27
[Merged by Bors] - build(fix): single platform for checks target running tests (PL-000)

theProf:

The directory structure for local output changes for a multiplatform build. We don't need a multi-platform build for the `checks` stage that runs test though so reverting for that step only.


xavidop (approved):

theProf:
Bors r+

